PR #18, opened 2020-01-30
Show log as info, not critical

fabioz:


int19h:
Can you rebase on the current master, so that it re-runs the updated tests? It should take care of the 32-bit run.

PR #36, opened 2020-02-04
Re-attach to pid when using --client. 

fabioz:
Fixes https://github.com/microsoft/ptvsd/issues/1817


tests/debugpy/test_attach.py
```diff
@@ -148,40 +148,53 @@ def code_to_debug():
...
 
-        session.before_request = before_request
-        session.config["redirectOutput"] = True
+    pid = session1.config["processId"]
 
-        with session.attach_by_pid(target(code_to_debug), wait=False):
-            session.set_breakpoints(code_to_debug, all)
+    # XXX: close_channel=False is a hack: when trying to close the channel, we seem to deadlock
+    # trying to get the JsonMessageChannel lock on channel.close().
+    session1.disconnect(close_channel=False)
```
fabioz: @int19h do you know the reason for the deadlock here when trying to close the channel? 

In the previous version of this test, if the code ran to completion it worked, but now that I'm disconnecting sooner it deadlocks if we try to get the lock to close the channel without waiting for the program to complete.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_trace_dispatch_regular.py
```diff
@@ -181,7 +181,7 @@ def fix_top_level_trace_and_get_trace_func(py_db, frame):
             return f_trace, False
 
     thread_tracer = additional_info.thread_tracer
-    if thread_tracer is None:
+    if thread_tracer is None or thread_tracer._args[0] is not py_db:
```
fabioz: This one line fix took me quite a while to find...

tests/debugpy/test_attach.py
int19h: There is a fundamental problem with close() on file descriptors on Windows - it takes a lock on the fd, the same one that read() and write() do. So if another thread is sitting on a read(), close() will block. Basically there's no way to interrupt a synchronous read of an fd. 

This is usually not a problem in tests, because the only time it's using fds is when talking to the adapter over stdio, which is mostly used for "launch", but also for attach-to-PID. And in most tests, you have a single debuggee process, and that process exits before `debug.Session` is closed, so the fd is closed from the other end. But if you're trying to disconnect while the debuggee is still running, and you're talking to the adapter over stdio, then you get this problem.

The annoying thing about all this is that it's specifically a limitation for the POSIX-compat file descriptor API in the C runtime. The underlying Win32 API can close a file handle that has an outstanding synchronous read on it just fine. But Python wraps the C runtime; and so we're stuck with this behavior, unless we do platform-specific implementations for the stdio transport.

int19h: Since this is a one-off workaround, rather than adding a new flag to `Session.disconnect()`, I think it would be better to just do what needs to be done directly - i.e. `request("disconnect")` and then wait for the "terminated" event.

fabioz: Done.

PR #37, opened 2020-02-05
Notify user about gevent before halting debugger. 

fabioz:
Fixes https://github.com/microsoft/ptvsd/issues/2057

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
@@ -151,7 +151,15 @@ try:
 except AttributeError:
     PY_IMPL_NAME = ''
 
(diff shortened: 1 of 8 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
```diff
@@ -151,7 +151,15 @@ def version_str(v):
 except AttributeError:
     PY_IMPL_NAME = ''
 
-SUPPORT_GEVENT = os.getenv('GEVENT_SUPPORT', 'False') == 'True'
+SUPPORT_GEVENT = os.getenv('GEVENT_SUPPORT', 'False') in ('True', 'true', '1')
+
+GEVENT_SUPPORT_NOT_SET_MSG = os.getenv(
```
fabioz: Note: I made the message configurable through an environment variable, but I `ptvsd` should probably set it on `force_pydevd.py` which also sets other constants `pydevd_constants` if needed.

PR #39, opened 2020-02-11
Refactor debugpy API and CLI for clarity and consistency.

int19h:


src/debugpy/__init__.py
```diff
@@ -34,133 +38,166 @@
...
 
+def breakpoint():
+    """If a client is connected to the debug adapter that is debugging
+    this process, pauses execution of all threads, and simulates a
+    breakpoint being hit at the line following the call.
 
-def break_into_debugger():
-    """If the IDE is connected, pauses execution of all threads, and
-    breaks into the debugger with current thread as active.
+    On Python 3.8 and above, this is the same as builtins.breakpoint().
```
karthiknadig:
```suggestion
    On Python 3.7 and above, this is the same as builtins.breakpoint().
```

PR #41, opened 2020-02-12
Deal with exceptions in hasattr/dir/isinstance. Fixes https://github.com/microsoft/ptvsd/issues/2076

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py
```diff
@@ -271,3 +271,28 @@ def notify_about_gevent_if_needed(stream=None):
                 return True
 
     return False
+
+
+def hasattr_checked(obj, name):
+    try:
+        getattr(obj, name)
+    except Exception:
```
int19h: Should this be just `except:`, like `isinstance_checked` below? I don't know if there's any way one can propagate something like `StopIteration` through `getattr()`, but just in case...

fabioz: Done

PR #45, opened 2020-02-13
Use launcher pid as ppid in DAP. Fixes #42

### LINKED ISSUE #42: Subprocesses not reported to client in virtualenv
This is specific to Python 3.7+ on Windows, when code is running from a virtual environment, and uses `sys.executable` to spawn a subprocess.

The underlying cause is the same as in https://github.com/microsoft/ptvsd/issues/1930: `sys.executable` points to a stub binary that spawns the actual Python interpreter, and so when the subprocess spawns and connects to the debug adapter, its parent PID is the stub. But the adapter needs to know the actual parent PID to determine which debug session to send the "debugpyAttach" event to - and if it can't find a session matching that PID, it doesn't send the notification at all. Meanwhile, the subprocess remains in suspended state, waiting for something to attach to it.

### LINKED ISSUE #43: Subprocesses are stuck if they have no known parent
This happens in #42 as a side effect of another bug, but it can actually happen even when everything works correctly - e.g. if a process spawns a subprocess and then exits immediately, by the time the subprocess will connect to the adapter, the session for its parent might be already finalized. 

Whenever this happens, no debug session receives the "debugpyAttach" event about the new subprocess, so no client connects to it - but the subprocess is waiting for something to attach, and won't run any user code until then.

fabioz:


int19h:
@fabioz, I just did a little tweak in 7454f89 that should help a bit with the tests - or at least it shouldn't mask the actual error. Can you rebase and restart the test run?

fabioz:
Note: I'm still investigating the error in the ci...

fabioz:
The problem was properly supporting fork+exec (the `ppid` ended up being == `pid`).

PR #61, opened 2020-03-06
Change references from debugStdLib to justMyCode. Fixes https://github.com/microsoft/ptvsd/issues/1469

fabioz:


fabioz:
It seems all the failures are related to:

```
   File "d:\a\1\s\src\debugpy\adapter/../..\debugpy\common\sockets.py", line 69, in serve

     listener = create_server(host, port, backlog, timeout)

   File "d:\a\1\s\src\debugpy\adapter/../..\debugpy\common\sockets.py", line 23, in create_server

     server.bind((host, port))

 OSError: [WinError 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted
```

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
```diff
@@ -37,8 +38,13 @@ def to_json(self):
         return json.dumps(dct)
 
     def update_fom_debug_options(self, debug_options):
-        if 'DEBUG_STDLIB' in debug_options:
-            self.debug_stdlib = debug_options.get('DEBUG_STDLIB')
+
+        if 'JUST_MY_CODE' in debug_options:
```
karthiknadig: @int19h Do we need the debug options part at all?

int19h: Nope.

int19h:
The socket failures seem to be a sporadic ones that have increased since I've added listen/connect for tests. Not entirely sure why, as the port ranges are distinct. I'll look into it.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
fabioz: I've created an issue to address that:

https://github.com/microsoft/debugpy/issues/64

Issue #72, opened 2020-03-13
"Wait on exit" blocks indefinitely when there's no terminal


int19h:
_Originally posted by @huguesv in https://github.com/microsoft/debugpy/issues/57#issuecomment-592196970_

int19h:
Also happens if using "internalConsole" on any Python version. It should be checking generically whether the terminal is interactive or not.

PR #74, opened 2020-03-13
Fix #69, #72

### LINKED ISSUE #69: "Couldn't spawn debuggee" with non-ASCII filename on Python 2.7
Issue Type: <b>Bug</b>

I tried running the debugger on a program. I then got an error message reading: "Couldn't spawn debuggee: 'ascii' codec can't encode character u/2013 in position 27: ordinal not in range(128)". What should I do?

Extension version: 2020.2.64397
VS Code version: Code 1.43.0 (78a4c91400152c0f27ba4d363eb56d2835f9903a, 2020-03-09T19:47:57.235Z)
OS version: Windows_NT x64 10.0.16299

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) m3-7Y30 CPU @ 1.00GHz (4 x 1608)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>protected_video_decode: enabled<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>viz_display_compositor: enabled_on<br>viz_hit_test_surface_layer: disabled_off_ok<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|3.90GB (0.53GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>

int19h:


src/debugpy/common/log.py
```diff
@@ -332,7 +332,7 @@ def report_paths(get_paths, label=None):
     site_packages = [
         p
         for p in sys.path
-        if os.path.exists(p) and os.path.basename(p) == "site-packages"
+        if os.path.exists(p) and os.path.basename(p) == compat.filename_str("site-packages")
```
fabioz: `sys.path` entries can actually be unicode (as was the case for the recent issue on Maya), so, you may want to do `filename_str` on `basename` too -- this use-case probably warrants a test too.

int19h: In this case, it wouldn't be a problem. The reason why it was a problem before is because the literal was `unicode` on Python 2, and that caused the left side to be implicitly converted to Unicode also, with "ascii" as codec - which breaks if it actually has any non-ASCII symbols in it. Now that the right side is always `bytes` on Python 2, if the left side is `unicode`, it's the right side that will be auto-decoded as "ascii" - which is fine, because it really is ASCII.

fabioz: Thanks for the explanation (it didn't occur to me that having ascii bytes on the right side would never raise an exception).

PR #89, opened 2020-03-27
Show bytecode when source code cannot be fetched.

fabioz:
Fixes https://github.com/microsoft/ptvsd/issues/1304


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_uncompyle.py
```diff
@@ -0,0 +1,642 @@
+"""
```
fabioz: Just to note, as commented, this was a bit of a proof of concept to see how much work it'd be to make a decompiler (so, I worked a bit more than a day just to see what'd be the size of the task if we wanted to do it given that I haven't found any non-GPL implementation -- after this experiment, I think it'd be a really big task -- I think it's probably close to 1-2 months to implement something really usable so, I'm unsure whether we want to tackle it or not in the future).

karthiknadig: Given the 1-2 month duration, I think we should put this on hold.

fabioz: Sure, agreed on the full decompile, still, this comment is only related to that particular file (the current pull request integrates a version that shows a simple bytecode decompile which should be functional).

int19h:
Since `uncompyle` is the name of a specific library, it would be best to avoid using that, so as to not create the impression that the code is derived from it. And since this implementation only does bytecode, I think `disassemble` would be more accurate, anyway?

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_uncompyle.py
int19h: Users are often confused by "Source not found" error message, and believe that it applies to debug session as a whole - i.e. they don't realize that they can still continue stepping or F5 to keep going. So I think this is worthwhile even with bytecode alone.

fabioz: Agreed... I'll rename it.

PR #120, opened 2020-04-15
Fix CLI help/usage text.

int19h:


src/debugpy/server/cli.py
```diff
@@ -26,7 +26,8 @@
 HELP = """debugpy {0}
 See https://aka.ms/debugpy for documentation.
 
-Usage: debugpy [--listen | --connect] [<address>:]<port>
+Usage: debugpy --listen | --connect
```
karthiknadig, line 29: We should probably block and show help if both listen and connect are used. This is not critical, but we should add that.

int19h: It already does that - it was just the docstring that didn't reflect it properly.

https://github.com/microsoft/debugpy/blob/2f21b08ae02abe5407ac3dbd74a19ce548fe87c4/src/debugpy/server/cli.py#L97-L98

Some other checks:

https://github.com/microsoft/debugpy/blob/2f21b08ae02abe5407ac3dbd74a19ce548fe87c4/src/debugpy/server/cli.py#L218-L223

PR #127, opened 2020-04-17
Show globals, group variables and show all dir() variables.

fabioz:
Fixes https://github.com/microsoft/ptvsd/issues/118
Fixes https://github.com/microsoft/ptvsd/issues/763
Fixes https://github.com/microsoft/ptvsd/issues/1621

### DIFF
--- a/tests/debug/config.py
+++ b/tests/debug/config.py
@@ -69,6 +69,7 @@ class DebugConfig(collections.MutableMapping):
         "connect": (),
         # Attach by PID
         "processId": (),
+        "variablePresentation": {},
     }
 
     def __init__(self, *args, **kwargs):
--- a/tests/debug/config.py
+++ b/tests/debug/config.py
@@ -69,7 +69,12 @@ class DebugConfig(collections.MutableMapping):
         "connect": (),
         # Attach by PID
         "processId": (),
-        "variablePresentation": {},
+        "variablePresentation": {
+            "special": "group",
+            "function": "group",
+            "class_": "group",
+            "protected": "inline",
+        },
     }
 
     def __init__(self, *args, **kwargs):
--- a/tests/debugpy/test_evaluate.py
+++ b/tests/debugpy/test_evaluate.py
@@ -102,7 +102,7 @@ def test_variables(pyfile, target, run):
                 {
                     "type": "int",
                     "value": "2",
-                    "name": "__len__",
+                    "name": "len()",
                     "evaluateName": "len(b)",
                     "variablesReference": 0,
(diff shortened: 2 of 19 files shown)


tests/debug/config.py
```diff
@@ -69,6 +69,7 @@ class DebugConfig(collections.MutableMapping):
         "connect": (),
         # Attach by PID
         "processId": (),
+        "variablePresentation": {},
```
int19h, line 72: This should have the default values for the properties (in case some test-generic code needs to query  what the value is, whether the test had set it or not).

tests/debugpy/test_evaluate.py
```diff
@@ -155,21 +160,45 @@ def code_to_debug():
             "__a_test",
             "__b_test",
             "__c_test",
-            "__a_test__",
-            "__b_test__",
-            "__c_test__",
         ]
 
+        special_vars_entry = [v for v in vars if v["name"] == "special variables"][0]
```
int19h, line 165: I usually use single-element tuple destructuring instead of indexing for cases like these, since it also implicitly asserts that there's only one such element, and throws otherwise; i.e.:
```py
special_vars_entry, = [v for v in vars if v["name"] == "special variables"]
```

tests/debugpy/test_evaluate.py
```diff
@@ -155,21 +160,45 @@ def code_to_debug():
...
+
         # String dict keys must be sorted as strings.
         b_test, = (v for v in vars if v["name"] == "b_test")
         b_test_vars = session.request(
             "variables", {"variablesReference": b_test["variablesReference"]}
         )["variables"]
         var_names = [v["name"] for v in b_test_vars]
         if sys.version_info[:2] >= (3, 6):
-            assert var_names == ["'spam'", "'eggs'", "'abcd'", "__len__"]
+            # Note that the special __len__ we manually create is not added to special variables.
```
int19h, line 182: `VariablePresentationHint` has a flag for generated variables like these, "virtual" - it would be nice to use it for synthetic `__len__`. I'm not sure if that makes any difference in UX in VSCode, but I recall that there is a different icon for it in VS.

Also, this may be confusing in a sense that `__len__` is an actual (class) variable that's expected to be a function. If I understand correctly, now we are going to have one `__len__` under "class variables" that's that, and then also `__len__` in normal variables that is the result of calling `len()` on the value. Way back when we had this in ptvsd 3, the way it was distinguished is using `len()` as the variable name. Since no actual attribute should have `()` in the name, it makes it clear that it's synthetic.

fabioz:
Ok, I've changed `__len__` to len() and did the changes to the tests.

PR #149, opened 2020-04-29
Only handle django and jinja2 exceptions if handling caught exceptions. Fixes #128

### LINKED ISSUE #128: Django templates intermittently throw failed key lookup exceptions
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/925339/django-templates-intermittently-throw-failed-key-l-1.html)._

---
When rendering a template in Visual Studio's Django test server, I intermittently observe the following error:

`Message=Failed lookup for key [addRowButton] in [...] `

The line this issue triggers on looks like this (lightly modified to hide structure):

`{% include '.../components/gen_formset_table_form.html' with form=next formset=formset addRowButton=addRowButton addSortable=addSortable %}`

It is true that addRowButton is undefined in this template. However, only on some runs is this exception thrown, and having references to undefined keys should *not* throw exceptions in Django templates.

Once this issue is observed on a given run of the test server, it will consistently reappear every time this template is rendered. And, vice-versa, if the issue does not appear, it never appears. Furthermore, simply pressing the "restart" button on the toolbar is likely to fix the issue.

I also observe on most runs of the test server the following, possibly related issue:

```
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 59880)
[20/Feb/2020 20:57:16] "GET /ajax/ing/setup/json/ HTTP/1.1" 200 712
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\socketserver.py", line 360, in finish_request
    self. RequestHandlerClass(request, client_address, self)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\socketserver.py", line 720, in __init__
    self.handle()
  File "[venvpath]\lib\site-packages\django\core\servers\basehttp.py", line 174, in handle
    self.handle_one_request()
  File "[venvpath]\lib\site-packages\django\core\servers\basehttp.py", line 182, in handle_one_request
    self.raw_requestline = self.rfile.readline(65537)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\socket.py", line 589, in readinto
    return self._sock.recv_into(b)
ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine
----------------------------------------
```

---
#### Visual Studio Feedback System on 2/20/2020, 11:15 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### Visual Studio Feedback System on 3/26/2020, 06:52 PM: 

I have detected that for the last 35 days, this issue didn't have much product team activity and a very small amount of new votes or comments. Based on this, its severity, and affected area, it’s my experience that this issue is very unlikely to be fixed.
#### Hugues Valois [MSFT] on 3/27/2020, 10:27 AM: 

<p>Thanks for the feedback.</p>
<p>We’ve just released Visual Studio 2019 16.5 and in which we’ve changed the “launch browser” action to wait until the console shows that web url, meaning it is ready to accept connections. What we had noticed in the past was that if it didn’t wait long enough, it could cause weird trouble in the connection leading to page errors.</p>
<p>There’s a small chance that it might help in your scenario, it would be worth trying when you get a chance to update VS.</p>
<p>If it doesn’t help, then I’d try to do those same things from command line rather than in VS and see if the issue still occurs.</p>
<p>thanks</p>

#### Visual Studio Feedback System on 4/7/2020, 07:18 PM: 

We will close this report in 14 days because we don’t have enough information to investigate further. To keep the problem open, please provide the requested details.
#### sirbayer on 4/12/2020, 00:43 PM: 

<p>Unfortunately, I can confirm that this hasn't resolved the issue; I'm still seeing the same exceptions after updating to 16.5.3.</p><p>I've never observed these issues when running from the Ubuntu on Windows terminal. In fact, the exceptions generated by VS do not appear in the VS terminal, either - they only seem to exist in the VS debug popups.</p>

#### Hugues Valois on 4/16/2020, 01:49 PM: 

<p>Thanks sirbayer.</p>
<p>I’ve set up a repro for this and I can see that the behavior has changed between VS 2017 and VS 2019.</p>
<p>In VS 2019, even though I have unchecked all Python exceptions in the Exception Settings tool window, it still breaks on that exception. It doesn’t cause any problems with the application, I press F5 to continue and some code must be handling that exception since things work fine, no error in terminal.</p>
<p>However, the debugger should not be breaking when the exception is thrown, if I have all of them disabled in settings, and there is code that handles the exception. In VS 2017, it did not break under the same circumstances.</p>
<p>I was worried this was broken everywhere but fortunately this seem to only be an issue when debugging django templates.</p>
<p>I’ll transfer this over to our bug repo on github. Thank you for filing this.</p>

---
### Original Solutions
(no solutions)

fabioz:


int19h:
Are all template exceptions always handled by Django code? I had the impression that this is only true for some specific errors, like missing variable.

fabioz:
> Are all template exceptions always handled by Django code? I had the impression that this is only true for some specific errors, like missing variable

In the debugger, we are only handling `'VariableDoesNotExist', 'TemplateDoesNotExist', 'TemplateSyntaxError'` as template exceptions in django (i.e.: https://github.com/microsoft/debugpy/blob/master/src/debugpy/_vendored/pydevd/pydevd_plugins/django_debug.py#L500).

The other exceptions are handled as usual in the debugger and aren't affected by the django setting.

PR #230, opened 2020-05-07
Fix #156: Add a flag for the old way of handling args from a launch config

### LINKED ISSUE #156: Add a flag for the old way of handling args from a launch config.
Prior to the most recent release, the "args" property of the launch config was passed directly to `spawn()`.  Now those args are passed to the launcher script where they can be shell-expanded.  However, in some cases that shell expansion is problematic and the old behavior is desired.  See https://github.com/microsoft/vscode-python/issues/11562.

This could be solved by supporting a flag in the launch config to indicate that the old behavior (no args sent to launcher script, args sent to `spawn()` directly) should be used.

int19h:
Add new property "argsExpansion", which defaults to "shell", but can be explicitly set to "none" to request no expansion.

Propagate "args" to launcher via CLI or JSON, depending on the value of "argsExpansion".

Move the logic to compute process name back to the adapter, alongside other processing of "launch" targets.

### DIFF
--- a/src/debugpy/adapter/clients.py
+++ b/src/debugpy/adapter/clients.py
@@ -272,22 +272,19 @@ class Client(components.Component):
 
             return value
 
-        # Propagate command line arguments via launcher CLI rather than "args", so that
-        # they get shell expansion applied to them in "runInTerminal" scenarios.
-
         program = module = code = ()
         if "program" in request:
             program = request("program", unicode)
             args = [program]
-            # process_name = program
+            request.arguments["processName"] = program
         if "module" in request:
             module = request("module", unicode)
             args = ["-m", module]
-            # process_name = module
+            request.arguments["processName"] = module
         if "code" in request:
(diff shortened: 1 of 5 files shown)


src/debugpy/adapter/clients.py
```diff
@@ -299,7 +296,11 @@ def property_or_debug_option(prop_name, flag_name):
                 '"program", "module", and "code" are mutually exclusive'
             )
 
-        args += request("args", json.array(unicode))
+        # Propagate "args" via CLI if and only if shell expansion is requested.
+        args_expansion = request("argsExpansion", json.enum("shell", "none", optional=True))
```
karthiknadig, line 300: Do we need to expose this in the extension? to provide completion or do we want to keep it internal.

int19h: I think it's best to expose. Some users might need the old behavior, and without this flag, it's just not possible.

PR #231, opened 2020-05-07
Consider filenames starting with '<' library code by default. Fixes #209

### LINKED ISSUE #209: Step into library import statement gives "Could not load source '<string>': Source unavailable."
Set a breakpoint to a ``import numpy`` statement and step into it. It'll raise: ``Could not load source '<string>': Source unavailable.``

VS Code Insiders 1.33.0
Extension version 2019.3.6215

launch.json file:

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        
        {
            "name": "Python-test",
            "type": "python",
            "request": "launch",
            "program": "${file}",
            "stopOnEntry": true,
            "cwd": "${workspaceFolder}",
        },
    ]
}
```

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
```diff
@@ -46,6 +46,23 @@ class DebugInfoHolder:
     PYDEVD_DEBUG_FILE = None
 
 
+# Any filename that starts with these strings is not traced nor shown to the user.
+# In Python 3.7 "<frozen ..." appears multiple times during import and should be ignored for the user.
+# In PyPy "<builtin> ..." can appear and should be ignored for the user.
+# <attrs is used internally by attrs
+# <__array_function__ is used by numpy
+IGNORE_BASENAMES_STARTING_WITH = ('<frozen ', '<builtin', '<attrs', '<__array_function__')
```
fabioz, line 54: Note that these are completely hidden and not traced at all (internally it's handled as if they were debugger files) -- these are still based on heuristics.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
```diff
@@ -46,6 +46,23 @@ class DebugInfoHolder:
...
+
+# Note: <string> has special heuristics to know whether it should be traced or not (it's part of
+# user code when it's the <string> used in python -c and part of the library otherwise).
+
+# Any filename that starts with these strings is considered user (project) code. Note
+# that files for which we have a source mapping are also considered as a part of the project.
+USER_CODE_BASENAMES_STARTING_WITH = ('<ipython',)
```
fabioz, line 61: This could potentially not be needed at all as to be traced they need to be in `setPydevdSourceMap` to properly work (but still, I decided to leave it as a reference and it may be useful to have the setting in the future).

fabioz:
Note: I see that some tests are failing. I'll investigate.

fabioz:
Ok, should be good now (the failures seem unrelated to the commit now).

PR #240, opened 2020-05-13
Test to check that rules are properly reapplied on attach. Fixes #226

### LINKED ISSUE #226: Just my code exclusions are cached and can't be updated.
## Environment data

- DEBUGPY version: 1.0.0b8
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.6

## Steps to reproduce:
Create a number of files in a folder:
Looper.py:
```
import baz
import foo
while(True):
    inp = input("Enter a string")
    foo.foo()
    baz.baz()
    print('End of loop')
```
foo.py
```
def foo():
    print('foo')
```
baz.py
```
def baz():
    print('baz')
```
And a launch.json with the following (replacing paths appropriately)
```
{
      "name": "Python: Attach using Process Id",
      "type": "python",
      "request": "attach",
      "processId": "${command:pickProcess}",
      "justMyCode": true,
      "rules": [
        {
          "include": true,
          "path": "d:/Training/DebugJMC/foo.py"
        },
        {
          "include": true,
          "path": "d:/Training/DebugJMC/looper.py"
        },
        {
          "include": false,
          "path": "**/*"
        }
      ]
    }
```

1. Start looper.py from VS code
1. Attach and see that breakpoints will bind in foo.py but not in baz.py
1. Detach
1. Switch the include value for foo.py to baz.py
1. Reattach. 

## Actual behavior

Breakpoints don't bind in either

## Expected  behavior

Breakpoints bind in baz.py but not foo.py

fabioz:


fabioz:
@int19h, I can't merge because the Lint is not working (and it's not really an issue in this pull request).

PR #243, opened 2020-05-13
Return empty type for DAPGrouper instances. Fixes #237

### LINKED ISSUE #237: Type for special variables appears as DAPGrouper
In VS debug locals window, when grouping is enabled, debugpy is surfacing the type for the groups as `DAPGrouper`. It should be reported as empty string, since that is a debugger internal type.

image

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
```diff
@@ -64,6 +64,9 @@ def get_var_data(self, fmt=None):
         elif name in (TOO_LARGE_ATTR, GENERATED_LEN_ATTR_NAME):
             attributes.append('readOnly')
 
+        if type_name == 'DAPGrouper' and self.value.__class__ == DAPGrouper:
```
int19h, line 67: It sounds like the first check is redundant - wouldn't it always be true for this class?

Also, I'm not sure if the `__class__` attribute can throw an exception somehow, and needs to be inside try/except.

fabioz: Well, I was trying to lower the surface there by checking the name first (as you said, accessing `__class__` can really throw an exception as would isinstance checks... with the code above this could still happen if the user does actually have a `DAPGrouper` class that didn't have a class or had some error accessing `__class__`, but I don't really see that happening in practice).

If you want I can change to just check the class and handle exceptions there instead of checking the name first (but personally I think the current code is reasonable).

int19h: Yeah, I was basically thinking of that pathological corner case. But if the point of the string name check is to handle the possibility that `__class__` might throw, I think it's better to just handle the potential exception, simply because it's clearer what the purpose of it is.

fabioz: Ok, I'll change.

PR #244, opened 2020-05-13
Pass presentationHint='locals' to 'Locals' scope. Fixes #236

### LINKED ISSUE #236: "presentationHint" is not set for variable scopes
DAP `Scope` interface defines `presentationHint` for the purposes of distinguishing standard scope (locals, globals etc). Visual Studio uses this to determine whether the scope is locals or not, to inline locals into the top level. However, pydevd does not set `presentationHint` for any of the scopes - only the label.

fabioz:


huguesv:
I've applied the fix locally and can confirm that this works as expected in VS.
image

int19h:
It's too bad they show the scope using the default icon for variables. Not sure who to ask, but is there anything that we can do when reporting scopes to change it? Or maybe they should just change it on their end. Doesn't much matter what, just so long as it's visually distinct.

huguesv:
Oh I didn't notice that. In MMD it's using a different icon:
image

I'll ask the VS debugger team.

PR #250, opened 2020-05-16
Don't duplicate memory when getting the contents of bytes/str. Fixes #242

### LINKED ISSUE #242: Debugger starts to paginate memory when getting the representation of big byte arrays
## Environment data

-   VS Code version: Version: 1.45.0
-   Extension version (available under the Extensions sidebar): 2020.5.78807
-   OS and version: MacOS Catalina 10.15.3
-   Python version (& distribution if applicable, e.g. Anaconda): Python 3.7.7 64 bit
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): pyenv
-   Jedi or Language Server? (i.e. what is `"python.jediEnabled"` set to; more info microsoft/vscode-python#3977): jedi
-   Value of the `python.languageServer` setting: Microsoft

## Expected behaviour

Show the local variables in debug 

## Actual behaviour

I can attach to a Flask web server - meaning that it stops on the line - but no local variables are displayed. 

image

## Steps to reproduce:

Launch debugger with either one of the below configs; neither works.

```

    "version": "0.2.0",
    "configurations": [
        
        {
            "name": "Flask Debug",
            "type": "python",
            "request": "launch",
            "module": "flask",
            "env": {
                "FLASK_APP": "app.py",
                "FLASK_ENV": "development",
                "FLASK_DEBUG": "0"
            },
            "args": [
                "run",
                "--no-debugger",
                "--no-reload"
            ],
            "jinja": true,
            "cwd": "${workspaceFolder}/server"
        },
        {
            "name": "Foo",
            "type": "python",
            "request": "launch",
            "module": "flask",
            "env": {
                "FLASK_APP": "app",
                "FLASK_DEBUG": "1",
                "FLASK_ENV": "development"
            },
            "args": [
                "run",
            ],
            "subProcess": true,
            "jinja": true,
            "cwd": "${workspaceFolder}/server"
        }
    ]
}
```

fabioz:
Note: I still haven't done anything about dealing with raw strings in this request (waiting for definition on the best way to handle it in that case).

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
@@ -6,7 +6,7 @@
 import sys
 from _pydevd_bundle.pydevd_constants import IS_PY2
 import locale
-import json
+from _pydev_bundle import pydev_log
 
 # Py3 compat - alias unicode to str, and xrange to range
 try:
@@ -33,6 +33,7 @@ class SafeRepr(object):
     maxstring_inner = 30
     if sys.version_info >= (3, 0):
         string_types = (str, bytes)
+        bytes = bytes
         set_info = (set, '{', '}', False)
         frozenset_info = (frozenset, 'frozenset({', '})', False)
         int_types = (int,)
@@ -40,6 +41,7 @@ class SafeRepr(object):
                            dict, set, frozenset)
     else:
         string_types = (str, unicode)
+        bytes = str
         set_info = (set, 'set([', '])', False)
         frozenset_info = (frozenset, 'frozenset([', '])', False)
         int_types = (int, long)  # noqa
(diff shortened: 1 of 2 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
```diff
@@ -279,8 +281,54 @@ def _repr_dict(self, obj, level, prefix, suffix,
...
+                else:
+                    yield self._convert_to_unicode_or_bytes_repr(repr(obj))
+                return
+
+            limit_inner = self.maxother_inner
+            limit_outer = self.maxother_outer
+            limit = limit_inner if level > 0 else limit_outer
+            if len(obj) <= limit:
+                # Note that we check the limit before doing the repr, so, this could account
+                # for up to 3 chars above the limit afterwards. i.e.: u'' or b''
```
int19h, line 298: It can be more than 3 if some characters inside the literal will get escaped. So the most pathological case is a Unicode string where every char is `\U0010ffff`. I don't think we need to worry about that, but it might be worth noting this in the comment so that it doesn't mislead in the future.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
```diff
@@ -279,8 +281,54 @@ def _repr_dict(self, obj, level, prefix, suffix,
...
     def _repr_str(self, obj, level):
-        return self._repr_obj(obj, level,
-                              self.maxstring_inner, self.maxstring_outer)
+        try:
+            if self.raw_value:
+                # For raw value retrieval, ignore all limits.
+                if isinstance(obj, bytes):
+                    yield obj.decode('latin-1')
+                else:
+                    yield self._convert_to_unicode_or_bytes_repr(repr(obj))
```
int19h, line 290: The semantics of "raw value" in VS is that it is *not* a repr - i.e. it shouldn't have any quotes around it, escaping etc.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
```diff
@@ -279,8 +281,54 @@ def _repr_dict(self, obj, level, prefix, suffix,
...
+            # up to 6 characters longer than limit. If you need precise formatting,
+            # you are using the wrong class.
+            left_count, right_count = max(1, int(2 * limit / 3)), max(1, int(limit / 3))  # noqa
+
+            # Important: only do repr after slicing to avoid duplicating a byte array that could be
+            # huge.
+
+            # Note: we don't deal with high surrogates here because we're not dealing with the
+            # repr() of a random object (which we have to convert to unicode to show as unicode
+            # when needed), we just show the unicode/string/bytes repr() directly to make clear
```
int19h, line 312: On Python 2, can't Unicode strings have surrogate pairs in them? So then the slices can potentially end up splitting the pair...

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
fabioz: Done

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
fabioz: Done (just returning the object directly in this case).

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
fabioz: It may split the pair, but as we do a `repr` afterwards, that's ok (i.e.: it's different from doing a `repr` on a random object where the returned string won't have an additional `repr` done it in -- in which case we do have to deal with high surrogates).

fabioz:
Note: MacOS failures seem to be unrelated to the change on the ci.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py
```diff
@@ -279,8 +281,61 @@ def _repr_dict(self, obj, level, prefix, suffix,
         yield suffix
 
     def _repr_str(self, obj, level):
-        return self._repr_obj(obj, level,
-                              self.maxstring_inner, self.maxstring_outer)
+        try:
+            if self.raw_value:
+                # For raw value retrieval, ignore all limits.
+                if isinstance(obj, bytes):
+                    yield obj.decode('latin-1')
```
karthiknadig, line 288: @int19h should this be using encoding from `getpreferredencoding`?

PR #271, opened 2020-05-29
Handle userUnhandled exception breakpoints. Fixes #111

### LINKED ISSUE #111: Break debugger on exceptions that aren't handled inside a pytest test
When running "Debug Test" from the code edit view, the test output pane shows me pytest's error message when it encounters an uncaught exception, but I would like it to launch a debugger instead. I currently have to insert a try/catch with a `breakpoint()` otherwise.

I actually feel this should have been requested already but could find neither an issue nor a setting for this...

fabioz:


fabioz:
As a note, Py35 was cancelled with 15 minutes, but a full run on Py36 took 14, so, maybe the timeout should be raised a bit.

PR #313, opened 2020-06-25
Fix #239: Timed out waiting for debugger to spawn

### LINKED ISSUE #239: Timed out waiting for debugger to spawn
Hello
I'm using the debugger with python and it gives me this wired message
"timed out waiting for debugger to spawn "
I'm using python 3.8.2 and version of vscode is 1.45

int19h:
Make timeouts for process spawning and exiting configurable, and increase the default value for spawning.

### DIFF
--- a/src/debugpy/common/__init__.py
+++ b/src/debugpy/common/__init__.py
@@ -4,5 +4,13 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import os
+
 
 __all__ = []
+
+# The lower time bound for assuming that the process hasn't spawned successfully.
+PROCESS_SPAWN_TIMEOUT = os.getenv("DEBUGPY_PROCESS_SPAWN_TIMEOUT", 15)
+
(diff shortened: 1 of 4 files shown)


src/debugpy/common/__init__.py
```diff
@@ -4,5 +4,13 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import os
+
 
 __all__ = []
+
+# The lower time bound for assuming that the process hasn't spawned successfully.
```
karthiknadig, line 12: Should we mention the units (seconds) for this? or is this common enough that people will know.

int19h: All time spans are customarily in seconds in Python, so I think the comment is descriptive enough. I'm hoping that new defaults are "good enough" for all practical uses, so users wouldn't really need to be aware of those environment variables - I added them there mostly to provide a workaround if it turns out to be wrong, but I don't think we should document this or treat this as a stable API.

PR #336, opened 2020-07-16
Improve dealing with blocking evaluate requests. Fixes #157

### LINKED ISSUE #157: Improve dealing with blocking evaluate requests
## Environment data
- PTVSD version: 4.3.2
- OS and version: MacOS and CentOS
- Python version: 3.6 and 3.7
- Using VS Code or Visual Studio: VS Code, and interacting with ptvsd directly

## Actual behavior
When justMyCode is set to false, operations on a spark dataframe that has been constructed in python using `session.createDataFrame` will hang if run in the debugger repl, but not if run normally. Spark appears to hang, printing `[Stage 0:>           (0 + 1) / 1]`.

## Expected  behavior
The code should not hang, or it should not matter what the state of justMyCode is.

## Steps to reproduce:
With pyspark installed (I have the latest spark 3.0 preview build), I can reproduce it by trying to debug the following script with `"justMyCode": false` in launch.json.
```python
from pyspark.sql import Row, SparkSession

def my_compute_function(session):
    row1 = Row(id='1', name='name')
    df = session.createDataFrame([row1])
    df.show() # set a breakpoint on this or the next line, and evaluate `df.show()` in the repl when it breaks
    return df


session = SparkSession.builder.master("local").appName("df-show").getOrCreate()
my_compute_function(session)
```

In ptvsd 4.2.10, it also hangs if you step over the `df.show()` line, but that was fixed by moving to 4.3.2. I guess that is because of https://github.com/microsoft/ptvsd/issues/1372, although I haven't verified that.

fabioz:
Sorry for the really big pull request... this issue did end up touching much more than I expected.


src/debugpy/adapter/clients.py
```diff
@@ -12,6 +12,7 @@
 from debugpy.common import compat, fmt, json, log, messaging, sockets
 from debugpy.common.compat import unicode
 from debugpy.adapter import components, servers, sessions
+from debugpy.common.messaging import NO_RESPONSE
```
int19h, line 15: We have a general policy for debugpy code to avoid imports of non-modules - they make things slightly worse with circular dependencies on 2.7, and break stuff that gets mutated by the module after it gets imported.

(`debugpy.common.compat` has a special dispensation for members that are intended to transparently override builtins, so it's okay to import `unicode`, `xrange` etc from it.)

src/debugpy/adapter/clients.py
```diff
@@ -511,6 +515,11 @@ def configurationDone_request(self, request):
             if conn.server is None and conn.ppid == self.session.pid:
                 self.notify_of_subprocess(conn)
 
+    @message_handler
+    def evaluate_request(self, request):
+        return self.server.channel.delegate_async(request)
+        return NO_RESPONSE
```
int19h, line 521: I think `OutgoingRequest.on_response()` should be able to solve this without a helper to spawn additional threads etc. It'll invoke the callback on a background thread dedicated to message handlers, with response as the argument. 

So, something like:
```py
    @message_handler
    def evaluate_request(self, request):
        propagated_request = self.propagate(request)

        @propagated_request.on_response
        def handle_response(response):
             
             request.respond(response.body)  # this will also do the right thing if body is Exception
               
        return messaging.NO_RESPONSE
```
(The callback is also guaranteed to be called if the server disconnects before responding - `response.body` will be `messaging.NoMoreMessages` in that case, which should still work fine with the above code.)

tests/debugpy/test_evaluate.py
```diff
@@ -677,3 +677,88 @@ def code_to_debug():
...
+        would be locked until secondary threads start running.
+        See: https://github.com/microsoft/debugpy/issues/157
+        """
+
+        import threading
+
+        try:
+            from queue import Queue
+        except:
+            from Queue import Queue
```
int19h, line 700: This is handled by `compat` - it exposes it as `queue`. Since this is an stdlib module, it's okay to use it unqualified after doing `import queue from debugpy.common.compat`.

tests/debugpy/test_evaluate.py
```diff
@@ -677,3 +677,88 @@ def code_to_debug():
...
+
+        class EchoThread(threading.Thread):
```
int19h, line 702: Tests get a handy helper to reduce verbiage like this - take a look at the use of `@daemon` in `test_timeline.py`. Aside from conciseness, it also checks that the thread doesn't outlive the test in which it was spawned.

tests/debugpy/test_evaluate.py
```diff
@@ -677,3 +677,88 @@ def code_to_debug():
...
+            processor.process(1)
+            processor.process(2)  # @bp
+            processor.process(3)
+            processor.finish()
+
+    with debug.Session() as session:
+
+        # During the evaluation we'll actually have continued/stopped events because
+        # we're letting threads run at that time. Let's ignore these in the test.
+        session.ignore_unobserved.extend([timeline.Event("stopped")])
```
int19h, line 748: I wonder how many other tests would be potentially affected by this (e.g. by helper threads spawned by stdlib)? If it's common, we might need to add it to the default exclusion list in session.py.

tests/debugpy/test_evaluate.py
int19h: Oh, nevermind, I missed the fact that this is inside debuggee, so you don't get the fixtures.

tests/debugpy/test_evaluate.py
```diff
@@ -677,3 +677,88 @@ def code_to_debug():
...
+        The idea here is that a secondary thread does the processing of instructions,
+        so, when all threads are stopped, doing an evaluation for:
+
+        processor.process('xxx')
+
+        would be locked until secondary threads start running.
+        See: https://github.com/microsoft/debugpy/issues/157
+        """
+
+        import threading
```
int19h, line 695: Needs to do:
```py
import debuggee
debuggee.setup()
```
to work with all the runners (`attach_pid`, `attach_listen`, and `attach_connect` all rely on this). It's is the standard scaaffolding for debuggee code, and should only be skipped when the test is doing some kind of highly custom launch or attach sequence all by itself (i.e. without using a runner).

tests/debugpy/test_evaluate.py
fabioz: I'm not sure... I think something like this is pretty rare, so, I think adding it per test makes sense.

PR #353, opened 2020-07-31
Fix issue with frame eval mode and multiple breakpoints in generator. Fixes #348

### LINKED ISSUE #348: Code work well but debugger returns a mistake.
## Environment data

-   VS Code version: 1.47.2
-   Extension version (available under the Extensions sidebar): v2020.7.96456
-   OS and version: windows10 1803
-   Python version (& distribution if applicable, e.g. Anaconda): 3.7.0 32-bit
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
-   Relevant/affected Python packages and their versions: N/A
-   Relevant/affected Python-related VS Code extensions and their versions: N/A
-   Value of the `python.languageServer` setting: "Microsoft"

## Expected behaviour

This code below is worked properly.

```py
a = [1, 2]
def b():
    yield from [j for j in a if j % 2 == 0]
for j in b():
    print(j)
```

## Actual behaviour

I add break points at each line,and debug it, then it returns 

image

## Steps to reproduce:

[**NOTE**: Self-contained, minimal reproducing code samples are **extremely** helpful and will expedite addressing your issue]

1. Add break points at each line and then debug.

> I guess the reason is

image

this `list_iterator object` is vanished at the last time. But debugger tried to visit it, and throw out a mistake.

## Logs

<details>

<summary>Output for <code>Python</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

fabioz:


src/debugpy/_vendored/pydevd/_pydev_runfiles/pydev_runfiles_pytest2.py
```diff
@@ -164,10 +164,16 @@ def pytest_collection_modifyitems(session, config, items):
 
 def _get_error_contents_from_report(report):
     if report.longrepr is not None:
-        tw = TerminalWriter(stringio=True)
+        try:
```
fabioz, line 167: Note: this is actually unrelated (just synchronizing latest pydevd changes).

src/debugpy/_vendored/pydevd/third_party/cython_json.py
```diff
@@ -8,61 +8,88 @@
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
+# Note: Cython has some recursive structures in some classes, so, parsing only what we really
```
fabioz, line 11: Note: this is actually unrelated (just synchronizing latest pydevd changes).

src/debugpy/_vendored/pydevd/_pydevd_frame_eval/pydevd_modify_bytecode.py
```diff
@@ -3,11 +3,44 @@
 from types import CodeType
 from _pydev_bundle import pydev_log
 from _pydevd_bundle.pydevd_constants import IS_PY38_OR_GREATER
+from os.path import os
```
int19h, line 6: This looks like it should just be `import os`?

Issue #334, opened 2020-07-11
Adapter times out when running from Maya 


luabud:
```
import cube
# Error: RuntimeError: file C:\Program Files\Autodesk\Maya2020\Python\lib\site-packages\debugpy\server\api.py line 224: timed out waiting for adapter to connect # 
```

cube.py:
```py
import maya.cmds as mc
import debugpy 
import time

#while debugpy.is_client_connected():
#    time.sleep(0.5)

debugpy.listen(("localhost",5678))
debugpy.wait_for_client()

debugpy.breakpoint()
r = mc.polyCube(w=2,h=9,d=2)

debugpy.breakpoint()
print("hey!")
```

Run selection for "import cube"

fabioz:
I don't really have Maya to test and it seems to work for me in the usual attach to port `5678`... Do you know which Python version is Maya using?

One thing to try: replace `localhost` with `127.0.0.1` (both in the launch and in `debugpy.listen`).

If it doesn't work, please provide the logs (from Maya and from VSCode).

To get the logs from Maya, you should be able to do something as:

```
import debugpy
debugpy.log_to('c:/temp')
```

To get the logs from VSCode:

* Open VS Code
* Select the command Extensions: Open Extensions Folder
* Locate the Python extension directory, typically of the form ms-python.python-2020.*.****
* In that directory ensure you do not have any debug*.log files, if you do, please delete them
* Go back into VS Code and modify your launch.json to add the setting "logToFile": true to the attach, see below:

```
"version": "0.2.0",
"configurations": [
        {
            "name": "Python Attach to 5678",
            "type": "python",
            "request": "attach",
            "port": 5678,
            "host": "127.0.0.1",
            "logToFile": true,
        },```

* Start debugging
* When done, go back into the extension directory and upload the debug*.log files into this GitHub issue.

int19h:
We had a collaborative debug session about this, and it's pretty complicated (there are various things that seem to not work in Maya). I'll investigate it further.

fabioz:
As a note, I believe that plain `pydevd` does work in `Maya` using the remote attach (from reports I got in the past), so, my hunch is that this is related to launching the secondary adapter process/connecting to it, but if it turns out you need something on the `pydevd` side, please let me know.

luabud:
FWIW Maya uses Python 2.7 😅 and I manually added debugpy to site-packages because they seem to use some sort of custom Python.

munkybuttballs:
Are there any updates on this?
[...]

Issue #262, opened 2020-05-25
Problem running debugpy in embedded python interpreter


nicelifeBS:
## Environment data

- debugpy version: 1.0.0b9
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): Blender 2.82, Nuke 11, Maya2018
- Using VS Code or Visual Studio: 1.45.0

## Actual behavior

Remote debugging works running it in a python environment in a shell. But when run through an application with embedded python interpreter it starts a new instance of this application and it cannot connect.

I think the problem is that `sys.executable` is not returning the path to the python interpreter but the path of the application executable itself. And because of this the arguments passed with subprocess are not recognized.

https://github.com/microsoft/debugpy/blob/7a2891b95054f132435585f6ab5dedf66be4f5ab/src/debugpy/server/api.py#L103

If I replace `sys.executable` with the path to the embedded interpreter remote debugging works as expected.

An older version of ptvsd worked out of the box btw: https://github.com/microsoft/ptvsd/commit/0825dbeb3758a3839bdc1f2fd870418191f2845b

tanant:
dumb question - do we need the _specific_ sys.executable that the process/app running is using, or for the purposes of this, do we just need _a_ valid python interpreter to mediate the comms back and forth?

karthiknadig:
@tenant In the case mentioned above it is needed for comms. For the above case, attach using process id would probably work. Essentially start debug adapter in server mode.
```
python  /path/to/debugpy/adapter --host 0.0.0.0 --port 5678
```
Then from VSCode, use the following to connect:
```
        {
            "name": "Remote Attach",
            "type": "python",
            "request": "attach",
            "debugServer": 5678,
            "processId": "", // <---- put process id of the process that has the embedded python
            "pathMappings": [{
                "localRoot": "${workspaceFolder}",
                "remoteRoot": "."
            }]
        }
```
This way you will use the python executable for comms with VS Code, and it should be able to load the debugger where the embedded python. This uses a different pathway where we don't try to start the python executable.

puremourning:
> I think the problem is that sys.executable is not returning the path to the python interpreter but the path of the application executable itself. And because of this the arguments passed with subprocess are not recognized.


I found the same thing. This was my solution. It's absolutely not art: https://github.com/puremourning/vimspector/blob/master/python3/vimspector/developer.py#L23

fabioz:
[...]

PR #387, opened 2020-08-25
Makes the listen process configurable.

bob-white:
Basic idea for making the listen process configurable for applications
that embed python, but aren't python processes. Can help with #334


int19h:
We already have a config property for this - it's "python" (with "pythonPath" as legacy spelling), it's just that it's currently only used for launch configurations, but we should repurpose that for attach. It also needs a test. 

Also, per https://github.com/microsoft/debugpy/issues/262#issuecomment-644915096, a useful heuristic is to probe `sys.exec_prefix`.

bob-white:
Interesting, didn't see the other config, and from looking at the listen code, it would get rejected if I tried to use it with `debugpy.configure` as it doesn't pre-exist in the `_config` dictionary.

So `python` would be a better spelling than `process`?

Also for reference in our world editor:
```python
>>> sys.exec_prefix
''
```
So not sure what exactly we'd get out of that in this case.

int19h:
The stuff in `_config` is properties that must be set using `debugpy.configure()` when setting up for attach (i.e. that can't go into launch.json or equivalent on client side, because by the time the debugger sees that, it's already too late). So "python" isn't there, because it doesn't have any meaning for attach currently. But since it already exists for launch, and its meaning there is to specify the path to Python interpreter used to perform the launch, I think it makes sense to repurpose it for this. 

As far as `exec_prefix` - if I understand @dalmago correctly, it is set for Blender and Maya, so for those cases having it would mean that users don't have to configure anything manually.

bob-white:
I think they'd still need to configure _something_ with `sys.exec_prefix` for example Maya doesn't ship with a `python` executable, instead it has `maya` and `mayapy`.  3dsMax and Houdini would have similar problems as they also ship with standalone python interpreters that aren't named python. (`3dsMaxpy` and `hython` respectively.)

Blender seems to be the oddly well behaved one out of the bunch because for it, `os.path.join(sys.exec_prefix, 'python')` actually works.

So something almost like this would need to be done.
```python
    adapter_python = _config.get('python', sys.executable)
    if not os.path.exists(adapter_python):
        adapter_python = os.path.join(sys.exec_prefix, adapter_python)
```
This sadly doesn't work if you're just saying "use the `python` executable I have on `PATH`"
So maybe attempt to launch the adapter a few times with variations of the input until one works or they all fail?

int19h:
If "python" is specified explicitly, it should just be used with no modification, without trying to exists-check it.

But for the cases where it's not specified, the default could be `sys.executable` if it exists, otherwise `exec_prefix`. Although I'm not sure that would also work reliably, since `sys.executable` isn't necessarily missing - it's just not Python. Let's keep it simple and stick to `sys.executable` for now, then.

bob-white:
Okay, that feels better to me as well.
The thing with `sys.exec_prefix` is you'd still need to feed it _something_ to complete the path, otherwise its just a folder that the python process probably lives in.

So one other question, when it comes to adding a test for this, what is the best way to go about that? I've never worked with `pytest` before, and I couldn't find any other tests that were working `debugpy.configure` to use as a starting point.

int19h:
I've just realized that this would be very tricky to test, because it's a process used by the debug server to spawn the debug adapter, and the communication between the two is an implementation detail from the perspective of our (black box) tests. 

So, to observe the difference, the test would need to provide a wrapper around the actual Python interpreter that just registers the fact that it was invoked. Which is doable in principle - but it can't be an actual binary for the tests to remain platform-independent. We could probably rig something reasonably portable up with `cython --embed`, but it's definitely not simple. 

I'll think some more about it, but we can do the tests in another PR, so as not to block this one.

int19h:
Is the PR ready for merging, or are you planning on adding any more commits?

PR #393, opened 2020-08-28
Update TPN and add cgmenifest

karthiknadig:


cgmanifest.json
```diff
@@ -0,0 +1,54 @@
...
+      "Component": {
+        "Type": "pip",
+        "pip": { "Name": "pycodestyle", "Version": "2.3.1" }
+      },
+      "DevelopmentDependency": false
+    },
+    {
+      "Component": {
+        "Type": "pip",
+        "pip": { "Name": "pydevd", "Version": "1.9.2" }
```
fabioz, line 27: We have to decide how to work if the pydevd version will be kept in sync (in reality the version in pydevd and debugpy would change at each commit as changes are applied in-sync to both repos at the same time and the diff is reviewed prior to inclusion in debugpy -- updating at each commit would be too slow due to the review policies in that case, but I guess that I can create a commit just to increase the pydevd version at the debugpy repo when a release is done).

cgmanifest.json
```diff
@@ -0,0 +1,54 @@
...
+        "git": {
+          "Name": "bytecode",
+          "RepositoryUrl": "https://github.com/vstinner/bytecode",
+          "CommitHash": "e3e77fb690ed05ac171e15694e1c5d0e0dc34e86"
+        },
+        "DevelopmentDependency": false
+      }
+    }
+  ]
+}
```
fabioz, line 54: I'm not sure about the things for which we have just some code copied and pasted.

i.e.:

There are some files extracted from python itself for backward compatibility in:
- _pydev_imps
- _pydevd_bundle\pydevconsole_code_for_ironpython.py
- third_party\pep8\lib2to3

There are some things from IPython -- see:
- pydev_ipython\README

There's a copy of the debug adapter protocol (it's MS copyright, so, maybe nothing needs to be added).
- _pydevd_bundle\_debug_adapter\debugProtocol.json

karthiknadig: For these (Python and IPyhton) there should be a license entry in the https://github.com/microsoft/debugpy/blob/master/src/debugpy/ThirdPartyNotices.txt . Looks like it does have these covered. They won't be picked up automatically. 

We may have to add the DAP license.

cgmanifest.json
karthiknadig: The way we are following now still works. The review process for this has changed a bit. If there are new components "vendored", then we will need to do a review where we may have to go through code scanning (which takes time). Otherwise, we can rely on the automated processing, that should not take time for PRs.

fabioz (approved):

PR #406, opened 2020-09-06
More compiler tweaks

karthiknadig:


src/debugpy/_vendored/pydevd/pydevd_attach_to_process/windows/run_code_on_dllmain.cpp
```diff
@@ -14,6 +14,10 @@ HINSTANCE globalDllInstance = NULL;
 
 class NotificationHelper {
 public:
+#pragma warning( push )
+#pragma warning( disable : 4722 )
+// disable c4722 here: Destructor newver returns warning. Compiler sees ExitThread and assumes that
```
fabioz, line 19: Minor: typo: newver -> never.

fabioz (approved):

src/debugpy/_vendored/pydevd/pydevd_attach_to_process/windows/run_code_on_dllmain.cpp
karthiknadig, line 19:
```suggestion
// disable c4722 here: Destructor never returns warning. Compiler sees ExitThread and assumes that
```

PR #432, opened 2020-10-08
Check if function return is None to consider exception unhandled on return. Fixes #404

### LINKED ISSUE #404: Debugger does not recognize that exception is handled in single line except statement scenario
## Environment data

- debugpy version: 1.0.0rc2 (run `import debugpy; print(debugpy.__version__)` if uncertain)
- OS and version: Windows
- Python version (& distribution if applicable, e.g. Anaconda): 3.7 64-bit
- Using VS Code or Visual Studio: Visual Studio

## Actual behavior

Debugger breaks on exception when it shouldn't given the exception is handled. Break on thrown is 'off'.

## Expected  behavior

Debugger does not break on exception

## Steps to reproduce:
1. pip install `beautifulsoup4` and `lxml` in Python 3.7
2. debug this code:

```
from bs4 import BeautifulSoup

def has_cost(tag):
    try: class_names = tag.attrs['class']
    except: return False
    return True

shtml = "<html><head><title>The Dormouse's story</title></head>"
soup = BeautifulSoup(shtml, 'lxml')
soup_row_cost_arr = soup.find_all(has_cost)
print(soup_row_cost_arr)
```

result: breaks on exception thrown
image

change the implementation to have the return statement on its own line like this:
```
def has_cost(tag):
    try: class_names = tag.attrs['class']
    except: 
        return False
    return True
```

result: does not break, it correctly recognizes that it is handled

exception settings:
image

logs for both scenarios:
[one-liner-breaks.txt](https://github.com/microsoft/debugpy/files/5177070/one-liner-breaks.txt)
[two-liner-no-break.txt](https://github.com/microsoft/debugpy/files/5177072/two-liner-no-break.txt)

from developer community feedback at https://developercommunity.visualstudio.com/content/problem/1158181/python-exception-in-try-block-stops-in-debugger.html?childToView=1173634

fabioz:


src/debugpy/_vendored/pydevd/_pydev_runfiles/pydev_runfiles_unittest.py
```diff
@@ -85,9 +85,17 @@ def stopTest(self, test):
         error_contents = ''
         test_name = self.get_test_name(test)
 
-
         diff_time = '%.2f' % (end_time - self.start_time)
-        if not self._current_errors_stack and not self._current_failures_stack:
+
```
fabioz, line 89: This is unrelated (just synchronizing with pydevd).

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.pyx
```diff
@@ -284,15 +281,15 @@ cdef class PyDBFrame:
...
 
             if should_stop:
-                if self.handle_exception(frame, event, arg, EXCEPTION_TYPE_HANDLED):
+                if self._handle_exception(frame, event, arg, EXCEPTION_TYPE_HANDLED):
                     return self.trace_dispatch
 
         elif event == 'return':
             exc_info = self.exc_info
-            if exc_info:
+            if exc_info and arg is None:
```
fabioz, line 292: This is the actual fix.

It's not perfect because if the user has a return in the same line as the except

i.e.:

```
try:
   ...
except: return
```

The debugger will still think that it's exiting due to an exception.

I couldn't think of a better workaround (it's something that probably should be fixed in CPython itself -- i.e.: https://bugs.python.org/issue34099 -- maybe I could take a look at doing this change in CPython myself in the future).

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.pyx
```diff
@@ -560,7 +557,12 @@ cdef class PyDBFrame:
 
         return stopped
 
-    def get_func_name(self, frame):
+    # IFDEF CYTHON -- DONT EDIT THIS FILE (it is automatically generated)
```
fabioz, line 560: Note: while reviewing the code in this module I ended up doing a bunch of minor cython improvements.

int19h (approved):

karthiknadig (approved):

PR #436, opened 2020-10-08
Update binary dependencies to support Python 3.9. Fixes #384

### LINKED ISSUE #384: Python 3.9 support
There was some preliminary effort to support 3.9, so, we know that the basics work, but features which require more internal knowledge still need more work.

i.e.:
- attach to process
- frame eval mode

Now that 3.9rc1 is out, this may be a good time to tackle it.

fabioz:


fabioz:
@karthiknadig @int19h it seems like tests aren't running (the last test run in https://dev.azure.com/debugpy/debugpy/_build?definitionId=1&_a=summary seems to be from September 30th). Do you know why this is happening?

karthiknadig:
This is broken for most of the repos. Looks like we may have to manually trigger the runs for now.

int19h (approved):

karthiknadig (approved):

Issue #362, opened 2020-08-11
Debugger times out when "qt":"auto" is used with ArcGIS Python


eykamp:
## Environment data

- PTVSD version: XXX <-- Not sure, how do I tell?
- OS and version: Win 10 Enterprise, v. 1709
- Python version (& distribution if applicable, e.g. Anaconda):  Python 3.6.9 , Anaconda (installed by ArcGis)
- Using VS Code or Visual Studio: VS Code 1.47.3

## Actual behavior

When firing up the debugger (F5) even a simple one-line script (print("hello")), I get the dreaded "timed out waiting for debuggee to spawn" dialog.  I can run the script using the green triangle with the same version/environment, and if I select a different version of Python, the debugger works fine.

Everything consistently worked for months until this morning when VSCode updated to the June version, and debugging stopped working.

## Expected  behavior

Debugging should work.

## Steps to reproduce:

Press F5 to debug current program.

I've read some of the past cases for this issue, and followed some of the suggestions, but they didn't really seem relevant.  I'm not even sure this is a ptvsd issue, but all searches on the error message led me here.

int19h:
What's the command line like in the terminal?

If you're using `"console": "internalConsole"` in launch.json, change to `"console": "integratedTerminal"` and try like that.

eykamp:
cd c:\dev\NESDI && cmd /C ""C:\Program Files\ArcGIS\Pro\bin\Python\envs\dev\python.exe" c:\Users\eykampC\.vscode\extensions\ms-python.python-2020.7.96456\pythonFiles\lib\python\debugpy\launcher 49258 -- c:\dev\NESDI\delme.py "

int19h:
Those old issues from the ptvsd repo shouldn't be relevant here. But it probably does have something to do with ArcGIS Python. We'll need the logs to figure that out, though. Can you please add this:
```json5
"logToFile": true
```
to your debug config, and then look for `debugpy*.log` in `c:\Users\eykampC.vscode\extensions\ms-python.python-2020.7.96456`?

eykamp:
I figured those old issues were, well, old, but I looked through them just to make sure.

Here are the logs.

debugger.log:
```
0 Starting Session:
{
    "name": "Python: Current File (Integrated Terminal)",
    "type": "python",
    "request": "launch",
    "program": "c:\\dev\\PIMS - dev\\delme2.py",
    "logToFile": true,
    "console": "integratedTerminal",
[...]

PR #452, opened 2020-10-30
Native thread tracing in frame eval mode / disable qt monkeypatching by default.

fabioz:
This pull request does a few things:

1. It automatically creates a DummyThread for native threads and sets the tracing (called from the frame eval mode). This enables native threads to be traced by default when using the frame eval mode.

2. Disable the qt monkeypatching by default (it's still available to be used when frame eval is not available, but given that just importing qt seems to have side effects: https://github.com/microsoft/debugpy/issues/362, disabling by default is probably the proper thing to do anyways).

3. Use the internal version of the call tracing hook (this fixes the CPython issue: https://bugs.python.org/issue42197 for the debugger for all supported versions of CPython as the internal call tracing hook doesn't call the functions to update the locals) -- the debugger isn't as affected by this because it tries to not trace any context whenever possible, but when actually stepping or if the tracing must be kept on because exceptions are being tracked, this should be a noticeable performance improvement.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
@@ -916,7 +916,7 @@ struct __pyx_obj_14_pydevd_bundle_13pydevd_cython_SafeCallWrapper {
 };
 
 
-/* "_pydevd_bundle/pydevd_cython.pyx":1400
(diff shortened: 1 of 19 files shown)


int19h:
Nice! Yes, the Qt change should be okay, since it would only apply to 2.7 and 3.5.

int19h (approved):

PR #462, opened 2020-11-04
Re-compile binaries

karthiknadig:


int19h (approved):

fabioz:
@karthiknadig is there any particular reason to recompile? We could start to build them in the ci if needed...

karthiknadig:
@fabioz The current binaries were compiled using `/Qspectre` but did not use spectre mitigated libs. Going forward we need to build using `C++ x64/x86 Spectre-mitigated libs`.

fabioz (approved):

PR #467, opened 2020-11-10
Fix #463: Enable overriding "pythonPath" in the launcher

### LINKED ISSUE #463: Enable overriding "pythonPath" in the launcher
Problem: https://github.com/microsoft/vscode-docker/issues/2455#issuecomment-720651526

To solve this, the launcher needs to provide some way to override "pythonPath" that comes from the client (via the adapter), such that the custom launcher script that the Docker extension registers can change it.

int19h:
Add "debugLauncherPython" debug config property to customize Python interpreter used to spawn the launcher.


tests/debugpy/test_run.py
```diff
@@ -137,44 +137,90 @@ def code_to_debug():
...
+
+    expected = ""
+    if debuggee_custompy:
+        debuggee_custompy = make_custompy(tmpdir, "debuggee")
+        expected += "debuggee;"
+    if launcher_custompy:
+        launcher_custompy = make_custompy(tmpdir, "launcher")
+        expected += "launcher;"
+    else:
+        # If "python" is set, it also becomes the default for "debugLauncherPython"
```
bwateratmsft, line 185: I think a consequence of this is that the Docker extension has to pass in the Python interpreter path, which is not ideal. IMO if `debugLauncherPython` is unset, it should be the same as `debugAdapterPython`. In turn if that is unset, both should be same as the Python interpreter path. `python` should control only what the launcher uses, or otherwise I'd say add a new property so that only that one particular thing can be controlled.

bwateratmsft: Er, I didn't notice this was test code; but I'm unsure--which of the three properties does Docker need to set?

int19h: Keep in mind that this is going to work in conjunction with the Python extension in all VSCode scenarios. The extension sets all three properties to the current active interpreter for the workspace by default when resolving the configuration, skipping those that are already set. Thus, in your resolver, you would only need to set "python" to whatever it should be in the container, and let the extension provide the default for "debugLauncherPython". 

But this branch would kick in for other DAP clients, such as VS or Vimspector, and maintain existing behavior for them.

bwateratmsft: Sounds good, thanks!

karthiknadig (approved):

PR #476, opened 2020-11-20
Run pydevd tests on github actions. Fixes #471

### LINKED ISSUE #471: Run pydevd tests on github actions
Right now the pydevd tests (which are run independent on debugpy) run on travis.org, but the tests seem to take a long time to start now (it seems travis.org will be deprecated in favor of travis.com).

Given that, I think the best approach is actually migrating to Github actions (I'm creating the issue here to track that I plan to tackle this right away as this slows down creating pull requests for debugpy as I wait for pydevd tests to run before submitting it to debugpy).

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_frame_eval/pydevd_frame_evaluator.template.pyx
```diff
@@ -35,10 +36,7 @@ cdef class ThreadInfo:
     # code for a function is None and there are breakpoints.
     cdef public bint force_stay_in_untraced_mode
 
-    def __init__(self):
-        cdef PyFrameObject * frame_obj
-        frame_obj = PyEval_GetFrame()
-
+    cdef initialize(self, PyFrameObject * frame_obj):
```
fabioz, line 39: This is actually a bugfix for native thread tracing (on some situations it's possible that during trace-eval `PyEval_GetFrame()` only returns null even if a frame is being executed -- one of the existing qt tests failed during this particular scenario and started failing on Github actions while it worked on Travis).

src/debugpy/_vendored/pydevd/conftest.py
```diff
@@ -106,13 +106,41 @@ def check_no_threads():
     _start_monitoring_threads()
 
 
+from _pydevd_bundle.pydevd_utils import is_current_thread_main_thread
+import threading
+
+
+@pytest.yield_fixture(autouse=True)
+def check_main_thread(request):
+    was_main = is_current_thread_main_thread()
```
fabioz, line 115: I'm still debugging this (but I wanted to merge what I already have anyways as this is proving to be quite hard to track down).

This fails only under Windows (and only sometimes) -- it seems that threading.current_thread() is returning a thread which is not threading.main_thread(), which is not expected at this point.

I added this because sometimes `tests_python.test_utilities.test_is_main_thread` fails due to that, so, this is to diagnose when that happens.

int19h (approved):

PR #484, opened 2020-11-26
Sync with latest pydevd.

fabioz:


karthiknadig:
This should also be updated:
https://github.com/microsoft/debugpy/blob/main/cgmanifest.json#L27

fabioz:
> This should also be updated:
> https://github.com/microsoft/debugpy/blob/main/cgmanifest.json#L27

Just updated

karthiknadig (approved):

PR #485, opened 2020-11-26
Properly dealing with -X python vmargs. Fixes #455

### LINKED ISSUE #455: Execution Error using Subprocess and faulthandler
When executing a Popen subprocess for python with the -X faulthandler command line flag, if the -X flag is used first, the command that is generated is incorrect and triggers a error starting python.

```
unknown option --port
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
Press any key to continue . . .
```

Minimal reproduction:
```
import subprocess

if __name__ == "__main__":
    args = ['python', '-X', 'faulthandler', '-c', 'print("code ran!")']
    proc = subprocess.Popen(args)
```

Using process monitor, I was able to see the call generated by the faulty call above:
```
python -X c:\Users\user\.vscode\extensions\ms-python.python-2020.10.332292344\pythonFiles\lib\python\debugpy\_vendored\pydevd\pydevd.py --port 59163 --ppid 5936 --client 127.0.0.1 --client-access-token REDACTED --multiprocess --skip-notify-stdin --json-dap-http --file faulthandler -c "print(\"code ran!\")"
```

You can see that the -X flag is being inserted after the first python call in the chain, and the faulthandler parameter is way at the end.

Reordering the arguments list for Popen to put the -c flag first, and -X last, I get the following call generated:
```
python -c "import sys; sys.path.insert(0, r'c:\Users\user\.vscode\extensions\ms-python.python-2020.10.332292344\pythonFiles\lib\python\debugpy\_vendored\pydevd'); import pydevd; pydevd.PydevdCustomization.DEFAULT_PROTOCOL='http_json'; pydevd.settrace(host='127.0.0.1', port=59174, suspend=False, trace_only_current_thread=False, patch_multiprocessing=True, access_token=None, client_access_token='REDACTED', __setup_holder__={'client': '127.0.0.1', 'client-access-token': 'REDACTED', 'json-dap-http': True, 'multiprocess': True, 'port': 59174, 'ppid': 8688, 'server': False, 'skip-notify-stdin': True}); print(\"code ran!\")" -X faulthandler
```

I'm running the following:

Version: 1.50.0 (system setup)
Commit: 93c2f0fbf16c5a4b10e4d5f89737d9c2c25488a3
Date: 2020-10-07T06:10:52.432Z (3 wks ago)
Electron: 9.2.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.17763

ms-python.python v2020.10.332292344

fabioz:


fabioz:
@int19h @karthiknadig ping (waiting for the review on this pull request)

src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -249,99 +230,147 @@ def patch_args(args, is_exec=False):
...
 
-        if is_python(args[0]):
-            ind_c = get_c_option_index(args)
+        # Note: we create a copy as string to help with analyzing the arguments, but
+        # the final list should have items from the unquoted_args as they were initially.
+        args_as_str = _get_str_type_compatible('', unquoted_args)
 
-            if ind_c != -1:
-                host, port = _get_host_port()
+        params_with_flag = ('-W', '-X', '--check-hash-based-pycs', '--jit')
```
int19h, line 251: There's a catch with these:
https://github.com/microsoft/ptvsd/blob/46e1b758bc3694df7b4f08f24b62e73c59494f73/src/ptvsd/multiproc.py#L275-L283

fabioz: I'll take a look at this (quite tricky indeed, even something as `python -qmmodule_to_run` may be done, so, even `-m` and `-c` may be combined).

fabioz: Ok, there's now code to handle the combination of short flags.

int19h (approved):

PR #488, opened 2020-12-04
Support running non-ascii files on Python 2. Fixes #206

### LINKED ISSUE #206: IOError: [Errno 22] when running files with non-ascii characters.
## Environment data

- PTVSD version: master
- OS and version: windows
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.16
- Using VS Code or Visual Studio:

Additional info:
File System encoding: `mbcs`
Default encoding: `ascii`

## Actual behavior

This error is shown:
```
  File "c:/Users/kanadig/.vscode/extensions/ms-python.python-2019.5.17059/pythonFiles/ptvsd_launcher.py", line 43, in <module>
    main(ptvsdArgs)
  File "C:\GIT\ptvsd\src\ptvsd\__main__.py", line 434, in main
    run()
  File "C:\GIT\ptvsd\src\ptvsd\__main__.py", line 312, in run_file
    runpy.run_path(target, run_name='__main__')
  File "C:\Python27\lib\runpy.py", line 251, in run_path
    code = _get_code_from_file(path_name)
  File "C:\Python27\lib\runpy.py", line 227, in _get_code_from_file
    with open(fname, "rb") as f:
IOError: [Errno 22] invalid mode ('rb') or filename: 'c:\\GIT\\pyscratch2\\??\\experiment.py'
```
image

## Expected  behavior

Should launch the file, like in pyhton 3.*

## Steps to reproduce:
1. Create a file with path `C:\测试\experiment.py`
2. Open the directory in vscode
3. Start debugging.

fabioz:


tests/debugpy/test_run.py
```diff
@@ -284,3 +282,38 @@ def code_to_debug():
...
         using_frame_eval = backchannel.receive()
         assert using_frame_eval == (frame_eval == "yes")
+
+
+@pytest.mark.skipif(
+    not sys.version_info[0] >= 3,
+    reason="Test structure must still be updated to properly support Python 2 with unicode",
+)
+@pytest.mark.parametrize("run", [runners.all_launch[0]])
+def test_unicode_dir(tmpdir, run, target):
```
fabioz, line 292: Note: this works on Python 3 but fails on Python 2... in the tests there seems to be quite a bit of unicode/bytes mixing and it'll probably take a good while to sort it out, so, I decided to provide the pull request as is for now and leave the test only for Python 3.

I manually tested that it's possible to debug a file with unicode chars -- as long as they're compatible with the filesystem (unfortunately this means that we could break this again in the future, but given that Python 2 is becoming less relevant now that it's officially unsupported I'm not sure it's worth it doing the revamp on the test structure to stop mixing bytes and unicode in it).

src/debugpy/adapter/launchers.py
```diff
@@ -153,7 +153,7 @@ def on_launcher_connected(sock):
                 "env": env,
             }
             if cwd is not None:
-                request_args["cwd"] = cwd
+                request_args["cwd"] = compat.filename_str(cwd)
```
int19h, line 156: I think this only ever comes up as non-Unicode if it's not specified but "program" is, and so the client request handler defaults to `os.path.dirname(program)`. It would be better for that handler to do the proper conversion - since it knows its origin, and thus knows the encoding - such that in this function, the argument is always `unicode` on entry. Basically the general policy for filenames is: convert to/from Unicode as early as possible at the boundary with non-Unicode OS / Python APIs, so that all code that's not at the boundary can just assume it's all Unicode.

src/debugpy/server/cli.py
```diff
@@ -43,7 +43,7 @@ class Options(object):
     address = None
     log_to = None
     log_to_stderr = False
-    target = None
+    target = None  # Unicode on Py2/Str on Py3
```
int19h, line 46: Can just mark it as `unicode`, since we have that as an alias in `compat` on all Python versions.

src/debugpy/adapter/launchers.py
fabioz: OK, I'll take a look at it (as a note, I'm already finishing for the day here, so, I'll take a look at this next week).

src/debugpy/server/cli.py
```diff
@@ -294,14 +313,9 @@ def run_module():
...
     # if __package__ is Unicode.
-    target = (
-        compat.filename_bytes(options.target)
-        if sys.version_info < (3,)
-        else options.target
-    )
 
     log.describe_environment("Pre-launch environment:")
-    log.info("Running module {0!r}", target)
+    log.info("Running module {0!r}", options.target)  # Log must be unicode.
```
int19h, line 318: `{!r}` should take care of necessary quoting and escaping if it's a byte string, no? OTOH we can see the final encoding result that gets passed to `runpy`, which may be useful for diagnosing encoding issues.

tests/debugpy/test_run.py
int19h: Yeah, I don't think we can run such a test reliably on Windows anyway, since it would depend on user-global locale to be set to something that can actually round-trip this path. It would need a container that the test suite can fully isolate and control. This might come anyway for other reasons eventually, but it's definitely not worthwhile for a py2-only corner case.

int19h (approved):

src/debugpy/server/cli.py
fabioz: You're right that the `!r` makes it work (instead of throwing a `UnicodeDecodeError` as would be the case if it wasn't there), but I believe that the output is nicer if both are unicode in the case where the target actually has unicode characters.

src/debugpy/adapter/launchers.py
fabioz: I removed this... it actually didn't do anything.

On Python 2, everything in the `cmdline` and `cwd` are actually `str` (and not `unicode`).

i.e.: on `launch_request`, although the `args` and `cwd` are built from something as:

```
program = request("program", unicode)
args = [program]
cwd = request("cwd", unicode, optional=True)
``` 

Logging it with something as:
`log.info('program {0!r}({1}), cwd: {2!r}({3})', program, type(program), cwd, type(cwd))`

shows that the types are actually `str`:
`program 'C:/temp/รก/script.py'(<class 'str'>), cwd: 'X:/ptvsd_workspace/ptvsd'(<class 'str'>)`

-- note that the actual path I was debugging is: `C:\temp\á\script.py`, so, it seems to be encoded with `mbcs` already.

So, the `args` and `cwd` that actually arrive on `spawn_debuggee` are all `list[str]` (and the launch works due to that).

It seems that internally there are still things that are inconsistent on Python 2 (even though the launch works), but given that the code seems to work even with those inconsistencies and given that this is a Python 2 only issue, I'm leaving that untouched.

fabioz: Oops, my bad, it seems I got those logs on a python 3, not python 2... on Python 2 it does indeed show `unicode` (which seems to be correct).

Removing that line still works properly ;)

PR #498, opened 2020-12-11
Provide full string when evaluating clipboard. Fixes #110

### LINKED ISSUE #110: Clipboard context value for the "evaluate" request
DAP clients (frontends) use the `evaluate` request when copying variable and expression values to the clipboard. To help debug adapters detect this scenario, a new value `clipboard` has been added to the value set for the `context` argument that is passed to the `evaluate` request. To ensure backward compatibility, a client is only allowed to pass the new value if the debug adapter returns a `supportsClipboardContext` capability.

fabioz:


fabioz:
Note: I also updated the spec to the latest version here.

karthiknadig (approved):

PR #508, opened 2020-12-21
Add a setuptools entry point

gwerbin:
Add a Setuptools entry point, so that Debugpy can be invoked from the command line as:

```bash
debugpy           # Equivalent to: python -m debugpy
debugpy-adapter   # Equivalent to: python -m debugpy.adapter
debugpy-launcher  # Equivalent to: python -m debugpy.launcher
```

The main purpose of this change is so that Debugpy can be installed with [Pipx](https://pipxproject.github.io/pipx/). This enables Debugpy to be easily installed into a self-contained environment, for use with arbitrary clients (such as [Nvim-DAP](https://github.com/mfussenegger/nvim-dap)). Then it can be conveniently invoked in "adapter" mode as `debugpy-adapter`.

It's also (in my opinion) easier and more user-friendly to type `debugpy` instead of `python -m debugpy` on the command line :)

### DIFF
--- a/src/debugpy/__main__.py
+++ b/src/debugpy/__main__.py
@@ -43,8 +43,3 @@ if __name__ == "__main__":
     from debugpy.server import cli
 
     cli.main()
-else:
-    # Instead of hard-coding "debugpy = debugpy.server.cli:main" in setup.py, import it
-    # here. If the fully-qualified name of our `main()` function changes, we don't need
-    # to change it in setup.py, as long as we import it here as `cli.main()`.
-    from debugpy.server import cli
--- a/src/debugpy/__main__.py
+++ b/src/debugpy/__main__.py
@@ -43,3 +43,8 @@ if __name__ == "__main__":
     from debugpy.server import cli
 
     cli.main()
+else:
+    # Instead of hard-coding "debugpy = debugpy.server.cli:main" in setup.py, import it
+    # here. If the fully-qualified name of our `main()` function changes, we don't need
+    # to change it in setup.py, as long as we import it here as `cli.main()`.
+    from debugpy.server import cli
--- a/setup.py
+++ b/setup.py
@@ -134,5 +134,10 @@ if __name__ == "__main__":
             "debugpy._vendored": list(iter_vendored_files()),
         },
         cmdclass=cmds,
+        entry_points={
+            "console_scripts": [
+                "debugpy = debugpy.__main__:cli.main"
+            ]
+        },
         **extras
     )
--- a/setup.py
+++ b/setup.py
@@ -136,7 +136,9 @@ if __name__ == "__main__":
         cmdclass=cmds,
         entry_points={
             "console_scripts": [
-                "debugpy = debugpy.__main__:cli.main"
+                "debugpy = debugpy.__main__:cli.main",
+                "debugpy-adapter = debugpy.adapter.__main__:cli.main",
+                "debugpy-launcher = debugpy.launcher.__main__:cli.main",
             ]
         },
         **extras
--- a/setup.py
+++ b/setup.py
@@ -137,8 +137,8 @@ if __name__ == "__main__":
         entry_points={
             "console_scripts": [
                 "debugpy = debugpy.__main__:cli.main",
-                "debugpy-adapter = debugpy.adapter.__main__:cli.main",
-                "debugpy-launcher = debugpy.launcher.__main__:cli.main",
+                "debugpy-adapter = debugpy.adapter.__main__:cli",
+                "debugpy-launcher = debugpy.launcher.__main__:cli",
             ]
         },
         **extras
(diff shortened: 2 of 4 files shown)


src/debugpy/__main__.py
```diff
@@ -43,3 +43,8 @@
     from debugpy.server import cli
 
     cli.main()
+else:
+    # Instead of hard-coding "debugpy = debugpy.server.cli:main" in setup.py, import it
+    # here. If the fully-qualified name of our `main()` function changes, we don't need
+    # to change it in setup.py, as long as we import it here as `cli.main()`.
+    from debugpy.server import cli
```
int19h, line 50: It's okay to hardcode this. From the perspective of an outside client, `debugpy.server` is indeed an implementation detail; but since `setup.py` belongs to the same package, it can depend on it. While it's not a public API entrypoint by itself, we don't anticipate changing the actual function name or location, so another abstraction layer here is likely to be wasted effort.

gwerbin: Noted, I will clean this up.

int19h:
@gwerbin, My apologies for a belated follow-up!

debugpy.launcher is basically an internal artifact, and should be considered undocumented for third party use at the moment. We do provide some hooks to customize it, but those are also undocumented. This may change later if the design proves to be sufficiently stable; but for now, it should not be obviously exposed.

We discussed this change among the broader team, and the consensus was that we don't want to go down that route, because exposing the command directly as `debugpy` makes it confusing as to which version of Python is used to launch the debuggee in attach mode, in the event that you have more than one installed. It was noted that similar issues have already prompted some other tooling packages to drop entry points to prevent user confusion.

This reasoning is not quite applicable to debugpy.adapter, because it's specifically written in such a way that, for launch scenarios at least, the version of Python used to run the adapter doesn't have to match the version it uses to spawn the debuggee. But using debugpy.adapter directly is considered a highly advanced scenario - we don't even document its CLI, except that running it without arguments gives you a DAP channel over stdin/stdout. So an entrypoint for it alone doesn't seem to make much sense.

gwerbin:
> We discussed this change among the broader team, and the consensus was that we don't want to go down that route, because exposing the command directly as `debugpy` makes it confusing as to which version of Python is used to launch the debuggee in attach mode, in the event that you have more than one installed. It was noted that similar issues have already prompted some other tooling packages to drop entry points to prevent user confusion.
>
> This reasoning is not quite applicable to debugpy.adapter, because it's specifically written in such a way that, for launch scenarios at least, the version of Python used to run the adapter doesn't have to match the version it uses to spawn the debuggee. But using debugpy.adapter directly is considered a highly advanced scenario - we don't even document its CLI, except that running it without arguments gives you a DAP channel over stdin/stdout. So an entrypoint for it alone doesn't seem to make much sense.

Thanks for the clarification. Maybe my desire to have this feature stems from my lack of understanding of how Debugpy works.

I was hoping to be able to install a single instance of Debugpy for my user profile with Pipx, because I was under the impression that the Python instance that runs Debugpy is not necessarily the same Python instance that is executing the code being debugged (is that the "debugee?").

I know that e.g. the Jedi Language Server (specifically, the Jedi server that it wraps) looks for the `VIRTUAL_ENV` environment variable in order to determine what Python is currently "in use" by the developer. I was under the impression that Debugpy worked similarly. Is that not the case?

If that *is* the case, my hope was that this patch would make it easy to use the Pipx setup, without requesting upstream changes to Pipx itself. Although it would arguably be a useful Pipx feature anyway, to allow the user to create their own entry points on top of (or instead of) what the distribution defines for itself in `setup.py`.

int19h:
Yep, the "debuggee" is the process being debugged.

Which Python is used to spawn the debuggee depends on the scenario. If you are running `debugpy.adapter`, and then giving it the "launch" request, it will respect the "python" setting in the debug configuration - thus, it is entirely possible for the adapter and the debuggee to be using different Python versions.

However, if you're starting the debugpy server (i.e. `debugpy --listen`  or `--connect`), it has to spawn the debuggee before it receives the debug configuration. Its behavior in that case is intentionally as close to stock Python as possible, i.e. something like:
```sh
python debugpy --listen ... foo.py ...
```
should behave exactly like:
```sh
python foo.py ...
```
except that there's a debug running server in the resulting process. Thus, it uses the same Python version (or rather, the `debugpy` process is the same one that's then running user code).

This attach server scenario doesn't really have a direct equivalent with the language server. The one that does is adapter / "launch", and so it sounds like that's really the one you'd want the entrypoint for?

gwerbin:
Thanks for the explanation!

> This attach server scenario doesn't really have a direct equivalent with the langauage server. The one that does is adapter / "launch", and so it sounds like that's really the one you'd want the entrypoint for?

Yes, I believe this is what I want.

1. I install Debugpy in a standalone env using Pipx, which (currently) requires an entry point and does not work with `-m`.
2. My code editor invokes the "debugee-independent" adapter from the standalone installation.
3. My code editor tells the adapter which debugee Python to launch.

Note that there are some open issues in the Pipx project which would obviate the need to add an explicit entry point: https://github.com/pipxproject/pipx/issues/177 and https://github.com/pipxproject/pipx/issues/600.

PR #522, opened 2021-01-07
Show full stack trace on exception with __cause__ or __context__. Fixes #391

### LINKED ISSUE #391: Support exception chains in debugger
Hi, I may be doing something wrong, but it doesn't seem like the python debugger supports error chaining? For example:

image
The only error information shown is the outer exception, losing track of the original cause.

Error chaining works if run without the debugger (in a terminal):
image

Error Chaining (PEP 3134) Spec: https://www.python.org/dev/peps/pep-3134/

fabioz:
Note: the result of this pull request is that the reason is shown in the stack trace:

image

Unfortunately it seems that the lines aren't properly linked up to the actual line there (so, the user can click the link and it opens the file, but it should actually hyperlink the line part too so that it goes to the proper line, but there doesn't seem to be a way to do that from the DAP).


int19h:
I believe the hyperlinks are based on matching a certain pattern in the output. It might work as intended if it uses a colon, i.e. "...\file.py:123". There might be some VSCode API to customize it, too.

Anyway, I think we can treat that as a separate feature.

int19h (approved):

PR #529, opened 2021-01-15
Smaller Linux .so. Fixes #169

### LINKED ISSUE #169: pydevd.so and frame_eval.so on linux are quite large compared to mac
See if symbols are getting included in the .so files while building wheels.

fabioz:


src/debugpy/_vendored/pydevd/setup_cython.py
```diff
@@ -161,12 +161,43 @@ def build_extension(dir_name, extension_name, target_pydevd_name, force_cython,
...
-                                 # extra_compile_args=["-Zi", "/Od"],
-                                 # extra_link_args=["-debug"],
-                                 )]
+        extra_compile_args = []
+        extra_link_args = []
+
+        if 'linux' in sys.platform:
+            # Enabling -flto brings executable from 4MB to 0.56MB and -Os to 0.41MB
+            # Profiling shows an execution around 3-5% slower with -Os vs -O3,
+            # so, kept only -flto.
```
fabioz, line 170: Note: if the additional reduction from `-O3` to `-Os` makes a difference in building for more versions out of the box (150kb/.so), the 3-5% performance penalty may be reasonable.

Note that I did the checks here with `gcc 7.5.0`.

PR #531, opened 2021-01-22
Add object id if variable name would be duplicate. Fixes #148

### LINKED ISSUE #148: Debugger: Dictionaries not correctly listed for multiple key objects with identical representation
## Environment data
-   VS Code version: 1.45.0-insider
-   Extension version (available under the Extensions sidebar): 2020.4.76186
-   OS and version: Linux Mint 18.2
-   Python version (& distribution if applicable, e.g. Anaconda): 3.6.8 : Anaconda
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
-   Relevant/affected Python packages and their versions: None
-   Relevant/affected Python-related VS Code extensions and their versions: None
-   Jedi or Language Server? (i.e. what is `"python.jediEnabled"` set to; more info microsoft/vscode-python#3977): [default]
-   Value of the `python.languageServer` setting: [default] 
(Fresh install of the insider build and default Python extension)

## Expected behaviour

Show the full dictionary listing in the debugger "Variables" window.

## Actual behaviour

The listing only shows the first occurance of duplicate `__repr__` values for dictionary key items.

## Steps to reproduce:

```
class T:
    def __init__(self, name, value):
        self.name = name
        self.value = value
    def __repr__(self):
        return self.name

td = {T("foo", 24): "bar",
      T("gad", 42): "zooks",
      T("foo", 12): "bur"}

print("Keys: " + str((td.keys())))  
print("Values: " + str((td.values()))) # Breakpoint here
```

1. We need Python objects that can return identical representations for different objects (`__repr__()`).
2. Create multiple of such objects and use them as key values for a dictionary.
3. Start the debugger and look at the dictionary in the variable listing. 
4. Only the first key from all duplicate representations will be displayed, even though the dictionary may contain multiple different key objects.

Screenshot using the code example:
Dictionary_Debugger_bug

fabioz:


src/debugpy/_vendored/pydevd/setup_cython.py
```diff
@@ -196,7 +196,7 @@ def build_extension(dir_name, extension_name, target_pydevd_name, force_cython,
             Extension(
                 "%s%s.%s" % (dir_name, "_ext" if extended else "", target_pydevd_name,),
                 c_files,
-                **kwargs,
+                **kwargs
```
fabioz, line 199: It seems having the trailing comma here is not valid on Python 2 (so, this is an unrelated fix for Python 2).

karthiknadig (approved):

int19h (approved):

PR #548, opened 2021-02-18
Code reloading. Fixes #212

### LINKED ISSUE #212: Code Reloading
Allow code to be modified while at break-point, then continue executing the modified code.

fabioz:


int19h:
It looks like the test runs were timing out on Windows. I've increased the timeout, but you'll need to rebase for the run to pick up the changes to `azure-pipelines.yml`.

fabioz:
@int19h @karthiknadig ping (waiting for review of this pull request).

karthiknadig (approved):

karthiknadig:
@luabud This is another thing that we can enable with experiments.

karthiknadig:
I think this should be disabled by default and turned on via experiment flag.

What do you think @int19h @luabud ?

int19h (approved):

PR #551, opened 2021-02-22
Update pipelines.yaml

int19h:
Increase test timeout on Windows.


int19h:
It looks like our Python 3.x suite is running for >30 minutes, which causes it to time out.

karthiknadig (approved):

PR #563, opened 2021-03-15
Support step into target. Fixes #288

### LINKED ISSUE #288: Support "step into targets" in VS Code
From VS Code's 1.46 update:
https://code.visualstudio.com/updates/v1_46#_step-into-targets

### _Step Into Targets_
_Debugging into a statement with many nested function calls can be painful because you might have to step through functions you are not really interested in. To improve this scenario, the Debug Adapter Protocol has supported the "Step Into Target" feature that displays UI for directly stepping into the function you are interested in._

_In this milestone, VS Code has started to use this feature. When debugging has stopped on a statement, the context menu shows a new action Step Into Targets. This action presents a list of all targets (typically functions) that are available at the current stack frame. Selecting one of them will step into that target._

_Note that every stack frame can have its own set of targets. So the result of the Step Into Targets actions depends on the currently selected frame in the CALL STACK view._

Is that something we can add support to? 😊

fabioz:


fabioz:
@int19h @karthiknadig this is ready to review.

karthiknadig (approved):

Issue #576, opened 2021-03-29
Step Into not listing targets properly


fabioz:
Doing some testing I found an issue in how the targets are computed.

A construct such as:
```
    def function():
        yield sys._getframe()
        bbb = foo.bar(
            Something(param1, param2=xxx.yyy), {}
        )
        call()
```

Is not finding `bar` properly.

The main issue is that a stack is created and later when a `CALL_FUNCTION` bytecode is found items are popped from the stack, but the stack is incomplete, so, either the way to deal with the stack becomes much more complete (as we need to deal with much more bytecode to have a complete picture on how the stack would look -- which may not be worth it) or go with a simpler approach to just list the names without validating if those would actually be function calls (which may be reasonable since cases such as accessing properties would also work -- those don't work right now with the current approach).

I'm pending towards implementing the 2nd approach...

fabioz:
@int19h @karthiknadig the approach I had in mind doesn't work because the bytecode offset won't work.

i.e.: What the step into target currently does is check in a called frame that the back frame was the initial frame where the step into was requested and the bytecode offset matches the expected bytecode for the call. The `LOAD_XXX` instructions have the wrong offset to check that, so, the only way I can see how to fix this is actually properly implementing the stack operations to know which name matches a given call operation (or possibly a load to deal with `@property`).

It'd be comparable to something as: https://github.com/nedbat/byterun/blob/master/byterun/pyvm2.py but only working at the metadata level, not really interpreting code.

So, I think this is going to take a while to implement (my guess is it'll take around 10-15 working days (which at my current time allocated means something around 2 months) as it needs to support many versions of python, which means lots of bytecodes to test). My plan is starting to work at that right now, but possibly pausing for smaller (or more critical) issues over time.

-- I'm also open to other approaches you may envision here.... another approach I thought of would be using the actual source code and creating an AST and then trying to do a match from which name loaded matches which called name to match the bytecode -- I think it'd be faster to implement, but I'm not 100% sure we can cover all cases this way and it wouldn't work on cases where sources aren't available.

fabioz:
@int19h @karthiknadig I think that if we could support only Python 3.5 onwards for this feature it'd be nicer to implement.

That would enable using https://github.com/MatthieuDartiailh/bytecode (which is already integrated into debugpy) as well as ignoring a bunch of Python 2.7 specific bytecode...

Do you think that's ok?

karthiknadig:
That works.

fabioz:
[...]

PR #602, opened 2021-04-29
Improvements to step into target. Work in progress: #576

fabioz:
Note that although this is still a work in progress (because I still don't handle all the bytecodes -- as well as some corner cases related to prediction of bytecode offsets where the bytecode offset communicated by CPython is not correct), I think that this is in a much better state than what's currently in the `main` branch, so, I'd like to get this in and then provide smaller pull requests as I progress as there's a lot of code already and the general structure shouldn't change that much anymore.

Changes in this pull request:

- Changes the structure to create a stack from the bytecode information
  so that function call names are correctly computed (this is still a work
  in progress as not all bytecode instructions are currently supported).
- Support to step into function calls inside comprehensions/generators.
- Updates the bytecode dependency to 0.12.0.
  Note: changes were done to the bytecode library to keep the bytecode offset.


fabioz:
Note: Linux tests seem to be failing due to:

`b"stderr: b'/bin/sh: 1: gdb: not found\\n'"`

int19h (approved):

PR #606, opened 2021-05-04
Fix Linux CI

int19h:
Use ubuntu-20.04 to run Linux tests.

Update apt packages before trying to install gdb.

### DIFF
--- a/azure-pipelines/linux-coverage.yaml
+++ b/azure-pipelines/linux-coverage.yaml
@@ -8,9 +8,10 @@ jobs:
       pool:
           name: Azure Pipelines
           demands: java
-          vmImage: "ubuntu-18.04"
+          vmImage: "ubuntu-latest"
       steps:
           - script: |
+              sudo apt-get update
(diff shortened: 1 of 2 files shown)


azure-pipelines/linux-coverage.yaml
```diff
@@ -8,9 +8,10 @@ jobs:
       pool:
           name: Azure Pipelines
           demands: java
-          vmImage: "ubuntu-18.04"
+          vmImage: "ubuntu-20.04"
```
karthiknadig, line 11: can we just use `ubuntu-latest`?

karthiknadig (approved):

azure-pipelines/linux-coverage.yaml
int19h: We can - I just didn't want the pipeline to suddenly start failing if it updates. Then again, if we suddenly start failing on the latest Ubuntu, we probably want to know, so I'll change that.

karthiknadig (approved):

PR #614, opened 2021-05-07
Install matplotlib on settrace. Fixes #603

### LINKED ISSUE #603: Debugging scripts that use matplotlib interactively on vscode....
## Environment data

-   VS Code version: 1.55.2 user setup
-   Extension version (available under the Extensions sidebar): v2021.4.765268190
-   OS and version: Windows 10 Version	10.0.19042 Build 19042
-   Python version (& distribution if applicable, e.g. Anaconda):  python 3.9.2 with conda 4.9.2
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): 
-   Relevant/affected Python packages and their versions: matplotlib 3.3.4
-   Relevant/affected Python-related VS Code extensions and their versions: None that I can think of
-   Value of the `python.languageServer` setting: I don't think this is related to my issue

## Expected behaviour

Would like to debug script that uses matplotlib interactively. By interactively I mean placing a break point and watching my figure update as I actually step through my code. PyCharm has this behavior out of the box and it is extremely useful.

## Actual behaviour

Stepping through a script using matplotlib will not generate an interactive figure.

## Steps to reproduce:

Create a file called test.py and paste the following code:

```
import matplotlib.pyplot as plt


figure = plt.figure()
ax = plt.gca()
ax.plot([1,2,3])
ax.set_title("hello")
```
 
Place a breakpoint on line 4, create a standard debug configuration and start the debugger. Step over lines 4, 5, and 6. No plot appears.

fabioz:


karthiknadig (approved):

awa5114:
@fabioz I know this issue is closed but how do I incorporate your changes into my current vscode install?

PR #622, opened 2021-05-13
Don't compare objects prior to calling __xreload_old_new__. Fixes #615

### LINKED ISSUE #615: Value error being thrown when using code reload
## Environment data

- debugpy version: 1.3.0
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): 3.8.5 (venv)

## Actual behavior

Using code reload on a Python script with a numpy array defined as a global variable is throwing the following error:

```
  File "c:\Users\luabud\.vscode-insiders\extensions\ms-python.python-2021.6.821261989-dev\pythonFiles\lib\python\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_reload.py", line 323, in _update
    if oldobj != newobj and str(oldobj) != str(newobj) and repr(oldobj) != repr(newobj):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```
If I make the np array a list instead, things work as expected. 

## Expected  behavior

No error

## Steps to reproduce:
1. Create a Python file with the following content:
```py
import random
import numpy as np

houses = np.array(["Gryffindor", "Hufflepuff", "Ravenclaw", "Slytherin"])

def select_house():
    house = random.choice(houses)
    return house

if __name__ == "__main__":
    print(f"You belong to {select_house()}!")

```
2. Create a launch.json file with the following configuration:

```json
{
            "name": "Python: Current File",
            "type": "python",
            "request": "launch",
            "program": "${file}",
            "autoReload": {
                "enable": true
            }
}
```
3. Add a breakpoint to line 11 (with the print call) to the first Python file and hit F5.
4. Change line 8 to e.g. `return "Hufflepuff"` and hit save
5. Watch the error in the terminal

fabioz:


fabioz:
Note: the failures seem related to a flask update and not due to the changes in this pull request.

karthiknadig (approved):

PR #653, opened 2021-07-02
Fix deprecated import for MutableMapping and MutableSet

mgorny:
Starting with Python 3.3, MutableMapping and MutableSet are part
of collections.abc module rather than collections.  Importing them
from the old module is deprecated and no longer works in Python 3.10.
Support both modules conditionally for best compatibility.

### DIFF
--- a/tests/debug/config.py
+++ b/tests/debug/config.py
@@ -4,11 +4,16 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import collections
 import os
+import sys
+
+if sys.version_info >= (3, 3):
+    from collections.abc import MutableMapping, MutableSet
+else:
+    from collections import MutableMapping, MutableSet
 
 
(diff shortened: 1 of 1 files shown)


tests/debug/config.py
```diff
@@ -137,8 +142,7 @@ def setdefault(self, key, default=None):
         return self[key]
 
     def setdefaults(self, defaults):
-        """Like setdefault(), but sets multiple default values at once.
-        """
+        """Like setdefault(), but sets multiple default values at once."""
```
mgorny, line 145: These somewhat irrelevant changes were made by `black`.

fabioz (approved):

fabioz:
Thanks for the pull request... I'll wait for the tests to pass and if everything works I'll merge it shortly.

karthiknadig (approved):

PR #662, opened 2021-07-16
Fix issue with step into target with `LOAD_METHOD`. Fixes #661

### LINKED ISSUE #661: Step into targets doesn't seem to respect selected option
## Environment data

- debugpy version: 1.3.0
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): 3.8.10 (venv)

## Expected  behavior

Selecting the "Image" option that shows after I run "Step into target..." should step into the Image file:

image

## Actual behavior

Selecting "Image" gives the same behaviour as if I did a normal "step into" (i.e. steps into the parse_content method)

## Steps to reproduce:
1. Clone https://github.com/luabud/pvscc-logo
2. Set up a venv and install the dependencies in the requirements.txt file
3. Create the following launch.json configuration:
```json
{
            "name": "Python: Current File",
            "type": "python",
            "request": "launch",
            "program": "${file}",
            "justMyCode": false,
            "autoReload": {"enable": true}
}
```
4. Open the generate_logo.py file (located on the root of the project)
5. Set up a breakpoint on line 13
6. Start debugging (F5) using the configuration we created above
7. When the debugger stops on line 13, right-click on line 15 (`wc = generate_wordcloud(parse_content(url),np.array(Image.open(mask_path)))` ) and select "Jump to Cursor". This is just to avoid the if condition on line 14, that will return false by default. 
8. Once in line 15, right click and select "Step into Targets..." 
9. Select "Image"
10. Watch it step into the parse_content method instead

fabioz:


src/debugpy/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -123,13 +123,15 @@ def write_list_threads(self):
     def wait_for_terminated(self):
         return self.wait_for_json_message(TerminatedEvent)
 
-    def wait_for_thread_stopped(self, reason='breakpoint', line=None, file=None, name=None):
+    def wait_for_thread_stopped(self, reason='breakpoint', line=None, file=None, name=None, preserve_focus_hint=None):
```
fabioz, line 126: Changes on `test_debugger_json.py` are unrelated to this issue (it seems that was missing from the previous pull request, so, just catching up here).

fabioz:
It seems I added this too soon (I still have some issues in the ci, so, closing for now -- I'll reopen when it's ready).

Issue #689, opened 2021-08-05
test_django fails on py3.10


int19h:
It's due to deprecation warnings (for `distutils`) produced by Django itself; see https://code.djangoproject.com/ticket/32074.

The corresponding tests are disabled on 3.10 for now, but will need to be re-enabled when Django 3.3 ships with the fixes.

int19h:
Django 3.2.9 shipped, need to update this now.

Issue #688, opened 2021-08-05
test_gevent crashes on py3.10


int19h:
Presumably due to bytecode changes.

fabioz:
I don't think so... for Python 3.10 we need to support a newer version of gevent (the one used in the tests can't really work with Python 3.10).

So, we need to tackle https://github.com/microsoft/debugpy/issues/515 for this to be fixed.

PR #690, opened 2021-08-05
Assorted fixes for Python 3.10 support

int19h:
Split test/requirements.txt into two different lists, one for py2, and the other for py3; update package versions in py3 list as needed to support py3.10.

Fix usage of deprecated threading functions getName(), setName(), isDaemon(), and currentThread().

Refactor test_invalid_breakpoints to be more declarative to handle Python version differences more easily and clearly, and fix it for py3.10.

Disable Django tests on py3.10 (#689).

Disable gevent tests on py3.10 (#688).

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_daemon_thread.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_daemon_thread.py
@@ -35,7 +35,7 @@ class PyDBDaemonThread(threading.Thread):
         created_pydb_daemon[self] = 1
         try:
             try:
-                if IS_JYTHON and not isinstance(threading.currentThread(), threading._MainThread):
+                if IS_JYTHON and not isinstance(threading.current_thread(), threading._MainThread):
                     # we shouldn't update sys.modules for the main thread, cause it leads to the second importing 'threading'
                     # module, and the new instance of main thread is created
                     ss = JyCore.PySystemState()
@@ -59,7 +59,7 @@ class PyDBDaemonThread(threading.Thread):
 
     def do_kill_pydev_thread(self):
         if not self._kill_received:
-            pydev_log.debug('%s received kill signal', self.getName())
+            pydev_log.debug('%s received kill signal', self.name)
             self._kill_received = True
 
     def _stop_trace(self):
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py
@@ -448,4 +448,3 @@ def interrupt_main_thread(main_thread):
                 main_thread._thread.interrupt()  # Jython
         except:
             pydev_log.exception('Error on interrupt main thread fallback.')
-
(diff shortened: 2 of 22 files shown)


int19h:
@fabioz I wasn't sure what the best place for the new threading helpers was, so this is a best guess - lmk if you'd prefer to organize that code differently.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py
```diff
@@ -449,3 +449,17 @@ def interrupt_main_thread(main_thread):
         except:
             pydev_log.exception('Error on interrupt main thread fallback.')
 
+
+def current_thread():
```
fabioz, line 453: `threading.current_thread` is already available in Python 2.6, so, I don't think we actually need this helper function. If an alias is interesting, I think it should be in `pydevd_constants`, right after the `threading` module is imported.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_utils.py
```diff
@@ -449,3 +449,17 @@ def interrupt_main_thread(main_thread):
...
+    try:
+        impl = threading.current_thread
+    except AttributeError:
+        impl = threading.currentThread
+    return impl()
+
+
+def get_thread_name(t):
```
fabioz, line 461: `threading.Thread.name` is already available in Python 2.6, so, I don't think we actually need to do the try..except.

Maybe we can just inline this method and just use `t.name` where needed...

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_daemon_thread.py
```diff
@@ -26,6 +26,11 @@ def __init__(self, py_db, target_and_args=None):
         mark_as_pydevd_daemon_thread(self)
         self._target_and_args = target_and_args
 
+        if hasattr(self, 'name'):
```
fabioz, line 29: Is this really needed? Ideally I think we could just set `thread.name = name` where needed.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -1317,7 +1317,11 @@ def has_user_threads_alive(self):
                     'Error in debugger: Found PyDBDaemonThread not marked with is_pydev_daemon_thread=True.\n')
 
             if is_thread_alive(t):
-                if not t.isDaemon() or hasattr(t, "__pydevd_main_thread"):
+                try:
+                    is_daemon = t.daemon
```
fabioz, line 1321: `t.daemon` is available in `Python 2.6` already, so, I don't think we need the try..except.

fabioz (changes requested):
Just some minors, mainly pointing that I think we should actually use the new method formats without worries about backward compatibility since I just checked and even Python 2.6 already had those.

(i.e.: just use t.name / t.daemon / threading.current_thread directly).

int19h:
I wasn't sure if that's also the case for Jython etc? If it's okay to just use the new APIs everywhere, that's even better.

fabioz:
> I wasn't sure if that's also the case for Jython etc? If it's okay to just use the new APIs everywhere, that's even better.

It should be (only Jython 2.7 is supported, so, it should be compatible with Python 2.7 in that regard).

fabioz (approved):

PR #697, opened 2021-08-13
Python 3.10: support for attach to process and set tracing to other threads.

fabioz:
Note that Python 3.10 support is still a work in progress.

What still needs work:
1. Unhandled exceptions (bytecode analysis)
2. Step into (still missing bytecode)
3. Frame eval mode


src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/python.h
```diff
@@ -48,217 +48,6 @@ class PyVarObject : public PyObject {
     size_t ob_size; /* Number of items in variable part */
 };
 
-// 2.4 - 2.7 compatible
-class PyCodeObject25_27 : public PyObject {
```
fabioz, line 52: Note that I removed many declarations which weren't really being used.

int19h (approved):

Issue #155, opened 2020-05-02
manylinux wheels don't pass auditwheel checks


SylvainCorlay:
## Environment data

- PTVSD version: 4.3.2 (latest)
- OS and version: Manylinux1
- Python version (& distribution if applicable, e.g. Anaconda): Any

## Actual behavior

auditwheel fails on ptvsd's maynlinux wheels with 

```
RuntimeError: Invalid binary wheel, found the following shared library/libraries in purelib folder:
	pydevd_cython.cpython-35m-x86_64-linux-gnu.so
	attach_linux_amd64.so
	attach_linux_x86.so
```

SylvainCorlay:
You may eventually need to fix it. I have read in a couple of places that the pypifolks were looking into running such checks to accept binary wheels.

SylvainCorlay:
Note that this is not an issue for us. I came across this as I was running auditwheel on xeus-python, for the Jupyter debugger.

int19h:
Turns out that those warnings are not entirely spurious. We messed up the packaging such that all our stuff goes inside `purelib` in the wheel, which is in violation of the manylinux PEP requirements. 

Furthermore, for the attach helper .so, we have some non-conformant dependencies (glibc version) - but the codepath that conditionally loads them only checks arch, not glibc. So it's possible to install that wheel in an environment that is supposed to be able to run manylinux1 binaries, and then have it fail at runtime (instead of fallback), if you use attach-to-PID.

PR #701, opened 2021-08-20
Build attach_linux_*.so for manylinux1 using the official containers from PyPA

int19h:
(for #155)


int19h:
Note that this doesn't actually make it pass `auditwheel` just yet, because the x86 version runs into https://github.com/pypa/auditwheel/issues/323 (and it's still present in the x86_64 wheel). The 64-bit wheel does pass audit with these changes if the x86 binary is manually removed from it.

karthiknadig (approved):

fabioz (approved):
As a note, I actually did a change to generate the binaries using GitHub Actions (https://github.com/fabioz/PyDev.Debugger.binaries/tree/master/.github/workflows), so, that's actually the standard way I'm now using to update the binaries for multiple platforms (I guess I could've let it in the `pydevd` repo, but it was a bit annoying to have lots of jobs which should almost never be run instead of just 1).

Anyways, this is just to note that I'll update that accordingly to use manylinux for subsequent updates. I have to investigate how to use that on the CI -- maybe I'll have to change that to a `.sh` when this is done -- are you ok with that or would you rather keep the `.cmd` too?.

int19h:
I think it would be best to have scripts for both - having a script locally is handy when iterating on something, regardless of the platform. Maintenance overhead should be minimal since it's literally a one-liner.

PR #715, opened 2021-09-01
Do not install tox into tox

frenzymadness:
Having tox as a dependency in requirements files means that
it's also installed in all the tox environments.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -6,7 +6,6 @@ pytest
 pytest-xdist
 pytest-cov
 pytest-timeout
-tox
 
 ## Used by test helpers:
 
--- a/tests/requirements27.txt
+++ b/tests/requirements27.txt
@@ -8,7 +8,6 @@ pytest-xdist<2
 
 pytest-cov
 pytest-timeout
-tox
 
 ## Used by test helpers:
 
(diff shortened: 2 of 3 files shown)


int19h:
The idea behind `test/requirements.txt` was that you should be able to `pip install -r` it, and have everything you need to run the tests, either with tox or without. Excluding tox from the list would preclude the former.

OTOH, having it installed in the environments that tox creates is redundant, but it doesn't harm anything.

frenzymadness:
I understand you but I have a feeling that we are mixing two approaches. Either I want to install all test dependencies into my current environment and then run tests without tox or I want to use tox. In the first case, I don't need tox, in the second case I need only tox and not the rest of the test dependencies. And `pip install tox` is much shorter than `pip install -r test/requirements.txt`.

I believe that the main way of running tests should be tox and then it does not make sense to install it into its own environments and you do not need to have requirements.txt file for only one requirement. OTOH, if you really want to use pytest directly, you can, and then tox is useless for you.

int19h:
Makes sense. I wanted to say that the docs on contributing should be changed accordingly; but, apparently, we already tell people to just install tox directly.

int19h (approved):

PR #721, opened 2021-09-09
Update pydevd_process_net_command_json.py

Daniel-DI:
- primarily, I changed the line `if self._options.django_debug:` to `if self._options.django_debug or self._options.flask_debug:`, so that if the debug option `"flask": true` is set, the debugger does not interrupt the program flow when the flask app is restarted due to a source file being changed. When a change in a source file is observed by the debugger, it causes a `SystemExit` exception with code `3`, and the reloader automatically restarts the flask app.
- the other restructuring I did was due to my interpretation that even if `self._options.break_system_exit_zero` is `true`, i.e. if the debugger interrupts the flow when the program exits with error code `0`, the debugger still should not interrupt if it reloads the flask app due to a detected change in a source file.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
@@ -395,13 +395,14 @@ class PyDevJsonCommandProcessor(object):
 
         self.api.set_show_return_values(py_db, self._options.show_return_value)
 
+        ignore_system_exit_codes = []
         if not self._options.break_system_exit_zero:
-            ignore_system_exit_codes = [0, None]
-            if self._options.django_debug:
-                ignore_system_exit_codes += [3]
-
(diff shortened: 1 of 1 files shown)


int19h:
@fabioz Can you take a look?

fabioz:
@Daniel-DI tests seem to be failing with this change. Can you take a look at adding a test for your use-case and fixing the failing test?

Daniel-DI:
I reverted the second change, because, as you wrote, it changes things in a way that breaks the tests as they are currently set up. I'm not versed well enough yet with tests management to "fix" those, so in the interest of simplicity and the "philosphy of small steps", i decided to undo that change.
the only change that remains is just affecting one single line, and should be self-explanatory: regarding the question of whether system exit code 3 should be ignored, both `self._options.django_debug` and `self._options.django_debug` should be considered in the same way.

I have no idea why the tests `debugpy-test-automation` and `debugpy-test-automation (Tests - Linux py36)` fail, but from what i've read under `Details`, i have the impression that it's not related to my change. Not sure though.

fabioz (approved):

PR #786, opened 2021-11-18
Update CONTRIBUTING.md

chubei-oppen:
The `debugAdapterPath` option should be set to `debugpy/src/debugpy/adapter`


int19h:
Good catch, thank you!

int19h (approved):

PR #787, opened 2021-11-18
Add option for enable gui event loop when not using matplotlib

### LINKED ISSUE #782: Enable gui event loop to run when a breakpoint is hit even if not using matplotlib
Currently `debugpy` checks if `matplotlib` is imported and runs its event loop if possible. This functionality can be useful for other libraries with their own event loop too.

For example, a 3D rendering library can register a `inputhook` and render one frame when called. This makes it possible to interactivly debug 3D rendering applications(games, scientific visualizations, etc.) with `debugpy`, specifically, in VS Code.

Very curious to hear your opinion about this.

chubei-oppen:
Closes #782

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
@@ -19,6 +19,7 @@ class DebugOptions(object):
         'flask_debug',
         'stop_on_entry',
         'max_exception_stack_frames',
+        'gui_event_loop',
     ]
 
     def __init__(self):
@@ -30,6 +31,7 @@ class DebugOptions(object):
         self.flask_debug = False
         self.stop_on_entry = False
         self.max_exception_stack_frames = 0
+        self.gui_event_loop = 'matplotlib'
 
     def to_json(self):
         dct = {}
@@ -92,6 +94,8 @@ class DebugOptions(object):
 
         self.max_exception_stack_frames = int_parser(args.get('maxExceptionStackFrames', 0))
 
+        if 'guiEventLoop' in args:
+            self.gui_event_loop = str(args['guiEventLoop'])
 
 def int_parser(s, default_value=0):
     try:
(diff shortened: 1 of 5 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_json_debug_options.py
```diff
@@ -61,6 +63,9 @@ def update_fom_debug_options(self, debug_options):
 
         # Note: _max_exception_stack_frames cannot be set by debug options.
 
+        if 'ENABLE_GUI_EVENT_LOOP' in debug_options:
+            self.enable_gui_event_loop = debug_options.get('ENABLE_GUI_EVENT_LOOP')
+
```
int19h, line 68: This bit shouldn't be necessary - `debugOptions` is a legacy property for boolean flags in debug config that we had before deciding to just use separate boolean properties; it is really only kept for backwards compatibility.

chubei-oppen: I was wondering what those optoins are for... Removed in https://github.com/microsoft/debugpy/pull/787/commits/f07013df3b0e1932f2f6701e8d8b5dbe9bb2c24a

fabioz:
@chubei-oppen I was thinking that in the current way, to get it to work users would need to do additional work (to call something as: `pydevd_ipython.inputhook.enable_gui('qt5')` or `pydevd_ipython.inputhook.set_inputhook(callback)`) and if `matplotlib` is indeed imported afterwards it'd override that value...

So, what do you think about making `guiEventLoop` a `string` and if it matches `matplotlib` (default), nothing would change, otherwise one of the gui constants in `pydevd_ipython.inputhook` would be expected, then it'd call `pydevd_ipython.inputhook.enable_gui` with that constant, but if it still doesn't match it'd be expected to map to a callable which could be then be imported/set in `pydevd_ipython.inputhook.set_inputhook`.

Does that make sense?

Also, note that if that setting is not `matplotlib`, `init_matplotlib_support` should be a no-op (so that we don't have the override issue).

chubei-oppen:
@fabioz Sounds good. I'll find time later for the implementaion.

chubei-oppen:
@fabioz I force pushed the new implementation.

fabioz:
It seems good to me, but I want to do some additional manual testing before merging (I'm finishing another issue related to gevent and as soon as I finish that I'll do the final checks/integrate this pull request, sorry for the delay here...)

fabioz:
Closing in favor of https://github.com/microsoft/debugpy/pull/787 (which includes the changes here and has some additional fixes/tests on top of it).

Issue #790, opened 2021-11-19
Add tooltips to exception categories


andersea:
The options for debugging exceptions are very confusion. There are now three different options, with some combinations seeming to not make sense.

![exception options](https://user-images.githubusercontent.com/10992982/142561057-48b03da4-b90e-4730-819a-e16ba0bdf00d.png)

I would very much like to see, not only thorough documentation, with examples, on each individual option, but also their expected interactions. With three options to chose from, there are now 8(!) possible combinations for how these settings could be picked. Probably most of these don't make sense.

UI wise, VSCode probably is lacking here, as it could be presented better, but good documentation is a first step.

I am seeing a lot of weird behaviour, when debugging exceptions. For instance, as far as I can see an [old bug from ptvsd #2027](https://github.com/microsoft/ptvsd/issues/2027) has regressed in debugpy (at least for async trio code), but it is hard for me to test and report anything, when I have no clue what the expected behaviour is, under the various conditions that are available.

andersea:
See also #364 for similar concerns. Actually I believe that that issue has regressed now for all trio code, but again, hard to know, when I don't know what to expect.

fabioz:
- Raised Exceptions:
Triggers whenever any exception is raised or reraised (if `"justMyCode":true` -- which is the default -- only triggers when an exception is raised or reraised in user code).

- Uncaught Exceptions:
Triggers only when an exception would get to the entry point method of a thread (so, only if an exception would make a thread exit).

- User Uncaught Exceptions:
Triggers when a raised/reraised exception would cross the barrier from user code to library code (so, when an exception is raised or reraised it checks whether the caller is library code and the current method is in user code, if it is, then the exception is shown -- in general it's helpful if a user wants to see exceptions which are being raised in a test case and would end up being caught inside the test framework -- it can be helpful to uncover bugs in other situations too, but it may be hard to use if in your usage you do get such exceptions regularly as a part of a framework you're using).

All modes are kind of independent (so, enabling one shouldn't change the behavior of the other, although `raised exceptions` with `"justMyCode":false` is kind of a superset of the others).

@karthiknadig @int19h maybe something like that could be added to: https://code.visualstudio.com/docs/python/debugging.

andersea:
Thanks for taking a look at this. I am sorry if this seems nitpicky, but I really think it would be helpful to be precise on these points. Maybe we can go through these one at a time, to make sure I completely understand and that we are in agreement?

* Raised Exceptions

**Precondition:**
[...]

PR #798, opened 2021-12-02
Add tooltips for exception breakpoint filters. (#790)

int19h:


src/debugpy/adapter/clients.py
```diff
@@ -143,12 +143,23 @@ def initialize_request(self, request):
...
+                "filter": "uncaught",
+                "label": "Uncaught Exceptions",
+                "default": True,
+                "description": "Break when the process is exiting due to unhandled exception.",
+            },
             {
                 "filter": "userUnhandled",
                 "label": "User Uncaught Exceptions",
                 "default": False,
+                "description": "Break when exception escapes into library code.",
```
karthiknadig, line 162: @luabud @brettcannon For input on the text being used for various exception breakpoint.

karthiknadig (approved):

src/debugpy/adapter/clients.py
brettcannon:
```suggestion
                "description": "Break when an exception escapes into library code.",
```

Maybe "3rd-party" instead of "library"? No real opinion honestly.

int19h: It has always been a tricky one to name concisely. The only baseline that we have here is "just my code" as the name of the related breakpoint feature.  But it's basically "stdlib + site-packages" by default, and there are also ways to customize it. So "third-party" is not necessarily accurate if there are some first-party dependencies that just happen to be installed into site-packages (including venv).

luabud: "non-user code" maybe? But also no preference here 😅 library code is also fine as is. a

int19h: "Non-user code" is where I started, but then it's kinda redundant wrt the option name; so I wanted the tooltip to give more clues as to what "non-user" actually means.

The other thing is that "user code" makes sense to *us*, but less so to the actual user - to them, it's "my code"! E.g. consider "just my code" vs "user-code only"... the catch here is that we label the option itself "User uncaught exceptions", so it's already wrong in this sense, tooltip or no tooltip. But we could rename it to "Exceptions not caught in my code" or something like that?

PR #800, opened 2021-12-03
Add option for enable gui event loop when not using matplotlib (guiEventLoop)

### LINKED ISSUE #782: Enable gui event loop to run when a breakpoint is hit even if not using matplotlib

fabioz:
This has additional fixes on top of: https://github.com/microsoft/debugpy/pull/787 (so, closing that one).

Fixes #782

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
@@ -125,6 +125,9 @@ class PyDevdAPI(object):
         '''
         pydevd_file_utils.set_ide_os(ide_os)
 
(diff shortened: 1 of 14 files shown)


int19h (approved):

chubei-oppen:
Very excited to see this merged! When will it be released?

int19h:
We'll have another debugpy release before the year end, but it'll take a bit more time for that to get picked up by the clients.

PR #866, opened 2022-03-09
Corrected versioneer install

r3m0t:
I tried using `git archive` and extracting the archive and running `python setup.py sdist`, but I got version `0+unknown`. This fixes that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-debugpy/_version.py export-subst
+src/debugpy/_version.py export-subst


r3m0t:
You could try reinstalling versioneer (I didn't try that) https://github.com/python-versioneer/python-versioneer

int19h (approved):

int19h:
We changed the repo layout from all-root to src/ a while ago, so it must have been lurking since then. Thank you!

PR #873, opened 2022-03-18
More code changes related to dropping Python 2.7.

fabioz:
This should conclude the work to drop Python 2.7 from pydevd.

Some noteworthy changes which could break some tooling:

The top-level `setup_cython.py` was renamed to `setup_pydevd_cython.py`

A couple of top-level packages were inlined/moved (i.e.: `_pydev_imps` and `pydevd_concurrency_analyser`).


int19h (approved):

int19h:
Looks like some package renames broke vendoring logic that tries to preload those packages.

Issue #346, opened 2020-07-24
Debugger Heisenbug where integer comparison appears to fail within a specific context


dddvision:
## Environment data

-   VS Code version:
Version: 1.47.2
Commit: 17299e413d5590b14ab0340ea477cdd86ff13daf
Date: 2020-07-15T18:22:15.161Z
Electron: 7.3.2
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Linux x64 5.4.0-42-generic snap

-   Extension version (available under the Extensions sidebar):
ms-python.python v2020.7.96456

-   OS and version:
LSB Version:	core-9.20170808ubuntu1-noarch:security-9.20170808ubuntu1-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.4 LTS
Release:	18.04
Codename:	bionic

-   Python version (& distribution if applicable, e.g. Anaconda):
Anaconda Python 3.7.7

-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...):
Python 3.7.7 64-bit ('base':conda)

-   Relevant/affected Python packages and their versions:
jsonpickle                    1.4.1  
numpy                         1.18.1  
SQLAlchemy               1.3.17 

-   Relevant/affected Python-related VS Code extensions and their versions:
ms-python.python v2020.7.96456

-   Value of the `python.languageServer` setting:
Microsoft

## Expected behaviour
When debugging the attached file and stepping into __main__ on line 33 with no arguments, the value of len(sys.argv) is 1, so the comparison N>=2 on line 35 should be False and the debugger should step to the else statement on line 39.

## Actual behaviour
The debugger steps to line 36 as if N>=2 were True. It also incorrectly steps to line 42 after comparing N>=3.

## Steps to reproduce:
Debug the attached file and step into __main__ on line 33 with no arguments, then see if the debugger steps to line 36 or 39. All settings are default with a fresh install of vscode and the python extension.

I have spent many hours trying to isolate the minimal code required to demonstrate this bug. As far as I can tell, every line of uncommented code in this file is required to demonstrate the problem. Furthermore, adding print statements before or after the bug makes it disappear.

If further evidence of this bug is needed, then please ask and I will provide a .gif.

## Logs
Nothing appears in the Output panel. In the terminal, the following appears:
[...]

PR #878, opened 2022-03-24
Re-enable frame eval mode by default.

fabioz:
It was disabled as temporary workaround in https://github.com/microsoft/debugpy/issues/346, but that was fixed and it appears it wasn't re-enabled...

### DIFF
--- a/src/debugpy/_vendored/force_pydevd.py
+++ b/src/debugpy/_vendored/force_pydevd.py
@@ -19,17 +19,10 @@ if _unvendored:
     # raise ImportError(msg)
     warnings.warn(msg + ':\n {}'.format('\n  '.join(_unvendored)))
 
-
 # If debugpy logging is enabled, enable it for pydevd as well
 if "DEBUGPY_LOG_DIR" in os.environ:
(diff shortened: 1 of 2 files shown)


fabioz:
As a note, this change does seem to make the tests more prone to failure.

I tried analyzing those but I couldn't find out why that happens (it's possible that it's just timing differences in the test run).

It's weird that the failing tests appear to be some variation of:

```
2022-04-01T14:14:17.3234090Z   File "/Users/runner/work/1/s/.tox/py38/lib/python3.8/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 491, in start_client
2022-04-01T14:14:17.3234950Z     s.connect((host, port))
2022-04-01T14:14:17.3235520Z ConnectionRefusedError: [Errno 61] Connection refused
```

Meaning that the debugger is trying to connect to the client but it's not successful (and the change isn't related to that at all).

Also, failures seem to be tied mostly to MacOS.

karthiknadig (approved):

int19h (approved):

Issue #865, opened 2022-03-09
Debugging through poetry drops subprocess


getglad:
## Environment data

- debugpy version: 1.5.1
- OS and version: masOS 12.2.1
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.9.10
- Using VS Code or Visual Studio: VS Code

## Actual behavior

I have a `click` application (https://click.palletsprojects.com/en/8.0.x/) that is configured to run as a `poetry` script. I have a `launch.json` config that can successfully launch the script, accept the arguments, and hit breakpoints that I set inside of the click functions.

The issue is that the debugger detaches after 6 seconds no matter if breakpoints are set, the code is allow to run uninterrupted, using different arguments to trigger a different function, etc.

The call stack is showing the CLI file and the functions I break at to be inside a subprocess, and I have `subprocess` set to `true`, but I am wondering if this is part of the issue.

I have reviewed [this ticket](https://github.com/microsoft/debugpy/issues/676) and would expect this to work fairly similarly.

## Expected  behavior

The debugger would launch and allow stepping, etc, as usual.

## Steps to reproduce:
1. Create a `click` file (below is using the example from their hello world)
```
# testcli.py

import click

@click.command()
@click.option('--count', default=1, help='Number of greetings.')
@click.option('--name', default='Your name')
def hello(count, name):
    """Simple program that greets NAME for a total of COUNT times."""
    for x in range(count):
        click.echo(f"Hello {name}!")

if __name__ == '__main__':
    hello()
```

2. Create a `poetry` project with a script pointed to a python file using `click`
```
# pyproject.toml

...
[tool.poetry.scripts]
hello = 'testcli:hello'
...
```

3. Add a `launch.json` config that can run `poetry`
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Foobar CLI",
      "type": "python",
      "request": "launch",
      "program": "${env:HOME}/.poetry/bin/poetry",
      "args": ["run", "hello"],
      "console": "integratedTerminal",
      "justMyCode": false,
      "subProcess": true,
    }
  ]
}
```

int19h:
[...]

PR #883, opened 2022-03-31
Send exit message before process is replaced. WIP #865

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
```diff
@@ -398,6 +401,19 @@ def make_send_breakpoint_exception_message(self, *args, **kwargs):
     def make_process_created_message(self, *args, **kwargs):
         return NULL_NET_COMMAND  # Not a part of the debug adapter protocol
 
+    @overrides(NetCommandFactory.make_process_about_to_be_replaced_message)
+    def make_process_about_to_be_replaced_message(self):
+        event = ExitedEvent(ExitedEventBody(-1, reason="processReplaced"))
```
int19h, line 406: Since we're adding a custom property to a standard message, let's do some namespacing here to prevent future conflicts - something like "pydevd.reason".

int19h (approved):

int19h:
@fabioz Is it good to merge?

fabioz:
I just need to do that change you asked for... I had to run some unforeseen errands here today and I wasn't able to do this and the other issue related to the attach (I'll put back the hours during the remaining of the month -- my plan is doing that change tomorrow and merging it).

fabioz:
I changed `reason` to `pydevdReason`.

PR #902, opened 2022-04-12
Allow using pydevd as a regular dependency.

Apteryks:
This makes it easier to test using PyDev-Debugger (pydevd) as a
standalone dependency rather than as a bundled (current status) one,
which comes with binaries checked in Git.

* setup.py (DEBUGPY_BUNDLING_DISABLED): New variable.
[DEBUGPY_BUNDLING_DISABLED]: Conditionalize 'debugpy._vendored'
import.
(PYDEVD_ROOT): Set only when DEBUGPY_BUNDLING_DISABLED is False.
(ExtModules.__bool__): Make the return value the negation of
DEBUGPY_BUNDLING_DISABLED.
(__main__)[DEBUGPY_BUNDLING_DISABLED]: New condition to call
'cython_build'.  Etch a __bundling_disabled__ variable on the debugpy
module.
<setuptools.setup>: Do not include the _vendored submodule in the
packages and package_data when DEBUGPY_BUNDLING_DISABLED is set.
Adjust has_ext_modules argument.
* src/debugpy/__init__.py: New '__bundling_disabled__' variable.
* src/debugpy/server/__init__.py [debugpy.__bundling_disabled__]:
Conditionalize 'debugpy._vendored.force_pydevd' import.
* src/debugpy/server/attach_pid_injected.py (attach): Import the
attach_script procedure from sys.path directly when
debugpy.__bundling_disabled__ is true and do not do any sys.path
manipulation in this case.
* tests/tests/test_vendoring.py: Skip test if
'debugpy.__bundling_disabled__' is true.
* src/debugpy/server/__init__.py: Partially copy the pydevd loading
logic from src/debugpy/_vendored/force_pydevd.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,9 @@ import subprocess
 import sys
 
 
+DEBUGPY_BUNDLING_DISABLED = bool(os.getenv('DEBUGPY_BUNDLING_DISABLED'))
+
+
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 import versioneer  # noqa
 
@@ -18,12 +21,15 @@ del sys.path[0]
 
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "src"))
 import debugpy
-import debugpy._vendored
+
+if not DEBUGPY_BUNDLING_DISABLED:
+    import debugpy._vendored
 
 del sys.path[0]
 
 
-PYDEVD_ROOT = debugpy._vendored.project_root("pydevd")
+PYDEVD_ROOT = (None if DEBUGPY_BUNDLING_DISABLED else
+               debugpy._vendored.project_root("pydevd"))
 DEBUGBY_ROOT = os.path.dirname(os.path.abspath(debugpy.__file__))
 
 
@@ -67,7 +73,7 @@ def iter_vendored_files():
 # relevant setuptools versions.
 class ExtModules(list):
     def __bool__(self):
-        return True
+        return not DEBUGPY_BUNDLING_DISABLED
 
 
 def override_build(cmds):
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -11,19 +11,25 @@
 import sys
 
 
+DEBUGPY_BUNDLING_DISABLED = bool(os.getenv('DEBUGPY_BUNDLING_DISABLED'))
```
int19h, line 14: For clarity, let's invert the meaning - i.e. `BUNDLE_DEBUGPY` which defaults to true but can be overridden.

setup.py
```diff
@@ -133,9 +139,24 @@ def tail_is(*suffixes):
 
 
 if __name__ == "__main__":
-    if not os.getenv("SKIP_CYTHON_BUILD"):
+    if not (os.getenv("SKIP_CYTHON_BUILD") or DEBUGPY_BUNDLING_DISABLED):
         cython_build()
 
+    # Etch bundling status in the source.
```
int19h, line 145: I don't think this is necessary - the code can check at runtime whether `debugpy._vendored` is present by trying to import it and catching `ImportError`. This can be cached as a package global to make one-liner checks easier.

Also, if caching, I'd prefer the variable name to not use the `__reserved__` naming pattern - it can be simply `is_pydevd_bundled` or something like that. We declare our public API surface via `__all__`, so the name not starting with `_` is not a concern.

src/debugpy/server/__init__.py
```diff
@@ -4,6 +4,50 @@
...
+        debugpy.breakpoint()
+
+    pydevd.install_breakpointhook(debugpy_breakpointhook)
+
+    # Ensure that pydevd uses JSON protocol
+    from _pydevd_bundle import pydevd_constants
+    from _pydevd_bundle import pydevd_defaults
+    pydevd_defaults.PydevdCustomization.DEFAULT_PROTOCOL = pydevd_constants.HTTP_JSON_PROTOCOL
+else:
+    import debugpy._vendored.force_pydevd  # noqa
```
int19h, line 53: That's a lot of duplicated code. I think `force_pydevd` should be refactored to only contain the logic that is related to vendoring, while this one can do all the configuration instead, and conditionally call the vendoring part at the right point (there might be more than one, so `force_pydevd` might need to be split into functions).

int19h (commented):
Can you please also run it through Flake8 and Black.

int19h:
Please re-open if this is still relevant.

PR #907, opened 2022-04-13
Fix typo in README example command

joshuatz:
I noticed that this example in the README does not seem to match the actual CLI usage (in source code, docs, and wiki). Figured I would just open a PR directly instead of an issue, as this seems like a straightforward edit 😀


int19h (approved):
Good catch, and thank you for the PR!

PR #911, opened 2022-04-14
Make errors with conditional breakpoints clearer. Fixes #893

### LINKED ISSUE #893: Make errors in conditional breakpoints more clear in logs
Are there ways to improve the error reporting when a conditional breakpoint is the source of a parsing error while debugging?

I had a bug in a breakpoint which I did not realize was set and it was printing a message like so:

```
Error while evaluating expression: str(p).lower.endswith('.txt')
  File "<string>", line 1, in <module>
AttributeError: 'builtin_function_or_method' object has no attribute 'endswith'
```

This does not provide any information that this code was being referenced from a breakpoint. I spent a significant amount of time trying to track down this bug in my code using search and wrapping things in try/except to no avail. Adding a formatter to my logger to try to print more precise locations for the message had no effect on these log messages and turning on `Raised Exceptions`, `Uncaught Exceptions` and `User Uncaught Exceptions` all also did not help.

It would be great to have some context that points to the error coming from the logic in the breakpoint. A few ideas:
- Make the debugger pause on a failed breakpoint (perhaps behind a checkbox that can be toggled)
- Make the log include some indication that it was from a breakpoint either with a message, custom exception or the file and line where the breakpoint was set that caused the error.

fabioz:
Note that the error is no longer printed to stderr. Rather it's sent as an `important` output message.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_api.py
@@ -129,7 +129,8 @@ class PyDevdAPI(object):
         py_db._gui_event_loop = gui_event_loop
 
     def send_error_message(self, py_db, msg):
-        sys.stderr.write('pydevd: %s\n' % (msg,))
+        cmd = py_db.cmd_factory.make_warning_message('pydevd: %s\n' % (msg,))
+        py_db.writer.add_command(cmd)
 
     def set_show_return_values(self, py_db, show_return_values):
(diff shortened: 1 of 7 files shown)


int19h:
Looks like it also needs an update to `test_error_in_condition`.

fabioz:
> Looks like it also needs an update to `test_error_in_condition`.

Done.

fabioz:
@int19h anything missing here?

int19h (approved):

PR #921, opened 2022-04-27
add automatic conversion from string to int/float type

giaco5988:
Relates to https://github.com/microsoft/vscode-python/issues/18841

Allow passing strings to the debugger for number-like inputs. With this fix strings will be accepted and automatically converted to int/float.

### DIFF
--- a/src/debugpy/common/json.py
+++ b/src/debugpy/common/json.py
@@ -93,6 +93,16 @@ class JsonObject(object):
 # some substitutions - e.g. replacing () with some default value.
 
 
+def _converter(value, classinfo):
+    """Convert value (str) to number, otherwise return None if is not possible"""
+    for one_info in classinfo:
+        if issubclass(one_info, numbers.Number):
+            try:
+                return one_info(value)
+            except ValueError:
+                pass
+
+
 def of_type(*classinfo, **kwargs):
     """Returns a validator for a JSON property that requires it to have a value of
     the specified type. If optional=True, () is also allowed.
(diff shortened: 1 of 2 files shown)


int19h:
Awesome, thank you! Could you also add a test for this? We don't have a separate group of tests for JSON utilities, but a simple test that just verifies that `request("foo", int)` works for a string value could go into `test_messaging`.

src/debugpy/common/json.py
```diff
@@ -106,6 +108,15 @@ def of_type(*classinfo, **kwargs):
     def validate(value):
         if (optional and value == ()) or isinstance(value, classinfo):
             return value
+        elif (
+            isinstance(value, str)
+            and all(x in string.digits + "." for x in value)
```
int19h, line 113: Might as well skip this check, and instead just fall through to the error branch below if neither conversion works.

src/debugpy/common/json.py
```diff
@@ -106,6 +108,15 @@ def of_type(*classinfo, **kwargs):
...
+            and any(issubclass(x, numbers.Number) for x in classinfo)
+        ):
+            try:
+                return int(value)
+            except ValueError:
+                return float(value)
```
int19h, line 119: One catch with this approach is that if the caller requests for `int`, they can now get a `float` instead, or vice versa. The former case is more dangerous, since the presence of the float will affect arithmetic down the line wherever that value flows. 

I would suggest that instead of doing a separate check and then a try/except series, to loop over the types in `classinfo`, and if they are a subtype of `Number`, then try to convert using that type object directly, and return if successful; if conversion fails, just move on to the next type. This will also take care of `bool` nicely.

giaco5988: I have implemented the above logic

int19h (approved):
Thanks again!

Issue #861, opened 2022-03-03
Support Python 3.11 frozen modules


fabioz:
See: https://github.com/fabioz/PyDev.Debugger/issues/213

YvesDup:
May be this issue can be reopen ? 
https://github.com/Microsoft/vscode-python/issues/941

int19h:
@YvesDup You should create an issue in the corresponding repository - this one is specifically for debugpy, but the issue you've linked to is about the way vscode-python *uses* debugpy.

gvanrossum:
A workaround is to add this to the configuation in launch.json:
```
"pythonArgs": ["-Xfrozen_modules=off"],
```

YvesDup:
@int19h Sorry :( for the misunderstood. I just proposed to add a new entriy to pass args to Python like proposed in that issue. But it already exists as the "pythonArgs" of lauch.json.

I am wondering what are the differences debugpy and pydev.debugger ? Are they both a debugger for Python or not or something more ? Thank

int19h:
PyDev.Debugger aka pydevd is one of the oldest Python debuggers around - it's the one originally made for Eclipse/PyDev, hence the name. It implements core functionality such as stepping and breakpoints. It runs inside the process being debugged, and thus, in general, it does things that are best done *from within* that process.

debugpy is basically a wrapper around pydevd that adds process lifetime management. It runs as a separate process from the debuggee, and thus, in general, it does things that are best done *from outside* of the process that is being debugged. So it can handle requests to launch processes for debugging or attach to running processes (and sets up pydevd as needed to handle those scenarios), and it keeps track of and coordinates processes in a multiprocess debug session. But it also does stuff such as stdout/stderr output capture.

To allow for easy deployment, debugpy vendors (i.e. embeds) all of its dependencies outside of the Python stdlib, including pydevd. So you don't have to worry about any of this while using it, but when we track work to be done, it can end up in either repo depending on where it was originally reported, and what the issue is.

int19h:
Fixed by https://github.com/microsoft/debugpy/pull/937

dc-avasilev:
not fixed https://github.com/microsoft/debugpy/pull/937/files#r1145404312

ergoithz:
related: https://github.com/python/cpython/issues/106723

PR #937, opened 2022-05-18
GH-861: Support-Python-3.11-frozen-modules

YvesDup:
Add "-X=frozen_modules=off" to cpython interpreter when version==3.11
see https://github.com/microsoft/debugpy/issues/861

### DIFF
--- a/src/debugpy/launcher/handlers.py
+++ b/src/debugpy/launcher/handlers.py
@@ -10,9 +10,6 @@ from debugpy import launcher
 from debugpy.common import json
 from debugpy.launcher import debuggee
 
-# see https://github.com/microsoft/debugpy/issues/861
-PYTHON_311 = sys.version_info[:2] == (3, 11)
-
 def launch_request(request):
     debug_options = set(request("debugOptions", json.array(str)))
 
@@ -40,11 +37,11 @@ def launch_request(request):
     python = request("python", json.array(str, size=(1,)))
     cmdline = list(python)
 
-    # see https://github.com/microsoft/debugpy/issues/861
-    if PYTHON_311:
-        cmdline += ["-X", "frozen_modules=off"]
-
     if not request("noDebug", json.default(False)):
+        # see https://github.com/microsoft/debugpy/issues/861
+        if sys.version_info[:2] >= (3, 11):
+            cmdline += ["-X", "frozen_modules=off"]
+
         port = request("port", int)
         cmdline += [
             os.path.dirname(debugpy.__file__),
--- a/src/debugpy/launcher/handlers.py
+++ b/src/debugpy/launcher/handlers.py
@@ -10,6 +10,8 @@ from debugpy import launcher
 from debugpy.common import json
 from debugpy.launcher import debuggee
 
+# see https://github.com/microsoft/debugpy/issues/861
+PYTHON_311 = sys.version_info[:2] == (3, 11)
 
 def launch_request(request):
     debug_options = set(request("debugOptions", json.array(str)))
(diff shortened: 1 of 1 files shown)


YvesDup:
Hi, I ran tests but had some errors with versions if Python < 3.11.  So I am wondering how to run tests with py3.11.
Could you help me ?

src/debugpy/launcher/handlers.py
```diff
@@ -10,6 +10,8 @@
 from debugpy.common import json
 from debugpy.launcher import debuggee
 
+# see https://github.com/microsoft/debugpy/issues/861
+PYTHON_311 = sys.version_info[:2] == (3, 11)
```
int19h, line 14: This should presumably also happen for all versions above 3.11 as well, at least until we can properly fix the underlying issue somehow.

We generally don't use named constants for such checks, and instead just do it in-place:
```py
if sys.version_info >= (3, 11): ...
```

src/debugpy/launcher/handlers.py
```diff
@@ -38,6 +40,10 @@ def property_or_debug_option(prop_name, flag_name):
     python = request("python", json.array(str, size=(1,)))
     cmdline = list(python)
 
+    # see https://github.com/microsoft/debugpy/issues/861
+    if PYTHON_311:
+        cmdline += ["-X", "frozen_modules=off"]
+
```
int19h, line 46: Given that the issue only arises when debugger has to deal with `__file__`, this should probably go under the conditional below (i.e. it shouldn't be applied for `"noDebug": true`).

src/debugpy/launcher/handlers.py
YvesDup: Ok, I follow your recommendations

src/debugpy/launcher/handlers.py
YvesDup: Test moved

src/debugpy/launcher/handlers.py
YvesDup: May be we can check if `os.path` is a frozen module with that line:

`'<frozen' in os.path.realpath.__code__.co_filename`

But I m not sure this module will be always frozen in next version of Python. 
I have to ask to a python core dev.

int19h (approved):
Thank you!

src/debugpy/launcher/handlers.py
YvesDup: > I think we can safely assume that this behavior will stay around for at least a few releases.
I agree
> But even if not, the nice thing about -X is that unrecognized switches are still valid, so it would continue working (as a no-op) even if the switch gets retired.
I hope so but I will check

gvanrossum:
Thanks! Much appreciated.

src/debugpy/launcher/handlers.py
```diff
@@ -39,6 +38,10 @@ def property_or_debug_option(prop_name, flag_name):
     cmdline = list(python)
 
     if not request("noDebug", json.default(False)):
+        # see https://github.com/microsoft/debugpy/issues/861
+        if sys.version_info[:2] >= (3, 11):
+            cmdline += ["-X", "frozen_modules=off"]
```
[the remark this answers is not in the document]
int19h: Space between the two is valid; indeed, it is canonical per the docs: https://docs.python.org/3/using/cmdline.html#cmdoption-X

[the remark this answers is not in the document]
dc-avasilev: > Space between the two is valid; indeed, it is canonical per the docs: https://docs.python.org/3/using/cmdline.html#cmdoption-X

it works for me only without a space, I can’t debug a flash application with such code as now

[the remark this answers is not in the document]
int19h: To clarify, does changing the code like this actually makes a difference for you? Or are you adjusting the flags manually via "args" in launch.json or directly from command line (if attaching to the process)?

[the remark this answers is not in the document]
dc-avasilev: I change code inside plugin, locally

PR #948, opened 2022-06-03
Auto-attach to subprocess with .zip targets. Fixes #945

### LINKED ISSUE #945: Auto-attach to subprocess with .zip target (named as .exe)

## Environment data

- debugpy version: 1.6.0
- OS and version: 21H2
- Python version (& distribution if applicable, e.g. Anaconda): Python 10.0.0
- Using VS Code or Visual Studio: VSCode

## Actual behavior

In the debugging session, running a subprocess with `python <shim>`  (where shim is an exe file in Scripts/ directory) fails with following message:

```
  File "d:\programs\python3-amd64\lib\tokenize.py", line 332, in find_cookie
    line_string = line.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte

During handling of the above exception, another exception occurred:

  File "c:\Users\ned\.vscode\extensions\ms-python.python-2022.6.2\pythonFiles\lib\python\debugpy\_vendored\pydevd\pydevd.py", line 3385, in <module>
    main()
  File "c:\Users\ned\.vscode\extensions\ms-python.python-2022.6.2\pythonFiles\lib\python\debugpy\_vendored\pydevd\pydevd.py", line 3378, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "c:\Users\ned\.vscode\extensions\ms-python.python-2022.6.2\pythonFiles\lib\python\debugpy\_vendored\pydevd\pydevd.py", line 2446, in run
    return self._exec(is_module, entry_point_fn, module_name, file, globals, locals)
  File "c:\Users\ned\.vscode\extensions\ms-python.python-2022.6.2\pythonFiles\lib\python\debugpy\_vendored\pydevd\pydevd.py", line 2453, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "c:\Users\ned\.vscode\extensions\ms-python.python-2022.6.2\pythonFiles\lib\python\debugpy\_vendored\pydevd\_pydev_bundle\_pydev_execfile.py", line 14, in execfile
    stream = tokenize.open(file)  # @UndefinedVariable
  File "d:\programs\python3-amd64\lib\tokenize.py", line 396, in open
    encoding, lines = detect_encoding(buffer.readline)
  File "d:\programs\python3-amd64\lib\tokenize.py", line 373, in detect_encoding
    encoding = find_cookie(first)
  File "d:\programs\python3-amd64\lib\tokenize.py", line 337, in find_cookie
    raise SyntaxError(msg)
SyntaxError: invalid or missing encoding declaration for 'D:\\<fakepath>\\.venv\\Scripts\\xonsh.exe'
```

## Expected  behavior

At minimum, it should run the shim without error. Preferably, it should run with subprocess debugging.

## Steps to reproduce:

1. Create a virtualenv with xonsh installed.
```
> poetry init -n
> poetry add xonsh prompt-toolkit
```

2. Locate the virtualenv in VSCode then start debugging with following launch.json:
```jsonc
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "xonsh",
            "type": "python",
            "request": "launch",
            "module": "xonsh",
            "justMyCode": true
        }
    ]
}
```

3. In the debugged xonsh shell, try: `$ python @($(which xonsh.exe).strip())`

fabioz:


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py
```diff
@@ -0,0 +1,353 @@
...
+
+__all__ = [
+    "run_module", "run_path",
+]
+
+
+# Note: fabioz: Don't use pkgutil (when handling caught exceptions we could end up
+# showing exceptions in pkgutil.get_imported (specifically the KeyError), so,
+# create a copy of the function we need to properly ignore this exception when
+# running the program.
```
int19h, line 33: What would be the consequence of using the one from stdlib? I'm curious because that's what debugpy does currently in launch scenarios.

fabioz: As a note, the patch changes debugpy to use this one instead of the one from the standard library too.

As far as debugpy is concerned, the main difference is that the stack related to runpy is no longer shown to the user (when justMyCode is false, which I believe is better for the user). Also, it was stopping inside of `runpy` when using `justMyCode` as false too along with `Raised Exceptions` multiple times...

For pydevd it had other repercussions because it treated the `_exec` inside of `pydevd.py` as the root, so, there were other issues related to unhandled exceptions which didn't work well (and a bunch of failing tests which didn't expect the changes in the stack) just by using `runpy` from the standard library with the pydevd entry point (it could also be fixed in pydevd. but it'd be more work than vendoring runpy and making that a part of the debugger).

PR #983, opened 2022-07-13
Fix #981: "repr was slow" warning is modal in Visual Studio

### LINKED ISSUE #981: "repr was slow" warning is modal in Visual Studio
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/HOW-TO-Disable-Pydevd-waring/10079495)._

---
\[regression\] \[worked-in:17.2.5\]  
Python Coding.  
When I try to inspect a var by hovering my cursor over the var name in the edit window (something like a breakpoint used). Previously, I got a nice text box with var info, but now, I also get an annoying window (tittle: Microsoft Visual Studio, content: pydevd warning:Computing repr of datas (list)was slow (took 0.17s) ).  
I must close it before returning to the edit window.  
Could you tell me how to disable it?  
I realy do not remember what I have done to cause this problem.

ps. I uploaded the screenshot via Additional Files.

---
#### Feedback Bot on 6/23/2022, 11:07 PM: 

(private comment, text removed)
#### Feedback Bot on 6/23/2022, 11:07 PM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

int19h:
Use the "console" category for warnings.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_json.py
@@ -293,7 +293,7 @@ class NetCommandFactoryJson(NetCommandFactory):
 
     @overrides(NetCommandFactory.make_warning_message)
     def make_warning_message(self, msg):
-        category = 'important'
+        category = 'console'
         body = OutputEventBody(msg, category)
         event = OutputEvent(body)
         return NetCommand(CMD_WRITE_TO_CONSOLE, 0, event, is_json=True)


fabioz:
@int19h the current PR will change it for all the messages, not only the ones related to `repr was slow`. I think most other messages should still be reported.

Also, I'm not sure this is the best fix... can't we fix VS itself? Otherwise, maybe we should do this only when the client is VS (in general, I'd like the user to know that the request is taking a long time due to his code being slow).

Or maybe the timeout could be a bit higher if the current one is being hit too often...

fabioz:
@int19h I agree that the modal notification is not acceptable for the `repr was slow` notification.

p.s.: If this needs a quick fix in the debugger side, I'm also ok in doing that change for all "important" messages if it detects that the client is VS (if you want I can do a PR for that). 

-- I guess showing it in the console for VS could be reasonable as it would probably be more visible than in the VSCode use-case (where the user is usually looking at the terminal, not the console).

PR #989, opened 2022-07-19
Custom subprocess environment variables for Python embedded

lmbelo:
Python embedded might not have environment variables set to Python paths and host application might not want to change its own environment settings.

For example (Android):
```
env = { 
  "LD_LIBRARY_PATH": shared_library_path,
  "PYTHONHOME": python_home,
  "PATH": python_executable_path,
  "TMPDIR": tmp
}
debugpy.configure(subProcessEnv=env)
```
Tested on Windows, Linux, MacOS (Intel and M1) and Android.

### DIFF
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -22,8 +22,8 @@ _tls = threading.local()
 _config = {
     "qt": "none",
     "subProcess": True,
-    "subProcessEnv": {},
     "python": sys.executable,
+    "pythonEnv": {},
 }
 
 _config_valid_values = {
@@ -191,9 +191,9 @@ def listen(address, settrace_kwargs):
 
         # On embedded applications, environment variables might not contain
         # Python environment settings.
-        subproc_env = _config.get("subProcessEnv")
-        if not bool(subproc_env):
-            subproc_env = None
+        python_env = _config.get("pythonEnv")
+        if not bool(python_env):
+            python_env = None
 
         # Adapter will outlive this process, so we shouldn't wait for it. However, we
         # need to ensure that the Popen instance for it doesn't get garbage-collected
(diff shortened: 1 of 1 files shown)


int19h:
Can you clarify the scenario for this? The name of the new property is `subProcessEnv`, but from the code, it is applied not to subprocesses in the debuggee (which the "subProcess" property corresponds too), but rather specifically to the debug adapter process. What's the purpose of the debug adapter using environment variables different from the debug server that spawned it?

lmbelo:
> Can you clarify the scenario for this? The name of the new property is `subProcessEnv`, but from the code, it is applied not to subprocesses in the debuggee (which the "subProcess" property corresponds too), but rather specifically to the debug adapter process. What's the purpose of the debug adapter using environment variables different from the debug server that spawned it?

The debug server might be an executable that embeds Python, then It might not contain the necessary Python's initialization paths on its environment paths. Once the subprocess would be a Python process, we need to initialize it with all required paths.
Did I clarify it? I will try to share some sample.

int19h:
Ah, I see, so this is basically a companion for the "python" configuration property then? I think calling it "pythonEnv" would make this clearer.

Out of curiosity, does setting the requisite variables directly in `os.environ` before doing `listen()` not work? I thought that any changes there would be automatically inherited by subprocesses.

lmbelo:
> Ah, I see, so this is basically a companion for the "python" configuration property then? I think calling it "pythonEnv" would make this clearer.
> 
> Out of curiosity, does setting the requisite variables directly in `os.environ` before doing `listen()` not work? I thought that any changes there would be automatically inherited by subprocesses.

Yes, I can change the host variables and it would be inherited by the subprocess, but maybe we don't want to change the host environment variables. It makes sense if i can launch the new process with custom variables. It worth mentioning that the host process can be a Delphi, or C# or whatever application language made, but not the Python executable itself. Then it wont have the Python environment paths settings.

lmbelo:
> > Ah, I see, so this is basically a companion for the "python" configuration property then? I think calling it "pythonEnv" would make this clearer.
> > Out of curiosity, does setting the requisite variables directly in `os.environ` before doing `listen()` not work? I thought that any changes there would be automatically inherited by subprocesses.
> 
> Yes, I can change the host variables and it would be inherited by the subprocess, but maybe we don't want to change the host environment variables. It makes sense if i can launch the new process with custom variables. It worth mentioning that the host process can be a Delphi, or C# or whatever application language made, but not the Python executable itself. Then it wont have the Python environment paths settings.

I totally agree to call it PythonEnv, though.

lmbelo:
Changed!

PR #1004, opened 2022-08-04
Fix #357, #1001

### LINKED ISSUE #357: "argsExpansion" does not do what it says in VSCode
See https://github.com/microsoft/vscode-python/issues/13264

The option might still be useful for other clients, but the issue highlights the fact that its current name is very client-behavior-specific, and doesn't accurately represent what it does: whether pass the arguments to the debuggee directly, or via the terminal. It should be called something like "argsFormat" or "argsPassing" instead.

Until that's decided, the option as is should not be considered stable, and the docs should reflect that.

### LINKED ISSUE #1001: Enable controlling shell expansion via "argsCanBeInterpretedByShell"
We now have the ability to let the user decide whether arguments passed to a debuggee in a terminal should be escaped for the shell. This is adopted in js-debug, https://github.com/microsoft/vscode-js-debug/issues/1335, also see https://github.com/microsoft/vscode/issues/149910, would love to see it adopted for python in August. In js-debug, the `args` property can be a string or an array of strings. If it's just a string, the args are not escaped. I can talk more about what it will take to  adopt this if needed.

int19h:
Fix #1001: Enable controlling shell expansion via "argsCanBeInterpretedByShell"

Fix #357: "argsExpansion" does not do what it says in VSCode

Treat non-array "args" in debug config as a request to prevent shell argument escaping and allow shell expansion.

Remove "argsExpansion".

### DIFF
--- a/src/debugpy/adapter/clients.py
+++ b/src/debugpy/adapter/clients.py
@@ -29,6 +29,7 @@ class Client(components.Component):
             "supportsVariablePaging": False,
             "supportsRunInTerminalRequest": False,
             "supportsMemoryReferences": False,
+            "supportsArgsCanBeInterpretedByShell": False,
         }
 
     class Expectations(components.Capabilities):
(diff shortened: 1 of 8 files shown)


int19h:
@karthiknadig I have manually tested this on the most recent VSCode insiders. 

The only change on vscode-python side is to update the JSON schema to allow "args" to be single string rather than an array. It would be nice if there's also a way to specify that this isn't supported for "internalConsole", but if not, no big deal - the error message should be clear enough if somebody tries that.

PR #1007, opened 2022-08-04
Add DEBUGPY_TRACE_DEBUGPY variable to allow debugpy to debug itself

rchiodo:
To allow using debugpy to debug itself (well at least the adapter), added a new environment variable that skips setting `dont_trace_start_patterns` and setting thread flags that hide threads.

### DIFF
--- a/src/debugpy/common/log.py
+++ b/src/debugpy/common/log.py
@@ -58,7 +58,7 @@ class LogFile(object):
                 platform.machine(),
                 platform.python_implementation(),
                 platform.python_version(),
-                64 if sys.maxsize > 2**32 else 32,
+                64 if sys.maxsize > 2 ** 32 else 32,
                 debugpy.__version__,
                 _to_files=[self],
             )
--- a/src/debugpy/common/util.py
+++ b/src/debugpy/common/util.py
@@ -150,16 +150,15 @@ def srcnameof(obj):
     return name
 
 
-def skip_trace():
-    """Returns True if tracing should be skipped for the current thread."""
+def hide_debugpy_internals():
+    """Returns True if the caller should hide something from debugpy."""
     return "DEBUGPY_TRACE_DEBUGPY" not in os.environ
 
 
-def disable_tracing(thread):
+def hide_thread_from_debugger(thread):
     """Disables tracing for the given thread if DEBUGPY_TRACE_DEBUGPY is not set.
     DEBUGPY_TRACE_DEBUGPY is used to debug debugpy with debugpy
     """
-    if skip_trace():
+    if hide_debugpy_internals():
         thread.pydev_do_not_trace = True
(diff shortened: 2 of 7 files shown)


src/debugpy/common/log.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, filename, file, levels=LEVELS, close_file=True):
                 platform.machine(),
                 platform.python_implementation(),
                 platform.python_version(),
-                64 if sys.maxsize > 2**32 else 32,
```
rchiodo, line 61: THis is a 'black' change. I just ran black on everything and it added this.

src/debugpy/launcher/handlers.py
```diff
@@ -10,6 +10,7 @@
 from debugpy.common import json
 from debugpy.launcher import debuggee
 
+
```
rchiodo, line 13: another black change

src/debugpy/common/util.py
```diff
@@ -148,3 +148,18 @@ def srcnameof(obj):
...
+
+
+def disable_tracing(thread):
+    """Disables tracing for the given thread if DEBUGPY_TRACE_DEBUGPY is not set.
+    DEBUGPY_TRACE_DEBUGPY is used to debug debugpy with debugpy
+    """
+    if skip_trace():
+        thread.pydev_do_not_trace = True
+        thread.is_pydev_daemon_thread = True
+        thread.daemon = True
```
int19h, line 165: This one doesn't belong here - it affects the behavior of the thread when the process shuts down (whether it waits for this thread or not before exiting), not whether it's traced by the debugger.

src/debugpy/common/util.py
```diff
@@ -148,3 +148,18 @@ def srcnameof(obj):
...
 
     return name
+
+
+def skip_trace():
+    """Returns True if tracing should be skipped for the current thread."""
+    return "DEBUGPY_TRACE_DEBUGPY" not in os.environ
+
+
+def disable_tracing(thread):
```
int19h, line 158: Let's call this something like `hide_thread_from_debugger` to distinguish it from other uses of "tracing" (that pertain to the debuggee) elsewhere in the code. 

Similarly for `skip_trace` - something like `hide_debugpy_internals` would be clearer, and will cover any other future tweaks we need to do to keep this working.

rchiodo: Sounds good, will do.

src/debugpy/common/util.py
rchiodo: Okay, wasn't sure. I'll put it back.

int19h (approved):

rchiodo:
I'm guessing the linux 3.9 test is flakey? It's checking thread count, so I thought it might be related.

PR #1052, opened 2022-09-12
Fix #1025: Debugger Does Not Enumerate ctypes Arrays

### LINKED ISSUE #1025: Debugger Does Not Enumerate ctypes Arrays
**Describe the bug**
The debugger in Visual Studio 2022 does not treat ctypes arrays as containers, and hovering over them or looking at them in the variables window does not show length or allow the items to be enumerated like a list or dict. This is a new problem in VS2022, and did not occur in VS2017 or VS2019, which is why I've only recently started using VS2022 and this problem is making it very unpleasant as I use ctypes extensively in my work.

Ordinarily I would post in a forum to ask if there is some setting that's set incorrectly (the settings are default for Python), but as there's no forum here I'm making a bug report.

**Steps to Reproduce**
For me this occurs any and all times I use ctypes arrays in any context, so I don't really have any specific instructions to reproduce. The only thing worth noting is that this occurs regardless of whether "Enable debugging of the Python standard library" is enabled (I mostly work with it enabled). Maybe read a random binary file into a bytearray and do something like

x = (ctypes.c_uint8 * len(file_bytes)).from_buffer(file_bytes)

Then hover over x (or look at it in the variables window). It will only show "special variables" and "protected variables".

**Expected behavior**
In past VS versions the debugger would display as if it was a list, showing the length and list of all elements.

**Configuration information**
VS Version: Community 2022 (64-bit) 17.3.0
PTVS version: 17.0.22179.3
Python version: 3.9.7 64-bit (x64)
Debugger Type: I don't think this is relevant for VS2022?

int19h:
Use tuple resolver for ctypes.Array

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_xml.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_xml.py
@@ -91,6 +91,12 @@ def _create_default_type_map():
     except:
         pass
 
+    try:
+        from ctypes import Array
+        default_type_map.append((Array, pydevd_resolver.tupleResolver))
+    except:
+        pass
+
     if frame_type is not None:
         default_type_map.append((frame_type, pydevd_resolver.frameResolver))
 
--- a/src/debugpy/_vendored/pydevd/tests_python/test_resolvers.py
+++ b/src/debugpy/_vendored/pydevd/tests_python/test_resolvers.py
@@ -321,3 +321,42 @@ def test_tuple_resolver_mixed():
     check_len_entry(len_entry, (GENERATED_LEN_ATTR_NAME, 2))
     assert contents_debug_adapter_protocol == [
         ('some_value', 10, '.some_value'), ('0', 1, '[0]'), ('1', 2, '[1]'), ]
+
+
+def test_tuple_resolver_ctypes():
+    import ctypes
+    from _pydevd_bundle.pydevd_xml import get_type
+
+    array_type = ctypes.c_int32 * 3
+    array_value = array_type(1, 2, 3)
+    _, _, resolver = get_type(array_value)
+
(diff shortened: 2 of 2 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_xml.py
```diff
@@ -91,6 +91,12 @@ def _create_default_type_map():
     except:
         pass
 
+    try:
+        from ctypes import Array
```
fabioz, line 95: Seems good to me, but it'd be nice to have a test too...

fabioz (approved):

r3m0t:
I also have a change for the resolver I want to upstream. Are you accepting PRs or should I go to the pydev.debugger repo?

My change removes useless attributes like x.index when x is a local of type `list`.

fabioz:
@r3m0t

Ideally changes in the debugger implementation would go to the `pydev.debugger` repo initially (because the tests for `pydev.debugger` are only run in that repo, not in this one which is geared towards the integration of `pydevd` with the debug adapter protocol).

Changes should ideally have a test too.

As a note, for the particular feature you're describing, please create an issue first (on the `debugpy` repo is fine) as for your use-case I believe the filtering capabilities should already suffice.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_xml.py
fabioz: I see what you mean (it tests the resolver themselves, not the resolution process), but I think that using `test_resolvers` for that is also a good place (the API to be used is `get_type(obj)` where the last item is the actual resolver... there's only one test there that calls this API, but I think we can extend that same module to add this test).

fabioz (approved):

PR #1073, opened 2022-10-03
Fix #1018: setup.py doesn't fail if compiler is missing

### LINKED ISSUE #1018: setup.py doesn't fail if compiler is missing
Build log:
```
PS C:\git\debugpy> py -3.10-64 setup.py bdist
C:\Python\3.10-64\python.exe
Compiling extension modules (set SKIP_CYTHON_BUILD=1 to omit)
running build_ext
building '_pydevd_bundle.pydevd_cython' extension
error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
running bdist
running bdist_dumb
running build
running build_py
creating build
creating build\lib.win-amd64-cpython-310
creating build\lib.win-amd64-cpython-310\debugpy
...
```
Note that it failed to build the pydevd extension modules, but still proceeded to package debugpy. If you don't pay close attention to the output, you might not even notice that said extension modules are missing.

The problem is that modules are built in a subprocess of setup.py, and we don't check the exit code of said subprocess:
https://github.com/microsoft/debugpy/blob/8b5eeee7e0b1678e701b07d94e6c3cf07b923597/setup.py#L36-L45

int19h:
Add setup.py custom command to build pydevd Cython extensions.

Expose environment variable to control build failure mode if extensions fail to compile.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,27 +33,6 @@ def get_buildplatform():
     return None
 
 
-def cython_build():
-    print("Compiling extension modules (set SKIP_CYTHON_BUILD=1 to omit)")
-    subprocess.call(
-        [
-            sys.executable,
-            os.path.join(PYDEVD_ROOT, "setup_pydevd_cython.py"),
-            "build_ext",
-            "-i",
-        ]
-    )
-
-
-def iter_vendored_files():
-    # Add pydevd files as data files for this package. They are not
-    # treated as a package of their own, because we don't actually
-    # want to provide pydevd - just use our own copy internally.
-    for project in debugpy._vendored.list_all():
-        for filename in debugpy._vendored.iter_packaging_files(project):
-            yield filename
-
-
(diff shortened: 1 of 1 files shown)


fabioz (commented):
It seems that this PR broke the compilation somehow (as the tests related to checking if frame eval is being used are failing...).

For the CI, where is `REQUIRE_CYTHON_BUILD` set to `1`?

int19h:
It broke because the binaries aren't packaged properly - I'm still working on this. I'll update the CI and release pipelines once that's done.

fabioz (approved):

int19h:
I updated the PR to fix the inclusion problem. No custom command anymore, but it still avoids compiling pydevd outside of build commands.

PR #1077, opened 2022-10-05
Fix socket problem when running with Emscripten CPython build

rchiodo:
SO_REUSEADDR is not supported in EMSCRIPTEN


rchiodo:
> Given the number of platforms we're testing for at this point, I think it's time to switch to feature tests:
> 
> ```python
>     try:
>         sock.setsockopt(socket.SOL_SOCKET, socket.SO_EXCLUSIVEADDRUSE, 1)
>     except AttributeError:
>         try:
>            ...
> ```

That does sound better. I'll try that instead.

int19h (approved):

PR #1082, opened 2022-10-10
Fix #1081: no need to call os.setsid() on session leader when attaching

### LINKED ISSUE #1081: os.setsid() potentially fails in adapter/__main__.py if attaching to session leader remotely
## Environment data

- debugpy version: 1.6.3
- OS and version: CentOS 8
- Python version (& distribution if applicable, e.g. Anaconda): python3.9
- Using VS Code or Visual Studio: vscode

## Actual behavior

Debugger client connection times out

## Expected  behavior

Debugger client can successfully attach to injected  debugpy server

## Steps to reproduce:
1. launch a python program with `setsid python ....`
2. Attach to it using a remote configuration from vscode, check stderr when attaching to see the actual error message

fecjanky:
If os.setsid() raises exception when attaching to a session leader process then the debugger client won't be able to attach to the remote process anymore

### DIFF
--- a/src/debugpy/adapter/__main__.py
+++ b/src/debugpy/adapter/__main__.py
@@ -30,7 +30,10 @@ def main(args):
             # On POSIX, we need to leave the process group and its session, and then
             # daemonize properly by double-forking (first fork already happened when
             # this process was spawned).
-            os.setsid()
+            # NOTE: if process is already the session leader, then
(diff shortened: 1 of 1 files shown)


int19h (approved):

int19h:
Thank you!

PR #1094, opened 2022-10-21
[WIP] Add breakpoint location discovery

tonybaloney:
Implements breakpoint locations request

TODO: 
- [x] Emit whole line
- [ ] Emit column fragments
- [ ] Emit multi-lines where possible?
- [ ] Test


tonybaloney:
screenshot 2022-10-21 at 14 04 35

This is how it looks in VS Code (with no changes to the extension)

fabioz:
The leftover print is breaking the tests... 

I think we shouldn't implement column fragments with the current infrastructure available from Python itself (I'd wait for something as PEP 669 to be available to implement this in a proper way).

PR #1121, opened 2022-11-10
Add column information to exception tracebacks. Fixes #1099

### LINKED ISSUE #1099: PEP657 Column position of raised exceptions
In Debugpy, exceptions raised in Python 3.11 are missing the fine-grained position:

For example, in the following code:

```python
d = {
    "x": {
        "y": {
            "z": {
                "theta": 1
            }
        }
    }
}

result = d["x"]["y"]["z"]["beta"]
```

On the CLI I get the position of the KeyError offset in the line underlined using [PEP657](https://peps.python.org/pep-0657/) API:

```command
$ python exploding_kitten.py
...
  File "/Users/anthonyshaw/tmp/exploding_kitten.py", line 11, in <module>
    result = d["x"]["y"]["z"]["beta"]
             ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'beta'
```

But in VS Code it only highlights the line and not the position of the exception, making it harder to work out where the error came from.

screenshot 2022-10-25 at 13 33 54

Please can PEP657 be implemented in debugpy so that the Stopped event sent back via DAP include the position (start col, end col) of the raised exception.

fabioz:


int19h:
Something weird is going on with macOS tests that's unrelated to the changes.

int19h (approved):

PR #1207, opened 2023-02-07
Fix #1074: Use "startDebugging" request for subprocesses

### LINKED ISSUE #1074: Use "startDebugging" request for subprocesses
See:
https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_73.md#new-startdebugging-reverse-request

int19h:
Use the request if client advertises the "supportsStartDebuggingRequest" capability.

### DIFF
--- a/src/debugpy/adapter/clients.py
+++ b/src/debugpy/adapter/clients.py
@@ -30,6 +30,7 @@ class Client(components.Component):
             "supportsRunInTerminalRequest": False,
             "supportsMemoryReferences": False,
             "supportsArgsCanBeInterpretedByShell": False,
+            "supportsStartDebuggingRequest": False,
         }
 
     class Expectations(components.Capabilities):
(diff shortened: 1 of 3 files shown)


int19h:
For "attach" + "connect", the expectation is that the client will interpret it to directly connect to the adapter that is already listening on the specified port - for multiproc, it'll be the same adapter for all sessions (although that is considered an implementation detail from client perspective).

mfussenegger:
> For "attach" + "connect", the expectation is that the client will interpret it to directly connect to the adapter that is already listening on the specified port

If I try that I manage to establish the connection and send `initialize` and also receive a response to that, but then when I send the `attach` request the server closes the connection and the process of the root session terminates.

Is there a way to get more logs? I can't use the `--log-to` option with `python -m debug.py.adapter`.

int19h:
Set the `DEBUGPY_LOG_DIR` environment variable to where you want the logs; it works for all the debugpy subprocesses.

mfussenegger:
Thanks a lot. That helped - turns out I forgot to include the `success` payload in the `startDebugging` response and that caused an error.

Sorry for the noise.

PR #1219, opened 2023-02-18
Set up the proper execution environment. so that Pylance knows the cor…

heejaechang:
Set up the proper execution environment so that Pylance knows the correct import root for the import statement.

This will let pylance know that import root is "src", so when some feature generates "import statement", it won't include "src"

ex) rather than `import src.debugpy` it will generate `import debugpy`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,10 +15,10 @@ exclude = '''
 
 [tool.pyright]
 pythonVersion = "3.7"
-include = ["src"]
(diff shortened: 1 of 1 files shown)


int19h:
Hm, but then `tests` is a top-level package, and it does do absolute imports like `from test.patterns import some` a lot. Will that be affected?

heejaechang:
updated to handle `from test.patterns import some` case.

int19h (approved):

PR #1244, opened 2023-03-17
GH-1230 Use find_namespace_packages()

### LINKED ISSUE #1230: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('pydevd_plugins.extensions')`
Causes test_run_relative_path fails, and the failure is consistent across all platforms, e.g. https://dev.azure.com/debugpy/debugpy/_build/results?buildId=3594&view=logs&j=4b649afe-af30-5b13-56ee-823ad097620f&t=1628023b-a9e3-551b-728a-f87538244b57&l=2498

Seems to be due to:
```
File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_extension_utils.py", line 5, in <module>
E               import pydevd_plugins.extensions as extensions
E             File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\_vendored\pydevd\pydevd_plugins\extensions\__init__.py", line 4, in <module>
E               __import__('pkg_resources').declare_namespace(__name__)
E             File "D:\a\1\s\.tox\py39\lib\site-packages\pkg_resources\__init__.py", line 2298, in declare_namespace
E               warnings.warn(msg, DeprecationWarning, stacklevel=2)
E           DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('pydevd_plugins.extensions')`.
E           Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

darrenmeehan:
I'm putting this up for review ahead of testing to allow some conversation on what the desired approach for addressing #1230  should be.

Fixes #1230

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -177,14 +177,12 @@ if __name__ == "__main__":
             "License :: OSI Approved :: MIT License",
         ],
         package_dir={"": "src"},
-        packages=[
-            "debugpy",
-            "debugpy.adapter",
-            "debugpy.common",
-            "debugpy.launcher",
-            "debugpy.server",
-            "debugpy._vendored",
-        ],
+        packages=setuptools.find_namespace_packages(
+            include=[
+                "debugpy",
+                "debugpy.*",
+            ]
+        ),
         package_data={
             "debugpy": ["ThirdPartyNotices.txt"],
             "debugpy._vendored": [
--- a/src/debugpy/_vendored/pydevd/pydevd_plugins/__init__.py
+++ b/src/debugpy/_vendored/pydevd/pydevd_plugins/__init__.py
@@ -1,8 +1,0 @@
-import warnings
-with warnings.catch_warnings():
-    warnings.filterwarnings("ignore", category=DeprecationWarning)
-    try:
-        __import__('pkg_resources').declare_namespace(__name__)
-    except ImportError:
-        import pkgutil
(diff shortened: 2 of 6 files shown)


darrenmeehan:

setup.py
```diff
@@ -177,14 +177,12 @@ def tail_is(*suffixes):
...
-            "debugpy.adapter",
-            "debugpy.common",
-            "debugpy.launcher",
-            "debugpy.server",
-            "debugpy._vendored",
-        ],
+        packages=setuptools.find_namespace_packages(
+            include=[
+                "debugpy",
+                "debugpy.*",
```
darrenmeehan, line 183: I need to confirm whether both entries are required here.

int19h:
Given the implications of the change, I think this should be a PR on https://github.com/fabioz/PyDev.Debugger for maximum visibility.

As I understand it, switching the pydevd packages from classic to init-less is going to affect all the pydevd extensions out there that use the old API. But, conversely, the extensions cannot update until pydevd itself does. So the hard part here isn't the fix itself, but rather coordinating the change across the ecosystem.

That said, depending on the expected timeline, we might want to apply a local fix temporarily.

int19h:
An equivalent fix has been merged.

PR #1279, opened 2023-04-20
Create auto-label.yml

StellaHuang95:
Add a bot that automatically removes the "waiting for response" label and adds a "user responded" label if the original poster replies to avoid accidentally close active issues.

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,0 +1,18 @@
+name: Auto-label when user responds
+permissions:
+  issues: write
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  run-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,18 @@
+name: Auto-label when user responds
```
karthiknadig, line 1: You might want to look at this for a better approach: https://github.com/microsoft/vscode-python/blob/main/.github/workflows/triage-info-needed.yml

It could be that more than one user reports the issue. so the original commenter many not be the one who responds with the details.

StellaHuang95: Thanks I will take a look😊

karthiknadig (approved):

PR #1292, opened 2023-05-15
Add python 3.11 to test matrix

### LINKED ISSUE #1107: Add Python 3.11 to the ci
Now that Python 3.11.0 final is available, we should run tests with it in the ci.

AdamYoblick:
Fixes #1107


int19h (approved):
Needs merging with the similar fix for column numbers in main but otherwise LGTM.

int19h:
An equivalent change made on top of updated main is now checked in.

PR #1319, opened 2023-06-21
Remove the EPL from the list of licenses

brettcannon:
https://github.com/microsoft/debugpy/blob/main/LICENSE only lists MIT.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,7 +173,6 @@ if __name__ == "__main__":
             "Operating System :: Microsoft :: Windows",
             "Operating System :: MacOS",
             "Operating System :: POSIX",
-            "License :: OSI Approved :: Eclipse Public License 2.0 (EPL-2.0)",
             "License :: OSI Approved :: MIT License",
         ],
         package_dir={"": "src"},


brettcannon:
Or is EPL listed because of debugpy (which is dual-licensed under Apache which is more permissive if you can list that as the other license)?

AdamYoblick:
> Or is EPL listed because of debugpy (which is dual-licensed under Apache which is more permissive if you can list that as the other license)?

I'm actually not sure. @gramster or @karthiknadig would know better. I don't want to tell you the wrong thing 😄

gramster:
This is a throwback to when pydevd was EPL only. Fabio added Apache 2.0 license so we can remove EPL, but I'm not sure if that means we need to add Apache 2.0.

AdamYoblick:
Who do I contact to find out if we have to add apache? CELA?

karthiknadig:
I have sent Adam the details on who to talk to and where to discuss this offline.

AdamYoblick (approved):

PR #1354, opened 2023-08-14
Don't assume column number to be 1

int19h:
This fixes all the tests currently broken on Python 3.11, except for gevent.

### DIFF
--- a/tests/patterns/dap.py
+++ b/tests/patterns/dap.py
@@ -26,7 +26,7 @@ def source(path, **kwargs):
     return some.dict.containing(d)
 
 
-def frame(source, line, **kwargs):
+def frame(source, line, column=some.int, **kwargs):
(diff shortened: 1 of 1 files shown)


karthiknadig (approved):

AdamYoblick:
@int19h I have a pull request pending to add 3.11 to the test matrix, I made a similar change: https://github.com/microsoft/debugpy/pull/1292/files#diff-91e4a5a86d8aa08f799fdef0de1d33c4fa73e1fcd0bf7eee6084838ece276972

Should I just revert mine?

AdamYoblick (approved):

int19h:
I think `some.int` is going to be somewhat more reliable since there may not be a guarantee that the correct column number is the first non-whitespace in all cases (although it probably is for the actual tests that we have rn).

PR #1399, opened 2023-09-22
Add some type annotation in log

heejaechang:


src/debugpy/common/log.py
```diff
@@ -68,12 +87,13 @@ def levels(self):
         return self._levels
 
     @levels.setter
-    def levels(self, value):
+    def levels(self, value: Union[Sequence[LevelTypes], Callable[..., bool]]):
```
int19h, line 90: Let's just ditch this misuse of `all` and use a string literal instead.

src/debugpy/common/log.py
```diff
@@ -143,12 +163,13 @@ def write(level, text, _to_files=all):
...
 
 
-def write_format(level, format_string, *args, **kwargs):
+def write_format(level: LevelTypes, format_string: str, *args: Any, **kwargs: Any):
     # Don't spend cycles doing expensive formatting if we don't have to. Errors are
     # always formatted, so that error() can return the text even if it's not logged.
     if level != "error" and level not in _levels:
         return
 
+    text = ""
```
int19h, line 172: Is it meant as a type assertion, or to silence a non-initialized variable warning? If the latter, it might be clearer to do `raise reraise_exception()` to indicate to the linter that there's no exit out of the `except` block.

src/debugpy/common/log.py
```diff
@@ -162,7 +183,7 @@ def write_format(level, format_string, *args, **kwargs):
 warning = functools.partial(write_format, "warning")
 
 
-def error(*args, **kwargs):
+def error(*args: Any, **kwargs: Any):
```
int19h, line 186: For args and kwargs, should these be list of Any and dict of Any, respectively?

src/debugpy/common/log.py
```diff
@@ -177,8 +198,8 @@ def error(*args, **kwargs):
     return AssertionError(write_format("error", *args, **kwargs))
 
 
-def _exception(format_string="", *args, **kwargs):
-    level = kwargs.pop("level", "error")
+def _exception(format_string="", *args: Any, **kwargs: Any):
```
int19h, line 201: Is there any way to designate `format_string` here and elsewhere as "`str` but must be a literal"? It should never be called with non-literal values as first arguments due to possibility of format string injection attack, and it would be nice to enforce this on type levels.

int19h:
Let's close this for now, and get back to it in the new debugger implementation.

PR #1404, opened 2023-10-03
Fix #1337: Get port info from debugpy

### LINKED ISSUE #1337: Report sockets opened by debugpy
In order to implement the [Port Attributes API](https://github.com/microsoft/vscode-python/issues/21292). We need to know the port info that debugpy generates for the debugging process. You can send this as a log message that we can track and then disable port forwarding on those ports.

int19h:
Send "debugpySockets" event with information about opened sockets when clients connect and whenever ports get opened or closed.


src/debugpy/adapter/clients.py
```diff
@@ -701,6 +702,24 @@ def disconnect_request(self, request):
...
         super().disconnect()
 
+    def report_sockets(self):
+        sockets = [
+            {
+                "host": host,
+                "port": port,
+                "internal": listener is not clients.listener,
+            }
+            for listener in [clients.listener, launchers.listener, servers.listener]
```
AdamYoblick, line 712: What's going on with the lack of indenting in this for loop?

src/debugpy/adapter/clients.py
```diff
@@ -701,6 +702,24 @@ def disconnect_request(self, request):
     def disconnect(self):
         super().disconnect()
 
+    def report_sockets(self):
+        sockets = [
```
AdamYoblick, line 706: I'm not familiar with this list initialization syntax. Is there supposed to be a comma after the dict? Does the code afterwards resolve to a list and it gets stuffed into the sockets parent list?

AdamYoblick (approved):
Approved with some questions :)

src/debugpy/adapter/clients.py
int19h: This is a Python list comprehension. C# equivalent would be something like:
```cs
from listener in ...
where listener is ...
let (host, port) = listener.getsockname()
select {...}
.ToList()
```

Issue #1401, opened 2023-09-27
debugpy release with Python 3.12 support


int19h:
Test runs already pass, so this is just verifying that things work as they should.

PR #1407, opened 2023-10-06
Flag Python 3.12 support in classifiers

tonybaloney:
Related #1401

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -169,6 +169,7 @@ if __name__ == "__main__":
             "Programming Language :: Python :: 3.9",
             "Programming Language :: Python :: 3.10",
             "Programming Language :: Python :: 3.11",
+            "Programming Language :: Python :: 3.12",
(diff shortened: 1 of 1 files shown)


karthiknadig (approved):

int19h:
Note that we haven't tested this extensively yet beyond the test run itself, so let's merge this after such testing is complete.

tonybaloney:
Do you have a list of test scenarios? I'm running a stack of 3.12 projects, so happy to test some and crowd source the rest

int19h:
Most areas are covered by the tests reasonably well but attach-to-PID is the one that is tricky to test (adversarial code injection into a process is inherently hacky and prone to random failures). Since it's also the area most likely to be affected by changes in CPython internals, that would be my biggest concern for Python 3.12.

int19h:
This has been merged in a separate PR that also updated the build pipeline

Issue #1475, opened 2023-11-29
Multi-thread debugging, the main thread is suspended, the child thread can not be suspended? 


YuhongZeng:
Multi-thread debugging, the main thread is suspended, the child thread can not be suspended? You can perform continue, parse, and step control on each thread.
Our expected multi-threaded debugging feature for VSCode includes the following:
 1. The main thread is suspended, but the sub-threads are not suspended at the same time.
 2. Each thread can be separately controlled to continue, pause, and step. 
The following video demonstrates the implementation of this multi-threaded debugging feature.

YuhongZeng:
I understand having the other threads running ended up incredibly confusing in most scenarios. I believe that this confusion is mainly caused by the vendored PyDev.Debugger. Here are the reasons we think this is happening.
1.

https://github.com/microsoft/debugpy/blob/ef9a67fe150179ee4df9997f9273723c26687fab/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L728C1-L813C81

YuhongZeng:
1. Breakpoint hits cause all threads to be suspended
- When setting breakpoints, pydevd forcibly set suspend policy to ALL.
https://github.com/microsoft/debugpy/blob/ef9a67fe150179ee4df9997f9273723c26687fab/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L728C1-L813C81

- When ` suspend_policy = 'ALL'`  causes the `suspend_other_threads` parameter to be set to `True`.
https://github.com/microsoft/debugpy/blob/ef9a67fe150179ee4df9997f9273723c26687fab/src/debugpy/_vendored/pydevd/pydevd.py#L1845C1-L1892C60

- The `self.multi_threads_single_notification=True` sets suspend_other_threads to True when `suspend_policy ='NONE'. `multi_threads_single_notification`cannot be set to False. If it is set to False, Development Tools cannot accept debugging signals from pydevd.

`
        if not suspend_other_threads and self.multi_threads_single_notification:
            # In the mode which gives a single notification when all threads are
            # stopped, stop all threads whenever a set_suspend is issued.
            suspend_other_threads = True
`

YuhongZeng:
2. StoppedEvent and ContinuedEvent returned by pydevd to Development Tools
[...]

PR #1512, opened 2024-02-01
[sys.monitoring] Breakpoints

int19h:
Refactor all tracing into a separate subpackage.

Implement conditional and hit-count breakpoints and logpoints.


src/debugpy/server/tracing/tracer.py
```diff
@@ -0,0 +1,417 @@
...
+            pass
+
+        @atexit.register
+        def disable():
+            self.debug = self.info = self.warning = self.error = self.exception = nop
+
+    def debug(self, *args, **kwargs):
+        # self.log.debug("{0}", *args, **kwargs)
+        # print(*args)
+        pass
```
gramster, line 40: Is this intentional? Any reason to keep the commented out code? If it is used sometimes, should it be conditional on a flag?

src/debugpy/server/tracing/tracer.py
```diff
@@ -0,0 +1,417 @@
...
+        self.log.error("{0}", *args, **kwargs)
+
+    def exception(self, *args, **kwargs):
+        self.log.exception("{0}", *args, **kwargs)
+
+
+class Tracer:
+    """
+    Singleton that manages sys.monitoring callbacks for this process.
+    """
```
gramster, line 58: A lot of this code used to be in tracing.py. Is there a reason to put it in a singleton class? I understand the needs for singletons in languages that force everything to be classes, but why in Python?

Note - I'm not saying undo it, just wondering what informed the design here.

gramster (commented):
Can't say I understand it all but I have a few questions.

src/debugpy/server/tracing/tracer.py
int19h: I'm still trying to figure out a good way to log at this level of detail without a significant perf penalty.

src/debugpy/server/tracing/tracer.py
int19h: Two reasons. First one is that experimenting with it in the prototype leads me to believe that writing similar code as global functions is more fraught with footguns in general, thanks to the need to declare all writable shared state as `global` explicitly in each function. More importantly, having it all in a class allows the state to be quickly reset - this isn't currently used because reattach isn't fully implemented yet, but the intent is to throw away all those singletons on disconnect, and recreate them on reconnect.

src/debugpy/server/tracing/tracer.py
```diff
@@ -0,0 +1,417 @@
...
+        self.log.info("sys.monitoring tracing callbacks registered.")
+
+    def pause(self, thread_ids: Iterable[int] = None):
+        """
+        Pause the specified threads, or all threads if thread_ids is None.
+        """
+        if thread_ids is None:
+            # Pausing is async, so additional threads may be spawned even as we are
+            # trying to pause the ones we currently know about; iterate until all
+            # known threads are paused, and no new threads appear.
```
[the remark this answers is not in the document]
int19h: Without this loop, the race will result in not all threads being paused occasionally, which breaks DAP invariants when "pause" request reports all threads to be paused (and VS in particular *requires* this), so it is crucial.

The loop should be robust in the face of any real-world multithreaded code. One case where in theory it can never terminate is if the code running is some kind of fork bomb, spawning threads which immediately spawn more threads etc; although even in this case, given that spawning a thread is significantly more expensive than a single iteration of the loop, I would expect it to converge eventually.

src/debugpy/server/tracing/tracer.py
```diff
@@ -0,0 +1,417 @@
...
+                    del self._steps[thread.id]
+                    self._pause_ids.add(thread.id)
+                    self._cvar.notify_all()
+                    stop_reason = "step"
+
+        if stop_reason is not None:
+            # Even if this thread is pausing, any debugpy internal code on it should
+            # keep running until it returns to user code; otherwise, it may deadlock
+            # if it was holding e.g. a messaging lock.
+            print(frame_obj.f_globals.get("__name__"))
```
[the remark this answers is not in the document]
int19h: Nope; debugging leftover.

src/debugpy/server/tracing/tracer.py
int19h: For this particular class, the third reason is as explained in the comment. With top-level functions, it'd require an import in each function, and it would also be slower because those imports would run (even if just to check `sys.modules` and bail if that module is already imported) every time the function executes.

src/debugpy/server/tracing/tracer.py
gramster: Dump to a string buffer with occasional flushes?

src/debugpy/server/tracing/tracer.py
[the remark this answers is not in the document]
gramster: I was thinking more of a thread starting as the loop gets exited, so not everything is paused on exit.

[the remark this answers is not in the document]
int19h: Yes, that's a good point. Now that I think about it, it's possible for a thread to be initializing, but not yet at the point where `enumerate()` would see it. This needs an additional check in line tracing so that any new thread that is spawned while paused also gets paused as soon as it starts running Python code.

src/debugpy/server/tracing/tracer.py
int19h: Yep, I'm going to try that, but as a separate PR.

src/debugpy/server/tracing/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+        "%": lambda expected_count, count: count % expected_count == 0,
+    }
+
+    hit_condition: str
+    _count: int
+    _operator: Callable[[int, int], bool]
+
+    def __init__(self, hit_condition: str):
+        self.hit_condition = hit_condition
+        m = re.match(r"([<>=]+)?(\d+)", hit_condition)
```
karthiknadig, line 331:
```suggestion
        m = re.match(r"([%<>=]+)?(\d+)", hit_condition)
```

src/debugpy/server/tracing/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+    """
+
+    expression: str
+    """Python expression that must evaluate to True for the breakpoint to be triggered."""
+
+    _code: CodeType
+
+    def __init__(self, breakpoint: "Breakpoint", expression: str):
+        self.expression = expression
+        self._code = compile(
```
karthiknadig, line 278: Should we have a syntax error catch here and re-raise with a message that the syntax error was in the condition? Otherwise, it might look like a generic syntax error.

src/debugpy/server/tracing/__init__.py
```diff
@@ -0,0 +1,533 @@
...
+    message: str
+    """The message to be logged. May contain expressions in curly braces."""
+
+    _code: CodeType
+    """Compiled code object for the f-string corresponding to the message."""
+
+    def __init__(self, breakpoint: "Breakpoint", message: str):
+        self.message = message
+        f_string = "f" + repr(message)
+        self._code = compile(f_string, f"breakpoint-{breakpoint.id}-logMessage", "eval")
```
karthiknadig, line 364: Same comment as for the `Condition` compile.

karthiknadig (changes requested):

src/debugpy/server/tracing/tracer.py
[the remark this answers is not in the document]
int19h: So I tried to implement this and ran into two fundamental issues.

The first one is that DAP is simply underspecified for various corner cases that arise when pause/unpause is supported both for all threads and for specific threads - e.g. if all threads are paused, but then a single thread is resumed, and that thread spawns a new thread, should the new thread be paused or resumed? There are similar issues with evaluate etc.

The other problem is that VS (unlike VSCode) simply doesn't support individual thread requests and requires all threads to be paused/resumed, otherwise it gets out of sync. And while DAP has a way for the client to communicate that it supports single-thread-continue, there's no equivalent for single-thread-pause; thus we have to always assume that the client is VS (or something with the same limitation) and always pause all threads.

Between the two, I'm going to simplify this by tracking paused/unpaused state globally and not supporting it for individual threads. This means that in VSCode, if you right-click a single thread and ask to unpause it, it'll unpause all threads in practice. I believe this is acceptable as that is how pydevd-based debugpy behaves already. While there has been a user request to add per-thread unpause (#1475), I don't think we can do a good job there until DAP spec is extended & clarified sufficiently for us to implement; https://github.com/microsoft/debug-adapter-protocol/issues/355 is the issue tracking this on DAP side.

src/debugpy/server/tracing/__init__.py
int19h: Since it's checking the operators against a predefined list on the next step anyway, I'll simplify it down to any sequence of non-digits followed by digits.

src/debugpy/server/tracing/__init__.py
int19h: You're right, it should catch. But it can't re-raise because that would crash the debuggee. Intuitively it feels like this should be an error response to the "breakpoints" request so that the user immediately gets an error, and this is what I implemented, but VSCode doesn't seem to be propagating that failure to the user in practice. I'll investigate further, but let's track it as a separate issue.

PR #1524, opened 2024-02-18
fixed 'PathDistribution' object has no attribute 'name'

cctv130:


src/debugpy/common/log.py
```diff
@@ -359,7 +359,7 @@ def report_paths(get_paths, label=None):
         report("Installed packages:\n")
         try:
             for pkg in importlib_metadata.distributions():
-                report("    {0}=={1}\n", pkg.name, pkg.version)
+                report("    {0}=={1}\n", pkg.metadata['Name'], pkg.version)
```
int19h, line 362: Does this actually fix the issue when name is missing? Looking at implementation of `importlib.metadata.Distribution`, `pkg.name` just calls `pkg.metadata['Name']`, so the effect should be the same.

int19h (approved):

AdamYoblick:
@cctv130 Just waiting on some PR checks that stalled. Re-opening the PR to trigger them again.

AdamYoblick:
> How is this triggered debugpy-test-automation

I'm looking into this.

PR #1549, opened 2024-03-28
PyQt6 support

### LINKED ISSUE #1112: Support pyqt6
## Environment data

- debugpy version: 1.5.1
- OS and version: GNU/Linux Pop!_OS 22.04
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.9.13, Anaconda 22.9.0
- Using VS Code or Visual Studio: VS Code (Codium) 1.72.2

## Actual behavior
The debugger just "stops" after importing matplotlib, with no sign or crash or error. Just seems to be complete. See the screencast at the end.

## Expected  behavior
Debugging continues to the next line

## Steps to reproduce:
1. Create a conda environment with python 3.9
2. Install matplotlib and PyQt6 through pip
3. Install jupyter through conda
4. Create a file, and import matplotlib (and anything on the next line)
5. Put a breakpoint on the matplotlib import
6. Debug the file
7. When reaching the breakpoint, Step Over.

To make it easier, here are the minimal reproducible example I could generate:
```bash
conda create -y -n bug-report python=3.9
conda activate bug-report
pip install python-xlib matplotlib pyqt6 pillow
conda install -y jupyter
```
Open VS Code, create a file, enter the following content

```python
import matplotlib.pyplot as plt # <-- put a breakpoint here
raise ValueError("I should get here")
```

Put the breakpoint, start debugging, step over.

## Additional information
It seems removing either of matplotlib, jupyter or PyQt6 solves the problem.

Installing an older version of the Python extension (ms-python.python) also works. v2022.4.1 was the first that caused the issue, while the previous (previous according to VS Code) v2022.2.1924087327 still works.

I could capture the behavior in the following video: 
[screencast.webm](https://user-images.githubusercontent.com/6746316/199797343-5a2146cf-4d97-4fc1-a4d6-e0007c115621.webm)

wmvanvliet:
This adds support for interactive matplotlib figures using the qt6 backend. The code could really use some de-duplication (it defines the list of possible qt backends more than 10 times all over the place) and the docstrings need some love too. But it seems to work.

fixes #1112

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydev_ipython/inputhook.py
+++ b/src/debugpy/_vendored/pydevd/pydev_ipython/inputhook.py
@@ -26,6 +26,7 @@ GUI_WX = 'wx'
 GUI_QT = 'qt'
 GUI_QT4 = 'qt4'
 GUI_QT5 = 'qt5'
+GUI_QT6 = 'qt6'
 GUI_GTK = 'gtk'
 GUI_TK = 'tk'
 GUI_OSX = 'osx'
@@ -173,8 +174,10 @@ class InputHookManager(object):
         self.clear_inputhook()
 
     def enable_qt(self, app=None):
-        from pydev_ipython.qt_for_kernel import QT_API, QT_API_PYQT5
-        if QT_API == QT_API_PYQT5:
+        from pydev_ipython.qt_for_kernel import QT_API, QT_API_PYQT5, QT_API_PYQT6
+        if QT_API == QT_API_PYQT6:
+            self.enable_qt6(app)
+        elif QT_API == QT_API_PYQT5:
             self.enable_qt5(app)
         else:
             self.enable_qt4(app)
@@ -232,6 +235,21 @@ class InputHookManager(object):
     def disable_qt5(self):
         if GUI_QT5 in self._apps:
             self._apps[GUI_QT5]._in_event_loop = False
+        self.clear_inputhook()
+
+    def enable_qt6(self, app=None):
+        from pydev_ipython.inputhookqt6 import create_inputhook_qt6
+        app, inputhook_qt6 = create_inputhook_qt6(self, app)
+        self.set_inputhook(inputhook_qt6)
+
+        self._current_gui = GUI_QT6
+        app._in_event_loop = True
+        self._apps[GUI_QT6] = app
+        return app
+
+    def disable_qt6(self):
+        if GUI_QT6 in self._apps:
+            self._apps[GUI_QT6]._in_event_loop = False
         self.clear_inputhook()
 
     def enable_gtk(self, app=None):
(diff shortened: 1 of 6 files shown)


wmvanvliet:

SebusFu:
Hi, is there any update here? I can't get debugging to work on macOS using PySide6 and Matplotlib, with all mentioned tricks. Or is there any other hint I am overseeing?

rchiodo (approved):

rchiodo:
@wmvanvliet can you close and reopen? Not sure why the lint portion is waiting but I don't think it's going to run since it's been waiting for 4 months.

wmvanvliet:
The linting is not running because it requires approval from a maintainer.

AdamYoblick:
It's running now :) We have to manually run it with `/azp run` when it comes from a fork, for security reasons.

rchiodo:
@AdamYoblick does this test failure look like the flakey test you added retry to? Maybe we need to merge this branch with main?

AdamYoblick:
> @AdamYoblick does this test failure look like the flakey test you added retry to? Maybe we need to merge this branch with main?

Yes, it does look similar. Test workers crashing. Although I don't think merging main in will make a difference in this case because the test I marked as flaky was not the test method that failed here (test_log_dir_env). Unfortunately, we probably need to mark more tests as flaky. I don't remember 100% but I *think* that it was only python 3.12 on windows crashing before, just like in this case.

I wonder if we just need to mark the entire test suite as flaky with retries until this is properly investigated.

int19h:
This one is a pytest crash reading from a socket, so likely to be environment-induced.

In general I recall two main categories of flaky tests. One was anything to do with attach-to-PID, and specific to tests that do that. The other is a general failure when sockets just get randomly closed or timed out, which I have only ever seen in CI and not locally, but can happen with almost every test (since anything that runs debugpy involves sockets); I always suspected it is also something coming from the environment for those runs but I never saw any clues as to what it could be.

SebusFu:
Thank you all for taking care, may I please ask if approved now (If I understand correctly?), when it can be released? Or how can I help to test e.g. in my environment?

AdamYoblick:
@sebusfu - You don't have to wait for a debugpy release to be able to test the changes. Here's what you do:

1. Clone or fork the repo (which you already have)
2. Add the following line to your launch configuration in launch.json:

`"debugAdapterPath": "<path_to_your_fork>/src/debugpy/adapter"` 

3. Launch the file

This will use your local debugpy source instead of the debugpy that comes installed with the Python Core Extension, and you can see if it works.

Thanks!

SebusFu:
Thank you very much, works like a charm now for me!

AdamYoblick:
This test fail has nothing to do with your changes, it's a flaky test, so I'm going to merge this in.

Issue #1573, opened 2024-04-29
[sys.monitoring] Jump to Cursor / Go To Statement


int19h:
Implement the "goto" DAP request.

Constrained to currently executing function on the specified thread.

int19h:
Setting `f_lineno` on the frame object should adjust the instruction pointer accordingly. Python itself will check that the new location is valid (i.e. is in the same code object as current IP).

PR #1575, opened 2024-05-01
Add support for the goto request

rchiodo:
Fixes https://github.com/microsoft/debugpy/issues/1573

Implement support for the goto request. Note that goto is only supported in the same code block (function usually).

### DIFF
--- a/src/debugpy/common/util.py
+++ b/src/debugpy/common/util.py
@@ -2,9 +2,12 @@
 # Licensed under the MIT License. See LICENSE in the project root
 # for license information.
 
+from functools import partial
 import inspect
+import itertools
 import os
 import sys
+from types import CodeType
 
 
 def evaluate(code, path=__file__, mode="eval"):
@@ -163,3 +166,29 @@ def hide_thread_from_debugger(thread):
         thread.is_debugpy_thread = True
         thread.pydev_do_not_trace = True
         thread.is_pydev_daemon_thread = True
+
+class IDMap(object):
+
+    def __init__(self):
+        self._value_to_key = {}
+        self._key_to_value = {}
+        self._next_id = partial(next, itertools.count(0))
+
+    def obtain_value(self, key):
+        return self._key_to_value[key]
+
+    def obtain_key(self, value):
+        try:
+            key = self._value_to_key[value]
+        except KeyError:
+            key = self._next_id()
+            self._key_to_value[key] = value
+            self._value_to_key[value] = key
+        return key
+
+def contains_line(code: CodeType, line: int) -> bool:
+    for co_line in code.co_lines():
+        if len(co_line) == 3:
+            if co_line[2] == line:
+                return True
+    return False
--- a/src/debugpy/common/util.py
+++ b/src/debugpy/common/util.py
@@ -2,12 +2,12 @@
 # Licensed under the MIT License. See LICENSE in the project root
 # for license information.
 
+from functools import partial
 import inspect
+import itertools
 import os
 import sys
 from types import CodeType
-
-from debugpy.server import new_dap_id
 
 
 def evaluate(code, path=__file__, mode="eval"):
@@ -170,10 +170,10 @@ def hide_thread_from_debugger(thread):
 
 class IDMap(object):
 
-    def __init__(self):
+    def __init__(self, id_generator=partial(next, itertools.count(0))):
         self._value_to_key = {}
         self._key_to_value = {}
-        self._next_id = new_dap_id
+        self._next_id = id_generator
 
     def obtain_value(self, key):
         return self._key_to_value[key]
--- a/src/debugpy/common/util.py
+++ b/src/debugpy/common/util.py
@@ -2,12 +2,12 @@
 # Licensed under the MIT License. See LICENSE in the project root
 # for license information.
 
-from functools import partial
 import inspect
-import itertools
 import os
 import sys
 from types import CodeType
+
+from debugpy.server import new_dap_id
 
 
 def evaluate(code, path=__file__, mode="eval"):
@@ -173,7 +173,7 @@ class IDMap(object):
     def __init__(self):
         self._value_to_key = {}
         self._key_to_value = {}
-        self._next_id = partial(next, itertools.count(0))
+        self._next_id = new_dap_id
 
     def obtain_value(self, key):
         return self._key_to_value[key]
--- a/src/debugpy/common/util.py
+++ b/src/debugpy/common/util.py
@@ -167,6 +167,7 @@ def hide_thread_from_debugger(thread):
         thread.pydev_do_not_trace = True
         thread.is_pydev_daemon_thread = True
 
+
 class IDMap(object):
 
     def __init__(self):
@@ -186,9 +187,8 @@ class IDMap(object):
             self._value_to_key[value] = key
         return key
 
+
 def contains_line(code: CodeType, line: int) -> bool:
-    for co_line in code.co_lines():
-        if len(co_line) == 3:
-            if co_line[2] == line:
-                return True
-    return False
+    return any(
+        (len(co_line) == 3 and co_line[2] == line for co_line in code.co_lines())
+    )
--- a/src/debugpy/server/adapters.py
+++ b/src/debugpy/server/adapters.py
@@ -354,7 +354,7 @@ class Adapter:
         path = request("source", dict)["path"]
         line = request("line", int)
         target_id = self._goto_targets_map.obtain_key((path, line))
-        target = {"id": target_id, "label": "%s:%s" % (path, line), "line": line}
+        target = {"id": target_id, "label": f"({path}:{line})", "line": line}
 
         return {"targets": [target]}
     
--- a/src/debugpy/server/adapters.py
+++ b/src/debugpy/server/adapters.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License. See LICENSE in the project root
 # for license information.
 
+import inspect
 import os
 import sys
 import threading
@@ -11,6 +12,7 @@ from debugpy import adapter
 from debugpy.adapter import components
 from debugpy.common import json, log, messaging, sockets
 from debugpy.common.messaging import MessageDict, Request
+from debugpy.common.util import IDMap, contains_line
 from debugpy.server import eval, new_dap_id
 from debugpy.server.tracing import (
     Breakpoint,
@@ -60,6 +62,7 @@ class Adapter:
     _capabilities: Capabilities = None
     _expectations: Expectations = None
     _start_request: messaging.Request = None
+    _goto_targets_map: IDMap = IDMap()
 
     def __init__(self, stream: messaging.JsonIOStream):
         self._is_initialized = False
@@ -344,6 +347,40 @@ class Adapter:
         if thread is None:
             raise request.isnt_valid(f'Unknown thread with "threadId":{thread_id}')
         self._tracer.step_over(thread)
+        return {}
+
+    def gotoTargets_request(self, request: Request) -> dict:
+        path = request("source", dict)["path"]
+        line = request("line", int)
+        target_id = self._goto_targets_map.obtain_key((path, line))
+        target = {"id": target_id, "label": "%s:%s" % (path, line), "line": line}
+
+        return {"targets": [target]}
+    
+
+    def goto_request(self, request: Request) -> dict:
+        thread_id = request("threadId", int)
+        thread = Thread.get(thread_id)
+        if thread is None:
+            return request.isnt_valid(f'Unknown thread with "threadId":{thread_id}')
+        target_id = request("targetId", int)
+        if target_id is None:
+            return request.isnt_valid('Unknown targetId for goto')
+        try:
+            path, line = self._goto_targets_map.obtain_value(target_id)
+        except KeyError:
+            return request.isnt_valid('Invalid targetId for goto')
+
+        # Make sure the thread is in the same source file
+        current_source = inspect.getsourcefile(thread.current_frame.f_code)
+        if current_source.casefold() != path.casefold():
+            return request.cant_handle(f'{path} is not in the same code block as the current frame', silent=True)
+        
+        # Make sure line number is in the same code black
+        if not contains_line(thread.current_frame.f_code, line):
+            return request.cant_handle(f'Line {line} is not in the same code block as the current frame', silent=True)
+
+        self._tracer.goto(thread, path, line)
         return {}
 
     def exceptionInfo_request(self, request: Request):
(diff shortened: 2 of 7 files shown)


tests/debugpy/test_step.py
```diff
@@ -56,7 +56,6 @@ def func():
         inner2_target = targets[0]["id"]
 
         session.request("goto", {"threadId": stop.thread_id, "targetId": inner2_target})
-        session.wait_for_next_event("continued")
```
rchiodo, line 59: Not sure why pydevd does a continue, but the spec says once a goto request happens, just a stop event should fire:

https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Goto

tests/patterns/dap.py
```diff
@@ -11,7 +11,7 @@
 from tests.patterns import some, _impl
 
 
-id = some.int.in_range(0, 10000)
+id = some.int.in_range(-10000, 10000)
```
rchiodo, line 14: ids seem to be negative in the new code

src/debugpy/common/util.py
```diff
@@ -163,3 +166,29 @@ def hide_thread_from_debugger(thread):
...
+            self._key_to_value[key] = value
+            self._value_to_key[value] = key
+        return key
+
+def contains_line(code: CodeType, line: int) -> bool:
+    for co_line in code.co_lines():
+        if len(co_line) == 3:
+            if co_line[2] == line:
+                return True
+    return False
```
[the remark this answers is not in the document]
rchiodo: Thanks, I've never seen that before.

src/debugpy/server/adapters.py
```diff
@@ -60,6 +62,7 @@ class Expectations(components.Capabilities):
     _capabilities: Capabilities = None
     _expectations: Expectations = None
     _start_request: messaging.Request = None
+    _goto_targets_map: IDMap = IDMap()
```
int19h, line 65: I would recommend using `new_dap_id()` for this. The main reason why I made that a single global counter is because that makes it so much easier to grep things in the message log when IDs are not reused.

rchiodo: You mean inside of the IDMap.

rchiodo: I used `new_dap_id` for the `next_id` function in the IDMap

src/debugpy/server/adapters.py
```diff
@@ -346,6 +350,40 @@ def next_request(self, request: Request):
         self._tracer.step_over(thread)
         return {}
 
+    def gotoTargets_request(self, request: Request) -> dict:
+        path = request("source", dict)["path"]
+        line = request("line", int)
+        target_id = self._goto_targets_map.obtain_key((path, line))
+        target = {"id": target_id, "label": "%s:%s" % (path, line), "line": line}
```
int19h, line 357: `%` is sort of legacy in modern Python. For cases where localization is not necessary, f-strings should be preferred: `f"{path}:{line}"`.

For display strings where placeholder order can vary when localized, the ideal form is `str.format()` with kwargs, i.e.: `"{path}:{line}".format(path=..., line=...)`. Existing code doesn't follow this, though, and mostly uses f-strings even then.

rchiodo: Ah I just copied this from pydevd, I'll switch it to use a format string.

src/debugpy/server/adapters.py
```diff
@@ -346,6 +350,40 @@ def next_request(self, request: Request):
         self._tracer.step_over(thread)
         return {}
 
+    def gotoTargets_request(self, request: Request) -> dict:
+        path = request("source", dict)["path"]
```
int19h, line 354: It should also check that "path" is a string. This will do all the necessary validation:
```py
source = request("source", json.object())  # if missing you get {}
path = source("path", str)
```

Also, this is not the only place that will have to deal with "source", since it's a recurring thing in DAP, both incoming and outgoing (e.g. take a look at `setBreakpoints_request`). For outgoing, there's class `Source` in `debug.server.tracing`; perhaps it should get a `from_json` class method?

Long-term I think it's best to drop this approach to validating JSON at point of use entirely, and instead parse and validate messages before they get dispatched to handlers. The repo with the DAP spec also has a complete JSON schema for it, from which strongly typed classes could be generated (this is what pydevd does).

src/debugpy/server/adapters.py
```diff
@@ -346,6 +350,40 @@ def next_request(self, request: Request):
...
+        return {"targets": [target]}
+    
+
+    def goto_request(self, request: Request) -> dict:
+        thread_id = request("threadId", int)
+        thread = Thread.get(thread_id)
+        if thread is None:
+            return request.isnt_valid(f'Unknown thread with "threadId":{thread_id}')
+        target_id = request("targetId", int)
+        if target_id is None:
```
int19h, line 368: No need for this check, since `request(..., int)` will throw the appropriate `InvalidMessageError` if the property is missing or has invalid type.

(If it really should be optional, you can do `request(..., int, optional=True)` to get an empty tuple in that case.)

src/debugpy/server/adapters.py
```diff
@@ -346,6 +350,40 @@ def next_request(self, request: Request):
...
+            return request.isnt_valid('Unknown targetId for goto')
+        try:
+            path, line = self._goto_targets_map.obtain_value(target_id)
+        except KeyError:
+            return request.isnt_valid('Invalid targetId for goto')
+
+        # Make sure the thread is in the same source file
+        current_source = inspect.getsourcefile(thread.current_frame.f_code)
+        if current_source.casefold() != path.casefold():
```
int19h, line 377: `casefold()` won't do the right thing for case-sensitive filesystems on Unix.

For breakpoints, I've used `pathlib.Path.resolve()` to normalize paths while also resolving symlinks etc - this is done in class `Source`, and should probably be reused here. In general I think since DAP always uses "source" whenever paths appear, it's best to handle paths throughout the debugger by converting them to `Source` as soon as possible, and then working with those objects - that way it will all be consistent.

src/debugpy/server/tracing/__init__.py
```diff
@@ -255,15 +261,7 @@ def stack_trace_len(self) -> int:
...
-        try:
-            with _cvar:
-                python_frame = self.current_frame
-        except ValueError:
-            raise ValueError(f"Can't get frames for inactive Thread({self.id})")
-        try:
-            return len(tuple(traceback.walk_stack(python_frame)))
-        finally:
-            del python_frame
+        return len(list(self.stack_trace()))
```
int19h, line 264: This is not ideal because `self.stack_trace` creates a `StackTrace` object for each frame on the stack, no matter how large, which is costly (since it needs to do `StackFrame.from_python_frame` for each to look up the existing one or generate ID for the new one).

That said, these two really should produce the same result, which the old code did not (because count included internal frames but actually listing them didn't). I think it might be best to refactor `stack_trace` into a new generator method that takes care of all the locking, iteration, logging, internal frame exclusion etc, but yields raw frame objects. Then both `stack_frame` and `stack_frame_len` can be implemented efficiently in terms of that.

src/debugpy/server/tracing/__init__.py
```diff
@@ -375,42 +373,61 @@ def invalidate(self, thread: Thread):
...
 
 
-@dataclass
 class Step:
-    step: Literal["in", "out", "over"]
-    origin: FrameType = None
-    origin_line: int = None
+    def __init__(
+        self,
+        step: Literal["in", "out", "over", "goto"],
```
int19h, line 379: "goto" is not needed anymore

int19h: Also I think this can be made simpler by making it a class variable and just setting it in derived classes, i.e.:
```py
class Step:
   kind: str
   ...

class StepIn(Step):
   kind = "in"
   ...
```

src/debugpy/server/adapters.py
```diff
@@ -346,6 +350,40 @@ def next_request(self, request: Request):
...
+            return request.cant_handle(f'{path} is not in the same code block as the current frame', silent=True)
+        
+        # Make sure line number is in the same code black
+        if not contains_line(thread.current_frame.f_code, line):
+            return request.cant_handle(f'Line {line} is not in the same code block as the current frame', silent=True)
+
```
int19h, line 383: For the checks above, I wonder if it would be better to let Python to take care of all that. Checking against the source files can be off because user can edit the file during debugging, but the code object will still contain bytecode and line number mappings for the original, so the check will be off. OTOH Python itself will check against the code object's line number map so it'll always do the right thing.

It looks like `Python` will throw `ValueError` for invalid line numbers when assigning to `f_lineno`, and the error message even makes sense ("line 1000 comes after the current code block") so you could just propagate that.

src/debugpy/server/tracing/__init__.py
```diff
@@ -375,42 +373,61 @@ def invalidate(self, thread: Thread):
...
-        else:
-            raise ValueError(f"Unknown step type: {self.step}")
+        raise ValueError(f"Unknown step type: {self.step}")
+
+
+class StepIn(Step):
+    def __init__(self):
+        super().__init__("in")
+
+    @override
```
int19h, line 398: The other step classes should also get `@override`.

tests/debugpy/test_step.py
int19h: It's possible that pydevd implemented it like that when the spec didn't have that bit; it's pretty common for them to add clarifications like that later on if someone comes and asks how exactly it should behave.

tests/patterns/dap.py
int19h: Yep, that's intentional, to distinguish synthetic thread IDs from OS-provided ones. I also found that using negative IDs makes it easier to distinguish from other numeric values in JSON logs, which tend to be positive (counts, user data etc). Thank you for catching this one!

src/debugpy/common/util.py
```diff
@@ -163,3 +166,29 @@ def hide_thread_from_debugger(thread):
...
         thread.pydev_do_not_trace = True
         thread.is_pydev_daemon_thread = True
+
+
+class IDMap(object):
+
+    def __init__(self):
+        self._value_to_key = {}
+        self._key_to_value = {}
+        self._next_id = new_dap_id
```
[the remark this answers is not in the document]
int19h: Kind of, but now we have `debugpy.common` depending on `debugpy.server`, which shouldn't be the case (it'll make it impossible to split `debugpy.server` into its own package later). I think it's better to make this object parametrized with the ID generator lambda, like `defaultdict` does.

src/debugpy/server/adapters.py
rchiodo: You mean skip these checks entirely? I'll try that and see what happens.

rchiodo: Unfortunately it doesn't work because it's in the callback. It causes the debugger to get screwed up.

src/debugpy/server/tracing/__init__.py
rchiodo: I thought that's what this was doing but I found something faster than just taking the length of a list. At least according to stack overflow:

```python
sum(1 for dummy in self._generate_stack_trace())
```

src/debugpy/common/util.py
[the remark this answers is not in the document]
rchiodo: Fixed it by passing in a func to generate the ids

src/debugpy/server/tracing/__init__.py
rchiodo: I think? I implemented what you suggested. I'm not 100% sure it's much different than what I already had

src/debugpy/server/adapters.py
int19h: I think this might be one of those cases where it would be best for the request handler to not respond right away, and instead for the code that actually updates the line number to do so. You can return `messaging.NO_RESPONSE` here to indicate this, and then when setting `f_lineno`, catch `ValueError`, convert it to `MessageHandlingError`, and do `request.respond(err)` from there.

src/debugpy/server/tracing/__init__.py
int19h: Python is counter-intuitive here - `sum` over a generator is often slower than `len()` of a list, because the latter is entirely native code while the former has to drop into Python for every item it iterates over. This is often the pattern in general - code that is more profligate with memory use but does it all in native is often faster than thrifty code that does the same in Python, just because Python bytecode is so slow.

(In this case I don't think `sum` vs `len` will make any difference in practice because it will be dominated by the code that actually generates the item either way.)

int19h: This code is currently wrapping every Python frame object into a `StackFrame` instance using `StackFrame.from_python_frame`. That's the part I'd rather avoid, because it is going to try to find an existing `StackFrame` first, then generate DAP IDs for all those frames, and at the end of it they just get thrown away after computing the length of the list.

My suggestion is to factor out the part that calls `from_python_frame()`, so that they are only created for `stack_frame()` but not for `stack_frame_len()`. The actual stack walking would then happen in a new private method shared by the two.

rchiodo: I mean this could be made faster if the stack was cached, but I didn't think you were suggesting that? We'd have to track if the ip changed or not in order to invalidate the cache.

src/debugpy/server/adapters.py
rchiodo: Implemented your suggestion using callbacks.

src/debugpy/server/tracing/__init__.py
rchiodo: Changed this to use a cached value

src/debugpy/server/tracing/__init__.py
```diff
@@ -255,21 +272,24 @@ def stack_trace_len(self) -> int:
...
         """
         Returns an iterable of StackFrame objects for the current stack of this thread,
         starting with the topmost frame.
         """
+        return self._get_stack_trace()
+    
+    def _get_stack_trace(self) -> list["StackFrame"]:
+        # If our current frame is none, this is invalid. Throw an error.
+        if self._current_frame is None:
+            raise InvalidMessageError(reason="Tread is not suspended")
```
int19h, line 287: This will crash the handler since `InvalidMessageError` will not be associated with the request, and so it won't be treated as a response to it automatically. It should use some non-messaging-specific exception type - I think `ValueError` might be appropriate here - and then the handler should catch it and do `return request.isnt_valid(...)`. Or the handler can check `current_frame` itself first, before it even tries to call `stack_trace` and `stack_trace_len`.

BTW, I don't recall if `stack_trace` and `stack_trace_len` were called anywhere other than that one place in `stackTrace_request`. If they aren't, given that they are one-liners now, might as well inline them there.

rchiodo: Okay I'll move the error check to the stackTrace_request. You were right, there's only one spot where this was called

int19h (approved):

PR #1580, opened 2024-05-07
Allow python 3.12 in attach code

AdamYoblick:
Creating this PR so the dll's get generated through github actions. Once they are available as artifacts, I will download them and test locally.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/py_version.hpp
+++ b/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/py_version.hpp
@@ -21,7 +21,8 @@ enum PythonVersion {
     PythonVersion_38 = 0x0308,
     PythonVersion_39 = 0x0309,
     PythonVersion_310 = 0x030A,
-    PythonVersion_311 = 0x030B
+    PythonVersion_311 = 0x030B,
+    PythonVersion_312 = 0x030C
 };
 
 
@@ -69,6 +70,9 @@ static PythonVersion GetPythonVersion(void *module) {
                     }
                     if(version[3] == '1'){
                         return PythonVersion_311;
+                    }
+                    if(version[3] == '2'){
+                        return PythonVersion_312;
                     }
                 }
                 return PythonVersion_Unknown; // we don't care about 3.1 anymore...


rchiodo (approved):

AdamYoblick:
Downloading windows artifacts from https://github.com/microsoft/debugpy/actions/runs/8993404815?pr=1580 and testing attach scenario locally.

AdamYoblick:
The attach DOES work now, we got past the "unknown python version" errors. But breakpoints are not getting hit like they were in when using python 3.11. 

@int19h - I know debugpy works with python 3.12, but is there anything special I need to do to the attach code for breakpoints to be hit when attaching? This scenario works for 3.11 but not 3.12. Thanks!

int19h:
The other place that needs to be updated is here. Without this DLL, it will not be able to set tracing on existing threads.

https://github.com/microsoft/debugpy/blob/09b5af0bb730e2707ec3bbc14c509523eca116c0/src/debugpy/_vendored/pydevd/pydevd_tracing.py#L252-L256

int19h:
But then the DLL itself needs to implement this function for Python 3.12:

https://github.com/microsoft/debugpy/blob/09b5af0bb730e2707ec3bbc14c509523eca116c0/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/py_custom_pyeval_settrace.hpp#L170-L190

https://github.com/microsoft/debugpy/blob/09b5af0bb730e2707ec3bbc14c509523eca116c0/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/py_custom_pyeval_settrace.hpp#L170-L190

int19h:
That is the trickiest part. This is the implementation for Python 3.11, and I believe much of this has changed for 3.12 (e.g. `use_tracing` doesn't even exist anymore):

https://github.com/microsoft/debugpy/blob/09b5af0bb730e2707ec3bbc14c509523eca116c0/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/common/py_custom_pyeval_settrace_311.hpp

sylvainmouquet:
we can also add python 3.13?

PR #1608, opened 2024-06-24
update macos image. macos-11 is no longer supported.

AdamYoblick:


karthiknadig (approved):

AdamYoblick:
Ugh. Looks like moving up to macos-latest image broke something in the numpy tests. Need to troubleshoot.

AdamYoblick:
I just ran the debugpy-test pipeline from main to compare the results, and there are several test failures there too: https://dev.azure.com/debugpy/debugpy/_build/results?buildId=4174&view=ms.vss-test-web.build-test-results-tab

I will compare the failures but I don't think moving up to macos-latest pool is the cause of any of these. If the failures are the same, I'm going to merge this in, since staying on macos-11 will definitely cause failures very soon.

AdamYoblick:
Yeah I'm going to merge this in. The same tests are failing when I run fresh from main. My changes has nothing to do with the breaks. I will investigate those test failures separately.

PR #1617, opened 2024-07-02
Add some extra information about logging

rchiodo:


AdamYoblick:
@rchiodo The checks are going to fail because of numpy issues. Probably the same ones you're seeing in pylance. Just ignore, I need to fix them 😄

AdamYoblick (approved):

PR #1619, opened 2024-07-03
Add some more typing for the public api

rchiodo:
Was just using Pylance to test debugpy and started added some typing to the public API

### DIFF
--- a/src/debugpy/server/cli.py
+++ b/src/debugpy/server/cli.py
@@ -7,6 +7,7 @@ import os
 import re
 import sys
 from importlib.util import find_spec
+from typing import Any, Optional
 
 # debugpy.__main__ should have preloaded pydevd properly before importing this module.
 # Otherwise, some stdlib modules above might have had imported threading before pydevd
@@ -18,6 +19,7 @@ import pydevd
 from _pydevd_bundle import pydevd_runpy as runpy
 
 import debugpy
+import debugpy.server
 from debugpy.common import log
 from debugpy.server import api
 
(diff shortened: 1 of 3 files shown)


src/debugpy/server/cli.py
```diff
@@ -41,13 +43,14 @@
 
 class Options(object):
     mode = None
-    address = None
+    address: "Optional[tuple[str, int]]" = None
```
debonte, line 46: I think the general feeling in the typing community is that `Foo | None` is more readable than `Optional[Foo]`.

https://typing.readthedocs.io/en/latest/source/modernizing.html#typing-union-and-typing-optional

rchiodo: I'll change them to that then.

src/debugpy/server/cli.py
```diff
@@ -41,13 +43,14 @@
 
 class Options(object):
     mode = None
-    address = None
+    address: "tuple[str, int] | None" = None
```
debonte, line 46: I think the quotes are unnecessary?
```suggestion
    address: tuple[str, int] | None = None
```

debonte (approved):

src/debugpy/server/cli.py
rchiodo: They're only unnecessary on newer Python versions. Debugpy has to run on 3.8 I believe.

image

debonte: Oh, interesting, I haven't run into that before.

rchiodo: Yeah debugpy's pyproject forces 3.8:

```
[tool.pyright]
pythonVersion = "3.8"
```

PR #1624, opened 2024-07-15
Add parsing args from environment

### LINKED ISSUE #1612: Terminal debug experience in Python Debbuger Extension
We are implemented a debug terminal with shell integration, in order to do that we will need an extra implementation in debugpy side.

In order to start debugging from the terminal, we need to specify a connect or listen port in the command: `python -m debugpy --connect 5678 file.py`.

What we want is to be able to use only: `python -m debugpy file.py` and debugpy set internally the connect port by default to 5678.

AdamYoblick:
Fixes #1612

- Added reading args from the environment. Command line args always take precedence.
- Added unit tests for environment args

Testing was done in several ways:

1. I added multiple unit tests for the new code that reads from the environment, as well as made sure that existing tests were passing.
2. I debugged a python file locally, pointing vscode at my local debugpy source, using the following launch.json config:

```json
{
      "name": "Python Debugger: Current File WITH ARGS",
      "type": "debugpy",
      "request": "launch",
      "program": "${file}",
      "debugAdapterPath": "E:/repos/github/debugpy/src/debugpy/adapter",
      "args" : ["--msg", "Hello World"]
    },
```
I verified that these extra args were passed to the file being run and that breakpoints could be hit. I also verified that this worked with no args.

3. I opened a command prompt and starting debugpy in listen mode:

`python E:\repos\github\debugpy\src\debugpy --listen 5678 --wait-for-client "C:\Users\advolker\OneDrive - Microsoft\Desktop\pythonTest\infiniteLoop.py"`

Then I attached through vscode using the following launch.json config:

```json
    {
      "name": "Attach",
      "type": "python",
      "request": "attach",
      "connect": {
        "port": 5678,
      }  
```

Breakpoints worked just fine.

4. I did the same test as above, but I set the options `--listen 5678` and `--wait-for-client` in the environment before running debugpy, like this:

`set DEBUGPY_EXTRA_ARGV=--listen 5678 --wait-for-client`

Then I started debugpy from the same command prompt with only the target specified. The other options were read from the environment:

`python E:\repos\github\debugpy\src\debugpy "C:\Users\advolker\OneDrive - Microsoft\Desktop\pythonTest\infiniteLoop.py"`

Then I attached through vscode in the same way as the previous step. Breakpoints worked just fine.

### DIFF
--- a/src/debugpy/server/cli.py
+++ b/src/debugpy/server/cli.py
@@ -7,10 +7,7 @@ import os
 import re
 import sys
 from importlib.util import find_spec
-from typing import Any
-from typing import Union
-from typing import Tuple
-from typing import Dict
+from typing import Any, Union, Tuple, Dict
 
 # debugpy.__main__ should have preloaded pydevd properly before importing this module.
 # Otherwise, some stdlib modules above might have had imported threading before pydevd
@@ -193,48 +190,22 @@ switches = [
 ]
 # fmt: on
 
-
-def consume_argv():
-    while len(sys.argv) >= 2:
-        value = sys.argv[1]
-        del sys.argv[1]
+# Consume all the args from a given list
+def consume_args(args: list):
+    while len(args) >= 2:
+        value = args[1]
+        del args[1]
         yield value
 
+# Parse the args from the command line, then from the environment.
+# Args from the environment are only used if they are not already set from the command line.
+def parse_args():
 
-def parse_argv():
+    # keep track of the switches we've seen so far
     seen = set()
-    it = consume_argv()
 
-    while True:
-        try:
-            arg = next(it)
-        except StopIteration:
-            raise ValueError("missing target: " + TARGET)
-
-        switch = arg
-        if not switch.startswith("-"):
-            switch = ""
-        for pattern, placeholder, action in switches:
-            if re.match("^(" + pattern + ")$", switch):
-                break
-        else:
-            raise ValueError("unrecognized switch " + switch)
-
-        if switch in seen:
-            raise ValueError("duplicate switch " + switch)
-        else:
-            seen.add(switch)
-
-        try:
-            action(arg, it)
-        except StopIteration:
-            assert placeholder is not None
-            raise ValueError("{0}: missing {1}".format(switch, placeholder))
-        except Exception as exc:
-            raise ValueError("invalid {0} {1}: {2}".format(switch, placeholder, exc))
-
-        if options.target is not None:
-            break
+    parse_args_from_command_line(seen)
+    parse_args_from_environment(seen)
 
     if options.mode is None:
         raise ValueError("either --listen or --connect is required")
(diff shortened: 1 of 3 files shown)


rchiodo (commented):
Maybe we should have some unit tests for this? The logic seems complex enough to warrant it?

AdamYoblick:
Definitely yes, I'm adding tests as we speak :)

AdamYoblick:
@rchiodo I've added unit tests for consuming switches from the environment. They are all green locally.

tests/debugpy/server/test_cli.py
```diff
@@ -62,16 +72,15 @@ def parse(args):
     return parse
 
 
+# Test a combination of command line switches
 @pytest.mark.parametrize("target_kind", ["file", "module", "code"])
 @pytest.mark.parametrize("mode", ["listen", "connect"])
 @pytest.mark.parametrize("address", ["8888", "localhost:8888"])
 @pytest.mark.parametrize("wait_for_client", ["", "wait_for_client"])
-@pytest.mark.parametrize("script_args", ["", "script_args"])
```
rchiodo, line 69: Is this test covered by the new unit tests?

rchiodo (approved):

tests/debugpy/server/test_cli.py
AdamYoblick: No, and I think I need to change my logic anyway. I didn't add a test for this case, I will give it a shot

src/debugpy/server/cli.py
```diff
@@ -193,23 +191,71 @@ def do(arg, it):
...
         value = sys.argv[1]
         del sys.argv[1]
         yield value
 
+# Consume all the args from a given list
+def consume_args(args: list):
+    if (args is sys.argv):
+        yield from consume_argv()
+    
+    while len(args) >= 1:
```
rchiodo, line 206: Doesn't this go down this path after it yields the args from consume_argv()? Like isn't an else needed here?

tests/debugpy/server/test_cli.py
```diff
@@ -71,7 +82,7 @@ def test_targets(cli, target_kind, mode, address, wait_for_client, script_args):
     expected_options = {
         "mode": mode,
         "target_kind": target_kind,
-        "wait_for_client": bool(wait_for_client),
+        "wait_for_client": False
```
rchiodo, line 85: This seems off? The parameter is being skipped now?

src/debugpy/server/cli.py
AdamYoblick: Yeah I think you are correct here. Although all my testing showed this as working, but maybe I missed an edge case. I will add the else for sure, since it seems more correct.

tests/debugpy/server/test_cli.py
AdamYoblick: This was just a refactor with no functional difference. It defaults to false, so in the case of the empty string, the conditional on line 96 will eval to false. But in the case where it has a value, line 96 will eval to true, and the default value will be overridden on line 98.

src/debugpy/server/cli.py
AdamYoblick: This is fixed in the latest commit.

rchiodo (approved):

src/debugpy/server/cli.py
```diff
@@ -193,23 +191,71 @@ def do(arg, it):
...
+
+    argsList = args.split()
+
+    #print("args:", file=sys.stderr)
+    #print(*argsList, file=sys.stderr)
+
+    seenFromEnvironment = set()
+    parse_args_helper(argsList, seenFromCommandLine, seenFromEnvironment, True)
+
+def parse_args_helper(args: list, seenFromCommandLine: set, seenFromEnvironment: set = None, isFromEnvironment=False):
```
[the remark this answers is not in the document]
karthiknadig: If `seenFromEnvironment` is always passed-in then I think a simpler update would be to remove the default `None`.

karthiknadig (approved):

src/debugpy/server/cli.py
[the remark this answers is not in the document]
AdamYoblick: `seenFromEnvironment` isn't passed in when the command line is being parsed, only when the environment is being parsed. In fact, that variable isn't accessed at all when parsing the command line. But I'm fine with empty set instead of None, it doesn't get used anyway. 😄

PR #1637, opened 2024-07-24
Add more typing to debugpy and switch to 'standard' type checking mode

rchiodo:
This PR adds type annotations across `debugpy` and switches Pyright to `standard` type-checking mode (`typeCheckingMode = "standard"`), with `# type: ignore` comments disabled repository-wide (`enableTypeIgnoreComments = false`). It was originally used to exercise Pylance, but the stricter typing is a worthwhile improvement on its own.

The vast majority of the diff is pure type annotations. However, a handful of non-annotation changes were needed (or surfaced by the stricter checking) and are called out below so reviewers do not have to infer them from the diff.

### Non-annotation (behavioral) changes
- **`adapter/clients.py`** – `launch`/`attach` handling now preserves `request_options`; out-of-order `configurationDone` requests fail explicitly; and an `evaluate` request that cannot be propagated to the debug server now receives a failure response instead of hanging.
- **`adapter/launchers.py`** – for a debug launch, `spawn_debuggee` now asserts `servers.listener is not None` before reading the address/token, restoring fail-fast behavior (and narrowing the Optional) instead of silently skipping connection info.
- **`common/messaging.py`**
  - `Message.__call__` no longer uses `assert` for control flow: it explicitly raises when the payload is an `Exception`, so behavior is identical under `python -O`.
  - `Request.respond(None)` now models a successful empty response as an empty `MessageDict` (instead of `None`), matching how incoming empty responses are already parsed, so `Response.body` is always a `MessageDict` or an `Exception`.
  - `OutgoingRequest` disables the inherited `_parse`/`_handle` via explicit class-level attributes.
- **`common/singleton.py`** – `ThreadSafeSingleton.assert_locked` is no longer a `@staticmethod` and uses a side-effect-free `lock._is_owned()` check instead of `acquire(blocking=False)`/`release()`, avoiding an RLock recursion-count leak (and an unbalanced `release()` under `python -O`).
- **`common/json.py`** – `_converter` narrows accepted numeric types from `numbers.Number` to `int`/`float` only (DAP number payloads are int/float; `Decimal`/`complex`/`Fraction` never appear).
- **`common/util.py`** – `Observable.observers` keeps an immutable `()` class default (typed `Sequence`) and the `+=` call sites were switched to reassignment, avoiding mutation of a shared class attribute.
- **`server/api.py`** – an empty adapter-endpoints read now raises the intended `EOFError` instead of falling through to `json.loads` and surfacing `JSONDecodeError`.

Focused regression tests were added for the behavioral changes (see `tests/debugpy/common/test_messaging.py`, `tests/debugpy/common/test_singleton.py`, `tests/debugpy/common/test_json.py`, and `tests/debugpy/server/test_api.py`).

### Suppression policy
Because `enableTypeIgnoreComments = false`, `# type: ignore` has no effect and must not be used. Where a diagnostic must be suppressed (correct code the checker cannot verify, e.g. access to a private-but-stable runtime API), a narrow `# pyright: ignore[reportSpecificRule]` is used instead. This convention is now documented in [CONTRIBUTING.md](CONTRIBUTING.md#type-checking).


src/debugpy/adapter/clients.py
```diff
@@ -215,15 +219,16 @@ def handle(self, request):
             if self.session.no_debug:
                 servers.dont_wait_for_first_connection()
 
+            request_options = request("debugOptions", json.array(str))
             self.session.debug_options = debug_options = set(
-                request("debugOptions", json.array(str))
+
```
debonte, line 224: Should `request_options` be passed in here?

src/debugpy/adapter/clients.py
```diff
@@ -770,7 +777,7 @@ def notify_of_subprocess(self, conn):
 
 def serve(host, port):
     global listener
-    listener = sockets.serve("Client", Client, host, port)
+    listener = sockets.serve("Client", Client, host, port) # type: ignore
```
debonte, line 780: Maybe use `# pyright: ignore [specific diagnostic]`? Here and other places `type: ignore` is used.

src/debugpy/common/messaging.py
```diff
@@ -407,6 +426,13 @@ def wrap(self, key, *args, **kwargs):
     del _invalid_if_no_key
 
 
+class AssociatableMessageDict(MessageDict):
```
debonte, line 429: I believe the correct word (can be associated) is "associable".

src/debugpy/common/messaging.py
```diff
@@ -451,7 +472,7 @@ def __init__(self, channel, seq, json=None):
         """
 
     def __str__(self):
-        return json.repr(self.json) if self.json is not None else repr(self)
+        return json.repr(self.json).__str__() if self.json is not None else repr(self)
```
debonte, line 475: Shouldn't you call `str(json.repr(self.json))` rather than using `__str__`? https://docs.python.org/3.11/reference/datamodel.html#object.__str__

src/debugpy/adapter/clients.py
rchiodo: yeah this seems like a bug. The old code caused a type error I must have been in the middle of fixing it and didn't finish.

src/debugpy/common/messaging.py
rchiodo: Okay, will change.

src/debugpy/common/messaging.py
rchiodo: Oh that looks better. Yeah, I'll do that

src/debugpy/adapter/clients.py
rchiodo: Wasn't there a way to do this automatically in Pylance? I thought we had a setting for it

debonte: You need to have `enableTypeIgnoreComments` disabled.

debonte (dismissed):

heejaechang:

src/debugpy/common/messaging.py
```diff
@@ -464,14 +486,17 @@ def describe(self):
...
         """
         raise NotImplementedError
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args, **kwargs) -> MessageDict | Any | int | float:
         """Same as self.payload(...)."""
+        assert not isinstance(self.payload, Exception)
+        if args.count == 0 and kwargs == {}:
+            return self.payload
         return self.payload(*args, **kwargs)
```
heejaechang: The no-arg fast path is dead code. `args` is a tuple, so `args.count` is the bound `tuple.count` method; `args.count == 0` compares a method object to an int and is always `False`, so the intended "`msg()` with no args returns the payload" branch never fires and falls through to `self.payload()` → `MessageDict.__call__` with no key → `TypeError`. Use `if len(args) == 0 and not kwargs:`. Separately, `assert not isinstance(self.payload, Exception)` is stripped under `python -O`, so it can't be relied on for control flow.

src/debugpy/common/messaging.py
```diff
@@ -778,8 +808,10 @@ def wait_for_response(self, raise_if_failed=True):
             while self.response is None:
                 self.channel._handlers_enqueued.wait()
 
-        if raise_if_failed and not self.response.success:
+        if raise_if_failed and not self.response.success and isinstance( self.response.body, BaseException):
             raise self.response.body
+        
+        assert not isinstance(self.response.body, Exception)
```
heejaechang: `wait_for_response(raise_if_failed=False)` on a failed response now hits `assert not isinstance(self.response.body, Exception)` and raises `AssertionError`, breaking the documented contract that it returns the error body (failed responses carry Exception bodies such as `NoMoreMessages` on channel close). Under `python -O` the assert is stripped and it silently returns the Exception body instead — inconsistent behavior across optimization modes. Drop the assert or gate it on `raise_if_failed`.

src/debugpy/common/util.py
```diff
@@ -19,7 +19,7 @@ def evaluate(code, path=__file__, mode="eval"):
 class Observable(object):
     """An object with change notifications."""
 
-    observers = ()  # used when attributes are set before __init__ is invoked
+    observers = []  # used when attributes are set before __init__ is invoked
```
heejaechang: Changing the class default `observers = ()` to `[]` purely to satisfy typing reintroduces a mutable-shared-class-attribute footgun. The comment notes this default is used when attributes are set before `__init__` — exactly the window where a shared list would leak observers across instances if a subclass ever does `.append(...)`/`+=` before `super().__init__()`. The tuple failed loudly (`TypeError`); the list fails silently. Prefer keeping `()` and typing it (e.g. `observers: Sequence = ()` / `ClassVar`).

src/debugpy/common/json.py
```diff
@@ -93,10 +93,10 @@ def __format__(self, format_spec):
...
 
 
-def _converter(value, classinfo):
+def _converter(value: str, classinfo) -> Union[int, float, None]:
     """Convert value (str) to number, otherwise return None if is not possible"""
     for one_info in classinfo:
-        if issubclass(one_info, numbers.Number):
+        if issubclass(one_info, int) or issubclass(one_info, float):
             try:
                 return one_info(value)
```
heejaechang, line 104: Replacing `issubclass(one_info, numbers.Number)` with `issubclass(one_info, int) or issubclass(one_info, float)` silently drops `Decimal`, `complex`, and `Fraction` from accepted numeric conversions — a semantic contract change bundled into a typing PR. Likely fine for DAP payloads (int/float only), but please either keep `numbers.Number` or add a comment justifying the narrowing.

src/debugpy/server/api.py
```diff
@@ -37,32 +38,36 @@
...
+    def __new__(cls, *args, **kwargs):
+        log.debug("pydevd.settrace(*{0!r}, **{1!r})", args, kwargs)
+        # The stdin in notification is not acted upon in debugpy, so, disable it.
+        kwargs.setdefault("notify_stdin", False)
+        try:
+            return pydevd.settrace(*args, **kwargs)
+        except Exception:
+            raise
+        finally:
+            cls.called = True
```
heejaechang: In the `_settrace` function→class conversion, `cls.called = True` runs in a `finally`, so it latches even when `pydevd.settrace` raises. The new `configure()` guard then permanently rejects all future `configure()` calls with "debug adapter is already running" even though settrace never succeeded — set `called = True` only on success. Also, `except Exception: raise` is a no-op, and a class whose `__new__` returns a foreign object solely to hang a `called` flag is hard to follow; a module-level bool set on success would express the guard more clearly. This is a user-visible API behavior change and deserves its own justification and a test.

src/debugpy/adapter/servers.py
```diff
@@ -171,7 +174,7 @@ def authenticate(self):
         auth = self.channel.request(
             "pydevdAuthorize", {"debugServerAccessToken": access_token}
         )
-        if auth["clientAccessToken"] != adapter.access_token:
+        if not isinstance(auth, Exception) and auth["clientAccessToken"] != adapter.access_token:
```
heejaechang: `if not isinstance(auth, Exception) and auth["clientAccessToken"] != adapter.access_token:` is a fail-open pattern on a trust boundary. Today `channel.request(...)` guarantees a non-exception result so this is a no-op, but if that contract ever changes and `auth` becomes an exception, the mismatched-token check is silently skipped and the server is treated as authorized. Prefer failing closed (raise) when `auth` is unexpectedly an exception.

src/debugpy/common/messaging.py
```diff
@@ -1337,10 +1374,10 @@ def _parse_incoming_message(self):
         # for all JSON objects, and track them so that they can be later wired up to
         # the Message they belong to, once it is instantiated.
         def object_hook(d):
-            d = MessageDict(None, d)
+            d = AssociableMessageDict(None, d)
             if "seq" in d:
                 self._prettify(d)
-            d.associate_with = associate_with
+            setattr(d, "associate_with", associate_with)
             message_dicts.append(d)
```
heejaechang, line 1407: The `AssociableMessageDict` + `is_associable()` refactor is a good typed replacement for the `associate_with` monkeypatch, but it's applied inconsistently: `object_hook` still does `setattr(d, "associate_with", associate_with)` on an `AssociableMessageDict`, overwriting the new real method with a closure. Consider relying on the class method (or storing the message directly) so the migration is complete and future readers aren't confused about which path is authoritative.

pyproject.toml
```diff
@@ -21,6 +21,8 @@ ignore = ["src/debugpy/_vendored/pydevd", "src/debugpy/_version.py"]
 executionEnvironments = [
   { root = "src" }, { root = "." }
 ]
+typeCheckingMode = "standard"
+enableTypeIgnoreComments = false
```
heejaechang, line 25: This PR makes several runtime-behavior changes (`_settrace`/`configure` gate, `wait_for_response` assert, `configurationDone_request` raising, `Message.__call__` fast path) but adds no tests. Given the verified bugs, please add targeted unit tests around `wait_for_response(raise_if_failed=False)`, `Message.__call__()` no-arg behavior, and the `configure()` "already running" guard (including the settrace-failure case) before merging.

heejaechang:
Mostly a solid typing pass, but a few runtime-behavior changes rode in with it and two are genuine bugs (`Message.__call__` no-arg path and `wait_for_response(raise_if_failed=False)`). Please fix those, reconsider the mutable class-default and settrace-latch changes, and add targeted tests for the behavior changes.

src/debugpy/common/messaging.py
rchiodo: Fixed - the no-arg fast path now uses `if len(args) == 0 and not kwargs:`. Previously `args.count == 0` compared the bound `tuple.count` method to an int and was always False, so `msg()` never returned the payload. The payload `assert` is kept only as an invariant (not for control flow).

src/debugpy/common/messaging.py
rchiodo: Fixed - dropped the assert so `wait_for_response(raise_if_failed=False)` returns the error body per the documented contract, with consistent behavior under `python -O`. Return type is now `MessageDict | Exception`. Added a regression test.

src/debugpy/common/util.py
rchiodo: Fixed - reverted the class default to an immutable tuple `()` (typed `Sequence[Callable[..., Any]]`) and switched the two `+=` call sites to reassignment, so nothing mutates a shared class attribute.

src/debugpy/common/json.py
rchiodo: Added a comment justifying the narrowing: DAP number payloads are int/float only, so `Decimal`/`complex`/`Fraction` are intentionally not handled.

src/debugpy/server/api.py
rchiodo: Fixed - `called` now latches only after `pydevd.settrace` succeeds (moved out of the `finally`), so a failed settrace no longer permanently blocks `configure()`. Removed the no-op `except Exception: raise`. Added tests for both the success and settrace-failure cases.

src/debugpy/adapter/servers.py
rchiodo: Fixed - now fails closed: if the authorize result is an Exception we close the channel and raise, instead of silently skipping the token check.

src/debugpy/common/messaging.py
rchiodo: Fixed - completed the migration. `object_hook` no longer sets a per-dict `associate_with` closure; `AssociableMessageDict.associate_with` is now the single authoritative method, backed by a shared `associated_dicts` list, so associating the top-level dict still wires up every nested dict.

pyproject.toml
rchiodo: Added targeted unit tests: `wait_for_response(raise_if_failed=False)` and `Message.__call__()` no-arg (tests/debugpy/common/test_messaging.py), and the `configure()` already-running guard including the settrace-failure case (new tests/debugpy/server/test_api.py).

src/debugpy/adapter/launchers.py
```diff
@@ -88,12 +89,13 @@ def spawn_debuggee(
     env = {}
 
     arguments = dict(start_request.arguments)
-    if not session.no_debug:
+    if not session.no_debug and servers.listener is not None:
         _, arguments["port"] = sockets.get_address(servers.listener)
         arguments["adapterAccessToken"] = adapter.access_token
```
heejaechang, line 97: Changing `if not session.no_debug:` to `if not session.no_debug and servers.listener is not None:` converts a fail-fast into a silent skip: on a real debug launch where `servers.listener` is unexpectedly `None`, `arguments["port"]`/`arguments["adapterAccessToken"]` are simply not set, so the debuggee is spawned unable to connect back (a hard-to-diagnose "debugger never attaches"). The previous code raised loudly in `get_address(None)`. Prefer asserting the invariant (`assert servers.listener is not None`) for a debug launch so it fails fast rather than dropping the connection info.

heejaechang:
Nice PR — the risky runtime changes (settrace/configure gate, wait_for_response semantics, no-arg Message.__call__, fail-closed auth) are backed by targeted tests, and several concerns a reviewer might raise are already handled with inline rationale. Approving; one small defensive-guard note left inline.

heejaechang (approved):
Approved via Review Center.

src/debugpy/adapter/launchers.py
rchiodo: Fixed - restored fail-fast behavior. For a debug launch (
ot session.no_debug), the code now ssert servers.listener is not None before reading the address/token, so an unexpectedly-missing listener raises loudly instead of silently spawning a debuggee that can't connect back. The assert also narrows the Optional type for the type checker.

src/debugpy/common/singleton.py
```diff
@@ -137,9 +141,9 @@ def __init__(self, *args, **kwargs):
     # with @threadsafe_method. Such methods should perform the necessary locking to
     # ensure thread safety for the callers.
 
-    @staticmethod
     def assert_locked(self):
         lock = type(self)._lock
+        assert lock is not None
         assert lock.acquire(blocking=False), (
             "ThreadSafeSingleton accessed without locking. Either use with-statement, "
```
heejaechang, line 152: 📍 src/debugpy/common/singleton.py:144
Dropping `@staticmethod` correctly makes `assert_locked` callable, but it also activates the side-effecting body: `assert lock.acquire(blocking=False)` *acquires* the lock as its "check" and never releases it — on an `RLock` already held by the current thread it bumps the recursion count (a leaked acquire that the paired `__exit__` won't fully undo), which can cause a later hang. Prefer a side-effect-free check (assert the current thread owns the lock rather than calling `acquire()`). Please confirm `_lock`'s concrete type / call sites — if it's an `RLock` invoked under lock, this is worth fixing before relying on the method.

src/debugpy/adapter/clients.py
```diff
@@ -487,7 +490,11 @@ def attach_request(self, request):
         else:
             if not servers.is_serving():
                 servers.serve(localhost)
-            host, port = sockets.get_address(servers.listener)
+            host, port = (
+                sockets.get_address(servers.listener)
```
heejaechang: 📍 src/debugpy/adapter/clients.py:490
Several of the typing-driven Optional fixes substitute a degraded value on the `None` path instead of failing fast: `attach_request` falls back to `("", 0)` here, `notify_of_subprocess` (L772) lets `body["connect"]["port"] = None` flow through, and `servers.py inject` / `debuggee.py wait_for_exit` (`code = 0`) do the same. These previously raised a loud error at the point the invariant broke; now they push failure downstream as "debugger never attaches" / "clean exit with no debuggee." For real debug flows prefer `assert`/raise on the `None` path — matching the `assert servers.listener is not None` pattern you correctly applied in `launchers.py` — or confirm these branches are genuinely reachable-and-benign.

src/debugpy/adapter/servers.py
```diff
@@ -78,9 +80,10 @@ def __init__(self, sock):
         try:
             self.authenticate()
             info = self.channel.request("pydevdSystemInfo")
-            process_info = info("process", json.object())
-            self.pid = process_info("pid", int)
-            self.ppid = process_info("ppid", int, optional=True)
+            if not isinstance(info, Exception):
+                process_info: Callable[..., int] = cast(Callable[..., int], info("process", json.object()))
+                self.pid = process_info("pid", int)
+                self.ppid = process_info("ppid", int, optional=True)
```
heejaechang: 📍 src/debugpy/adapter/servers.py:80
The `if not isinstance(info, Exception):` guard is effectively dead once `channel.request()` gained `assert not isinstance(result, BaseException)` in this same PR — `request()` now either returns a non-exception or raises, so the branch is always taken. The latent risk: `self.pid`/`self.ppid` are assigned only inside that guard, but `if self.ppid == ():` runs unconditionally right after. Under `-O` (asserts stripped, `info` an Exception), the guard is skipped and you hit `AttributeError` on an unset attribute, since `Connection` (unlike `Session`) has no class-level `pid`/`ppid` default. Either drop the redundant guard (rely on `request()` raising) or add class-level defaults and move the `ppid == ()` normalization inside the guard.

heejaechang:
Overall this is a solid, well-tested typing pass. A few non-blocking notes below about failing-fast vs. substituting degraded fallback values on the newly-added `None` paths, plus one latent `-O`-mode crash path in servers.py. Nothing here blocks merge.

heejaechang (approved):
Approved via Review Center.

src/debugpy/common/singleton.py
rchiodo: Fixed - `assert_locked` now uses a side-effect-free `lock._is_owned()` check (the intended semantic: the current thread must own the lock) instead of `acquire(blocking=False)`/`release()`. `_lock` is confirmed an `RLock`, so the old pattern bumped the recursion count on the held lock and, under `python -O` (assert stripped), ran an unbalanced `release()`. `_is_owned()` avoids both. Annotated the private-API access with `# pyright: ignore[reportAttributeAccessIssue]` since `enableTypeIgnoreComments` is off. (Note: `ThreadSafeSingleton` currently has no subclasses/callers, so this is latent-only.)

src/debugpy/adapter/clients.py
rchiodo: Fixed - switched these None paths from degraded fallbacks to fail-fast asserts, matching the `assert servers.listener is not None` pattern from launchers.py:\n- `attach_request` (clients.py): `assert servers.listener is not None` after `servers.serve()` instead of `("", 0)`.\n- `notify_of_subprocess` (clients.py): `assert listener is not None` on the `port is None` path instead of letting `body["connect"]["port"] = None` through.\n- `inject` (servers.py): `assert listener is not None` instead of connecting the injector to an empty `:0` address.\n- `wait_for_exit` (debuggee.py): `assert process is not None` instead of reporting a bogus clean exit (code 0). This path is only reached after a successful spawn, so the assert documents the invariant; under `-O` it falls back to the existing `except` (code -1), preserving prior semantics.

src/debugpy/adapter/servers.py
rchiodo: Fixed - dropped the redundant `if not isinstance(info, Exception):` guard in favor of `assert not isinstance(info, Exception)` (matching the launchers.py pattern). `channel.request()` either returns a non-exception or raises, so the assert narrows the type and keeps `pid`/`ppid` unconditionally assigned before `if self.ppid == ():` runs — removing the `-O` `AttributeError` path on the unset `ppid`. If the invariant is ever violated the `AssertionError` is caught by the surrounding `except` and the connection is dropped cleanly.

tests/debugpy/common/test_messaging.py
```diff
@@ -437,6 +437,59 @@ def response4_handler(resp):
...
+        channel.start()
+
+        request = channel.send_request("pause")
+
+        # raise_if_failed=False must return the error body instead of raising, even
+        # though a failed response carries an Exception as its body.
+        body = request.wait_for_response(raise_if_failed=False)
+        assert isinstance(body, messaging.MessageHandlingError)
+        assert body is request.response.body
+        assert "pause not supported" in str(body)
```
heejaechang: `assert "pause not supported" in str(body)` is a partial substring check. `MessageHandlingError.__str__` returns the bare reason, so assert it exactly: `assert str(body) == "pause not supported"`. Per the repo's `tests-no-partial-asserts` rule.

tests/debugpy/server/test_api.py
```diff
@@ -0,0 +1,50 @@
...
+    # ensure_logging() writes log files; stub it out for this unit test.
+    monkeypatch.setattr(api, "ensure_logging", lambda: None)
+
+    def failing_settrace(*args, **kwargs):
+        raise RuntimeError("settrace failed")
+
+    monkeypatch.setattr(api.pydevd, "settrace", failing_settrace)
+
+    with pytest.raises(RuntimeError, match="settrace failed"):
+        api._settrace()
```
heejaechang: `pytest.raises(RuntimeError, match="settrace failed")` uses substring/regex matching. The message is controlled by the test itself, so assert it exactly:
```python
with pytest.raises(RuntimeError) as exc_info:
    api._settrace()
assert str(exc_info.value) == "settrace failed"
```
Per the repo's `tests-no-partial-asserts` rule.

tests/debugpy/server/test_api.py
```diff
@@ -0,0 +1,50 @@
...
+    # ... so configure() must not reject the call as "already running".
+    api.configure()
+
+
+def test_settrace_success_blocks_configure(monkeypatch, reset_settrace_called):
+    monkeypatch.setattr(api, "ensure_logging", lambda: None)
+    monkeypatch.setattr(api.pydevd, "settrace", lambda *args, **kwargs: None)
+
+    api._settrace()
+    assert api._settrace.called is True
```
heejaechang: `pytest.raises(RuntimeError, match="already running")` matches the superstring `"debug adapter is already running"`. Assert the exact message instead:
```python
with pytest.raises(RuntimeError) as exc_info:
    api.configure()
assert str(exc_info.value) == "debug adapter is already running"
```
Per the repo's `tests-no-partial-asserts` rule.

src/debugpy/server/api.py
```diff
@@ -73,6 +77,9 @@ def log_to(path):
 
 
 def configure(properties=None, **kwargs):
+    if getattr(_settrace, "called"):
+        raise RuntimeError("debug adapter is already running")
```
heejaechang: `_settrace.called` is a process-global latch that is only ever set `True` and never cleared. The new `configure()` guard therefore permanently raises `"debug adapter is already running"` for the interpreter's lifetime, so any legitimate reconfigure/re-attach after disconnect in a long-lived process is now blocked (previously a second `configure()` proceeded). Confirm whether re-attach-after-disconnect is supported — if so, clear the latch on disconnect/teardown; if not, document the one-shot contract (the tests only cover first-call behavior).

heejaechang:
Overall the added typing and the `_settrace` success-latching fix look good. Two things to resolve before merge: (1) tighten the new/changed test assertions to exact-match per the repo's partial-assert rule, and (2) confirm the intended contract for the new `configure()` "already running" guard, since `_settrace.called` is a process-global latch that is never cleared.

src/debugpy/launcher/output.py
```diff
@@ -95,7 +95,7 @@ def _process_chunk(self, s, final=False):
             while i < size:
                 written = self._stream.write(s[i:])
                 self._stream.flush()
-                if written == 0:
+                if written == 0 and self._fd is not None:
```
heejaechang: If `_fd` was cleared previously and decoder finalization emits buffered text, a zero-byte write bypasses this branch and `i` never advances, causing an infinite loop. Treat every `written == 0` as terminal: conditionally close `_fd`, then break unconditionally.

heejaechang (approved):
Approved via Review Center.

tests/debugpy/server/test_api.py
rchiodo: Fixed by removing the PR-introduced `configure()` guard and its now-obsolete tests. The `already running` exception assertion no longer exists because `configure()` no longer raises that lifetime-wide error.

tests/debugpy/server/test_api.py
rchiodo: Fixed by removing the PR-introduced `_settrace` latch and its now-obsolete tests. The `settrace failed` exception assertion no longer exists because there is no latch state to verify.

src/debugpy/launcher/output.py
rchiodo: Fixed — every zero-byte write now terminates the loop; `_fd` is closed only when still present. Added focused coverage for decoder finalization after `_fd` was already cleared.

src/debugpy/adapter/components.py
```diff
@@ -124,7 +131,7 @@ def report():
         except Exception as exc:
             log.reraise_exception("{0} in {1}", exc, session)
 
-    return Missing()
+    return cast(type, Missing())
```
heejaechang: 📍 src/debugpy/adapter/components.py:134

Standard checking reports `reportInvalidTypeForm` because the value parameter `type` cannot be used as a type expression. Cast to the type variable instead (`cast(T, Missing())`) or otherwise model this factory without using the runtime parameter as a type.

[verified]

src/debugpy/common/log.py
```diff
@@ -359,7 +373,11 @@ def report_paths(get_paths, label=None):
         report("Installed packages:\n")
         try:
             for pkg in importlib_metadata.distributions():
-                report("    {0}=={1}\n", pkg.name, pkg.version)
+                if has_name(pkg):
+                    name = pkg.name
+                    report("    {0}=={1}\n", name, pkg.version)
```
heejaechang: 📍 src/debugpy/common/log.py:378

Standard checking reports that `HasName` has no `version` attribute. Extend the protocol and guard to require both `name` and `version`, or narrow `version` separately before accessing it.

[verified]

src/debugpy/adapter/components.py
rchiodo: Fixed — cast to the type variable `T` (`cast(T, Missing())`) instead of the runtime `type` parameter, so `reportInvalidTypeForm` no longer fires and the return still narrows to `T`.

src/debugpy/common/log.py
rchiodo: Fixed — extended the protocol/guard to require both attributes: `HasNameAndVersion` (`name` + `version`) with a matching `has_name_and_version` `TypeIs` guard that checks `hasattr` for both. The enumeration loop uses the new guard before accessing `pkg.version`, so `version` is no longer reported as missing.

src/debugpy/common/messaging.py
```diff
@@ -464,14 +497,17 @@ def describe(self):
...
+    def payload(self) -> MessageDict | Exception:
         """Payload of the message - self.body or self.arguments, depending on the
         message type.
         """
         raise NotImplementedError
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args, **kwargs) -> MessageDict | Any | int | float:
         """Same as self.payload(...)."""
+        assert not isinstance(self.payload, Exception)
```
heejaechang: 📍 src/debugpy/common/messaging.py:504
`Message.__call__` uses an `assert` to reject exception payloads. Under `python -O`, the assertion is removed, so calls on failed responses can instead return an exception or attempt to call it. Handle exception payloads explicitly so behavior is consistent regardless of optimization mode.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/common/messaging.py:L726](https://github.com/microsoft/debugpy/blob/fdb63ef8549a6cd80a924a4b571f52b42a965a34/src/debugpy/common/messaging.py#L726).

📍 src/debugpy/common/messaging.py:726
`Response._parse` can pass its default `body=None` for a successful response without a body, but `Response.__init__` now accepts only `MessageDict | Exception`. Model empty successful responses in this contract (or normalize them) so standard type checking passes and matches the runtime behavior.

src/debugpy/common/messaging.py
rchiodo: Fixed in d9532bb7 — `Message.__call__` no longer relies on `assert not isinstance(self.payload, Exception)` for control flow. It now checks the payload explicitly and `raise`s the exception when the payload is an `Exception`, so the behavior is identical whether or not assertions are stripped under `python -O` (previously, under `-O` a no-arg call returned the exception object and a call-with-args tried to call the exception). Non-exception payloads keep the same no-arg-returns-payload / forward-args behavior.

rchiodo:
Fixed in d9532bb7 (re: src/debugpy/common/messaging.py:726, empty successful responses). Request.respond now normalizes a None body for a successful no-body response into an empty payload (_payload(None)) before constructing the Response, so Response.body is always a MessageDict or an Exception. This makes standard type checking pass (Response.__init__ keeps its MessageDict | Exception contract) and matches how incoming empty responses are already parsed (Response._parse fills a missing body with an empty dict), so the outgoing and incoming paths are now consistent. The emitted JSON is unchanged — an empty/None body is still omitted from the wire message.

src/debugpy/adapter/clients.py
```diff
@@ -579,9 +585,9 @@ def on_output(category, output):
             request.cant_handle("{0} is already being debugged.", conn)
 
     @message_handler
-    def configurationDone_request(self, request):
+    def configurationDone_request(self, request: messaging.Request):
         if self.start_request is None or self.has_started:
-            request.cant_handle(
+            raise request.cant_handle(
```
heejaechang, line 598: This changes a previously ineffective guard into a user-visible error path, but adds no regression coverage. Add a targeted test for `configurationDone` received before a start request or after startup.

heejaechang (approved):
Approved via Review Center.

src/debugpy/adapter/clients.py
rchiodo: Fixed in e1a5a0fd — added `tests/debugpy/adapter/test_clients.py` with targeted regression coverage for the `configurationDone` guard. It parametrizes the two out-of-order cases the guard now rejects: (1) `configurationDone` received before any start request (`start_request is None`), and (2) after startup has already begun (`has_started is True`). Each case drives `Client.configurationDone_request` directly and asserts it raises `MessageHandlingError`, that a `success: false` failure response with the expected message is sent, and that `has_started` is left unchanged (i.e. the handler does not fall through to the startup/delegate path). I verified the test fails against the previous non-raising `request.cant_handle(...)` guard and passes with the fix.

src/debugpy/adapter/clients.py
```diff
@@ -268,7 +270,7 @@ def handle(self, request):
                 except messaging.MessageHandlingError as exc:
                     exc.propagate(request)
 
-            if self.session.no_debug:
+            if self.session.no_debug and isinstance(request, messaging.Request):
                 self.start_request = request
```
heejaechang, line 277: `handle` is entered only for launch/attach requests, but typing it as `messaging.Message` adds `isinstance(request, messaging.Request)` fallbacks after the entry assertion. With assertions disabled, a non-request message could now skip initialization and responding rather than failing immediately, leaving the client waiting. Type this parameter as `messaging.Request` and remove the redundant guards.

heejaechang (approved):
Approved via Review Center.

src/debugpy/adapter/clients.py
rchiodo: Fixed in e6d38a45 — typed the parameter as `messaging.Request` and removed the redundant `isinstance(request, messaging.Request)` guards on lines 226, 229, and 273. This restores fail-fast behavior even under `python -O` (assertions disabled). Verified pyright reports 0 errors on the file in standard mode.

src/debugpy/server/api.py
```diff
@@ -237,7 +235,10 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 sock.settimeout(None)
                 sock_io = sock.makefile("rb", 0)
                 try:
-                    endpoints = json.loads(sock_io.read().decode("utf-8"))
+                    bytes = sock_io.read()
+                    if bytes is None:
+                        raise EOFError("EOF while reading adapter endpoints")
+                    endpoints = json.loads(bytes.decode("utf-8"))
```
heejaechang: `sock_io.read()` returns `b""` at EOF, not `None`, so this guard never raises the intended `EOFError`; an empty adapter handshake still falls through to `json.loads` and produces a `JSONDecodeError`. Use a non-shadowing name and test for empty data, e.g. `data = sock_io.read(); if not data: raise EOFError(...)`.

heejaechang (approved):
Approved via Review Center.

src/debugpy/server/api.py
rchiodo: Fixed in d3357b1a. You're right — `sock_io.read()` returns `b""` at EOF, not `None`, so the `is None` guard never fired and an empty adapter handshake fell through to `json.loads` producing a `JSONDecodeError`. Renamed the (built-in-shadowing) `bytes` to `data` and changed the guard to `if not data:` so an empty read now raises the intended `EOFError`.

heejaechang (approved):
Approved via Review Center.

StellaHuang95:

src/debugpy/common/singleton.py
```diff
@@ -137,15 +141,18 @@ def __init__(self, *args, **kwargs):
...
+        assert lock is not None
+        # Side-effect-free check: verify the current thread already owns the lock,
+        # rather than acquire()/release() (which mutates the RLock recursion count
+        # and, under `python -O` with the assert stripped, would leak a release()).
+        # _is_owned() is a private but stable RLock helper (used by threading itself).
+        assert lock._is_owned(), (  # pyright: ignore[reportAttributeAccessIssue]
             "ThreadSafeSingleton accessed without locking. Either use with-statement, "
             "or if it is a method or property, mark it as @threadsafe_method or with "
             "@autolocked_method, as appropriate."
         )
```
StellaHuang95, line 155: > **Warning** · Non-blocking recommendation

This changes `assert_locked` from an uncontended-acquire probe to requiring that the current thread already owns the lock. Please add focused coverage for owned and unlocked access (or otherwise document the invariant) so callers outside the expected `with self:` path do not regress unexpectedly.

StellaHuang95 (approved):
Approved via Review Center.

src/debugpy/common/singleton.py
rchiodo: Fixed in 6372ded2 — added `tests/debugpy/common/test_singleton.py` with focused coverage for the new `assert_locked` invariant. It exercises a `ThreadSafeSingleton` subclass through both paths: (1) owned access — inside `with widget:` the current thread owns the lock, so attribute read/write (which route through `assert_locked` → `_is_owned()`) succeed; (2) unlocked access — outside any `with`, no thread owns the lock, so both reading and writing a non-threadsafe attribute raise `AssertionError` (fail-fast). I verified the assertions pass standalone against the current implementation.

StellaHuang95 (approved):
Approved via Review Center.

src/debugpy/common/messaging.py
```diff
@@ -684,6 +727,11 @@ def respond(self, body):
 
         with self.channel._send_message(d) as seq:
             pass
+        if body is None:
+            # A successful response with no body is modeled as an empty payload so
+            # that Response.body is always a MessageDict or an Exception, matching
+            # how incoming empty responses are parsed.
```
StellaHuang95, line 733: > **Info** · Optional note

Changing successful empty responses from `None` to an empty `MessageDict` changes the response contract. Please add coverage for, or confirm, callers that distinguish empty bodies with `is None` so this does not silently alter their behavior.

src/debugpy/adapter/launchers.py
```diff
@@ -89,11 +90,15 @@ def spawn_debuggee(
 
     arguments = dict(start_request.arguments)
     if not session.no_debug:
+        # For a debug launch the server listener must already be up; fail fast
+        # rather than silently spawning a debuggee that can't connect back.
```
StellaHuang95, line 94: > **Info** · Optional note

These new “fail fast” guards use `assert`, which is stripped under `python -O`. If the listener invariant must be enforced at runtime, use an explicit exception consistently with the protocol-boundary checks.

src/debugpy/common/messaging.py
```diff
@@ -754,16 +802,18 @@ class OutgoingRequest(Request):
...
-    _parse = _handle = None
-
     def __init__(self, channel, seq, command, arguments):
         super().__init__(channel, seq, command, arguments)
         self._response_handlers = []
 
+        # Erase the parse and handle methods, as they are not needed for outgoing.
+        setattr(self, "_parse", None)
+        setattr(self, "_handle", None)
+
```
StellaHuang95: > **Info** · Optional note

Using string-keyed `setattr` hides the outgoing-only `_parse` and `_handle` state from Pyright. Please model these as explicit typed optional attributes so the intended override remains statically visible.

src/debugpy/common/json.py
```diff
@@ -93,10 +93,13 @@ def __format__(self, format_spec):
...
     """Convert value (str) to number, otherwise return None if is not possible"""
+    # Only int/float are accepted here (DAP number payloads are int or float);
+    # this deliberately narrows from numbers.Number and does not handle
+    # Decimal/complex/Fraction, which never appear in DAP messages.
     for one_info in classinfo:
-        if issubclass(one_info, numbers.Number):
+        if issubclass(one_info, int) or issubclass(one_info, float):
             try:
                 return one_info(value)
             except ValueError:
```
StellaHuang95, line 105: > **Warning** · Non-blocking recommendation

This intentionally narrows numeric conversion from all `numbers.Number` subclasses to `int` and `float`. Please add a focused regression test that pins the supported numeric types, protecting this shared utility from accidental future expansion or narrowing.

pyproject.toml
StellaHuang95, line 25: > **Warning** · Non-blocking recommendation

This configuration change is bundled with several behavior changes. Please enumerate the non-annotation changes in the PR description, and document the repository-wide prohibition on `# type: ignore` in CONTRIBUTING so contributors understand the expected suppression mechanism.

StellaHuang95 (approved):
Approved via Review Center.

src/debugpy/common/messaging.py
rchiodo: Fixed in c694b9f5 — replaced the string-keyed `setattr(self, "_parse"/"_handle", None)` in `OutgoingRequest.__init__` with explicit class-level typed attributes:

```python
_parse: ClassVar[None] = None  # pyright: ignore[reportIncompatibleMethodOverride]
_handle: ClassVar[None] = None  # pyright: ignore[reportIncompatibleMethodOverride]
```

The override is now statically visible. The `# pyright: ignore` is required (not spurious): the parent `_parse` is a `@staticmethod` and `_handle` is a method, so overriding them with `None` is genuinely an incompatible override under `standard` mode — I verified Pyright reports `reportIncompatibleMethodOverride` on both lines without the suppression, and `0 errors` with it.

src/debugpy/common/json.py
rchiodo: Added in c694b9f5 — `tests/debugpy/common/test_json.py` pins the numeric contract of `json._converter`:

- `int` and `float` string values convert correctly.
- When both `int` and `float` are supplied, the first matching type wins (`type(...) is int`).
- `Decimal`, `complex`, and `Fraction` (all `numbers.Number` subclasses) are parametrized and asserted to return `None` — i.e. intentionally not converted.
- A mixed `(Decimal, int)` classinfo still converts via the supported `int` entry.

I verified each assertion against the current `_converter` implementation. This guards the shared utility against accidental future widening or narrowing.

src/debugpy/common/messaging.py
rchiodo: Confirmed — no caller distinguishes an empty body via `is None`, so this does not silently alter behavior, and I added coverage to pin it (c694b9f5).

Evidence:
- `Response`'s own docstring already documents the contract: for a successful response, `body` is a `MessageDict` (empty dict when `body` was missing/null). Incoming empty responses already parse to `{}` via `Response._parse` → `message_dict("body", _payload)`, never `None`. The `respond(None)` change just makes locally-produced responses consistent with that existing contract.
- I grepped `src/debugpy` (excluding vendored pydevd, which has its own separate `Response`/schema): the only consumer of a `messaging.Response.body` is `adapter/clients.py:638` (`request.respond(response.body)`), which handles dict/`{}`/`None` identically. No `response.body is None` / `== None` checks exist on this type.

Added `test_respond_none_produces_empty_body` in `tests/debugpy/common/test_messaging.py`: a handler calls `request.respond(None)`, and the test asserts `response.body is not None`, `response.body == {}`, `response.success`, and that `"body"` is omitted from the serialized JSON. Verified against the current implementation.

pyproject.toml
rchiodo: Done (c694b9f5):

- **PR description** now enumerates all non-annotation (behavioral) changes in a dedicated section — `clients.py` request_options, `launchers.py` fail-fast assert, the three `messaging.py` changes, `singleton.assert_locked`, `json._converter` narrowing, `util.Observable.observers`, and the `server/api.py` guard hardening — plus a "Suppression policy" section.
- **CONTRIBUTING.md** now has a "Type checking" section documenting that debugpy runs Pyright in `standard` mode with `enableTypeIgnoreComments = false`, so `# type: ignore` has no effect and must not be used; a narrow `# pyright: ignore[reportSpecificRule]` is the expected suppression mechanism, and fixing the underlying typing issue is preferred over suppression.

src/debugpy/adapter/launchers.py
rchiodo: Not taking this one — keeping the `assert` here is deliberate, for two reasons:

1. **Direct reviewer conflict.** This exact guard was added in response to @heejaechang's earlier review comment, which explicitly asked for `assert servers.listener is not None` (both to restore fail-fast and to narrow the Optional for the type checker). Switching it to a raised exception would reverse an accepted change from another reviewer.

2. **Consistency with the module's convention.** `common/messaging.py` uses `assert` for its invariant/protocol-boundary checks (e.g. `assert self.response is None` in `respond`, and the handler-contract asserts in `_handle`). `spawn_debuggee` is not a wire/protocol-boundary against untrusted input — it's an internal invariant on our own adapter state — so an `assert` matches how debugpy expresses these invariants elsewhere. debugpy is not run under `python -O` in practice.

Where `-O` stripping genuinely mattered (control flow that must run under `-O`), I did convert away from asserts in this PR — see `Message.__call__` (now raises explicitly) and `singleton.assert_locked` (side-effect-free check). This listener guard is a pure invariant assertion, so it stays an `assert`. Marking resolved as won't-fix; happy to revisit if debugpy adopts a runtime-enforced invariant convention.

src/debugpy/server/api.py
```diff
@@ -238,7 +236,10 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 sock.settimeout(None)
                 sock_io = sock.makefile("rb", 0)
                 try:
-                    endpoints = json.loads(sock_io.read().decode("utf-8"))
+                    data = sock_io.read()
+                    if not data:
+                        raise EOFError("EOF while reading adapter endpoints")
+                    endpoints = json.loads(data.decode("utf-8"))
```
StellaHuang95, line 242: > **Warning** · Non-blocking recommendation

The PR description says `tests/debugpy/server/test_api.py` covers the behavioral changes, but that file is not included here. Please add coverage for the new empty-read → `EOFError` path in `listen()` and correct the test reference so the documented coverage matches the diff.

StellaHuang95 (approved):
Approved via Review Center.

src/debugpy/server/api.py
rchiodo: Fixed in 2c9236dd — restored `tests/debugpy/server/test_api.py` (the file the PR description references) with focused coverage for the empty-read → `EOFError` path. `test_listen_empty_adapter_read_raises` stubs the adapter-endpoints socket so `sock_io.read()` returns `b""` (EOF) and asserts `listen()` surfaces `error retrieving adapter endpoints: EOF while reading adapter endpoints` — i.e. the `if not data:` guard fires instead of falling through to `json.loads` on empty input. Verified the assertion passes against the current implementation; the documented coverage now matches the diff.

src/debugpy/adapter/clients.py
```diff
@@ -631,7 +637,8 @@ def evaluate_request(self, request):
         def handle_response(response):
             request.respond(response.body)
 
-        propagated_request.on_response(handle_response)
+        if propagated_request is not None:
+            propagated_request.on_response(handle_response)
```
StellaHuang95: > **Issue** · Please address or respond

When `propagated_request` is `None`, this returns `NO_RESPONSE` without registering a handler, so the client's `evaluate` request is never answered. Raise a request error on this path rather than silently suppressing the response.

tests/debugpy/server/test_api.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
StellaHuang95, line 1: > **Warning** · Non-blocking recommendation

The PR description claims a `configure()` already-running guard fix with coverage for success and settrace-failure cases, but this file only adds the adapter-endpoints EOF test and the corresponding guard is absent from the diff. Please add the promised implementation and tests, or correct the PR description to reflect the actual scope.

src/debugpy/adapter/clients.py
rchiodo: Fixed in f29d22f6 — if `channel.propagate(request)` returns `None`, `evaluate_request` now raises `request.cant_handle(...)`, which sends an explicit failure response instead of returning `NO_RESPONSE` with no registered response handler. Added focused coverage that verifies the exact failed `evaluate` response; the targeted adapter client tests pass.

src/debugpy/common/messaging.py
```diff
@@ -464,15 +497,22 @@ def describe(self):
         raise NotImplementedError
```
StellaHuang95, line 497: > **Issue** · Please address or respond

`messaging.py` does not import `from __future__ import annotations`, so this evaluated `MessageDict | Exception` annotation raises at import time on Python 3.8/3.9. The same applies to other new `|` and builtin-generic annotations here and in `adapter/clients.py`. Please postpone these annotations (or use `typing.Union`/`typing.Type`) to preserve compatibility with supported older runtimes.

src/debugpy/adapter/clients.py
```diff
@@ -203,9 +204,9 @@ def initialize_request(self, request):
     #
     # See https://github.com/microsoft/vscode/issues/4902#issuecomment-368583522
     # for the sequence of request and events necessary to orchestrate the start.
-    def _start_message_handler(f):
+    def _start_message_handler(f: Callable[..., Any])-> Callable[..., object | None]:   # pyright: ignore[reportGeneralTypeIssues, reportSelfClsParameterName]
```
StellaHuang95: > **Warning** · Non-blocking recommendation

`reportGeneralTypeIssues` is a broad suppression and conflicts with the new guidance to keep suppressions as narrow as possible. Please narrow this to the specific diagnostic rule(s) produced by this decorator pattern.

rchiodo: Fixed in 27a45ccf — `_start_message_handler` is now explicitly a `@staticmethod`, which matches its class-local decorator role. Pyright no longer interprets `f` as an instance-method `self` parameter, so both `reportGeneralTypeIssues` and `reportSelfClsParameterName` suppressions are removed.

src/debugpy/common/messaging.py
rchiodo: Already addressed — both `src/debugpy/common/messaging.py` and `src/debugpy/adapter/clients.py` now import `from __future__ import annotations`, so these annotations are postponed rather than evaluated at import time. Also, this branch declares `python_requires=">=3.10"` in `setup.py`, with matching 3.10+ classifiers and test environments, so Python 3.8/3.9 are no longer supported targets.

StellaHuang95 (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

Issue #1525, opened 2024-02-27
Display order of properties of a tensor object in debugging


Jesbong:
Can I adjust the display order of properties of a tensor object when debugging Python in vscode? Because in deep learning, we focus more on the shape of a tensor rather than specific values. Now we want to see the value of a tensor by pulling it down a long distance. This has reduced my debugging efficiency.

int19h:
We don't officially support this, but I believe you could author a pydevd extension to customize the representation for your type - have a look at https://github.com/microsoft/debugpy/tree/main/src/debugpy/_vendored/pydevd/pydevd_plugins/extensions. Note however that we will be moving away from pydevd in the future.

However, given that ML/AI is a particularly important scenario these days, I'd love to hear more about the nature of customizations in question, so that we can properly address this in our new debugger.

Jesbong:
The shape(dimension) of a tensor in vscode is far away from the top.  Frequent need to expand and view results in low efficiency.
![v2-d26e166332f23874578a00156232284e_r](https://github.com/microsoft/debugpy/assets/43310393/5c52df79-7102-4786-afb3-b430acb33bb1)


And pycharm can see the type and shape without opening the variable, as shown in the figure. We are very concerned about the dimensions of tensors in ML/AI,  it's so convenient.
![pycharm](https://github.com/microsoft/debugpy/assets/43310393/51cbb0a9-571d-41aa-9858-f6ccdf489526)

Jesbong:
I believe the following issue are all about the same thing. The shape of numpy and tensor is so important for debuger in ML/AI.
https://github.com/microsoft/debugpy/issues/1191
https://github.com/pytorch/pytorch/issues/35071
https://github.com/pytorch/pytorch/issues/63855
https://github.com/microsoft/vscode-python/pull/14244

54nzg:
The shape of a variable cannot be easily viewed.Whether mouse hover will be supported in the future to see the shape of the variable.

makaspacex:
It is obvious that everyone is very satisfied with the debugging of pycharm. If there are no patents or other restrictions, you can copy the display stype and information of pycharm. If there are restrictions, ignore my comments pls.

makaspacex:
In addition to tensor and numpy, pycharm also has good type display support for list, tuple, dict, etc.

haduoken:
I tempory use this
[...]

Issue #1191, opened 2023-01-24
Change debugger info of pytorch tensor from value to shape


shenmishajing:

int19h:
We use `repr()` normally on the assumption that whatever the type author made it do is the preferred output (since it is also used in REPL), so ideally it should be fixed on that end so that the whole tooling ecosystem benefits. 

However, if that is not feasible, it can be special-cased. Can you tell more about the scenario, and why one representation is superior to the other here?

shenmishajing:
I find some issue in pytorch repo about this, like pytorch/pytorch#63855 and pytorch/pytorch#35071. Seems like there are already many people ask for this feature, but pytorch does not do this, as those two issues have been stable for two or three years. I leave some commets on them, but I'm pessimistic about this. Therefore, let's make this special-cased. 

First, when we are debugging some pytorch code, there maybe some error in that code, which are almost caused by wrong shape. For example, a fc layer (technically matrix multiply) will raise error when A(3*4 matrix) @ B(5*4 matrix). If vscode display the shape info of those two tensor, we can fix this by A@B.T immediately. But if vscode display some thing like `tensor([[[ -38.3458,  214.4476,  767.8055,  ...` and `tensor([[[  29.1078,  215.7885,  676.3702,  ...`, they are helpless and we can not fix it until we print shape of them.

Second, we init network randomly or load pre-train checkpoint, most of the time. So the value of network's weight is just random or meanless to human. In fact, not just network's weight, almost all of tensors except just some input tensor are random or meanless. On the other hand, if we want to check whether the value of tensor is correct or not, we check the loss or metrics indirectly or display some heat map to visualize them instead of check the value directly.

For now, vscode display the value of tensor when debug, It looks like: `tensor([[[ -38.3458,  214.4476,  767.8055,  ...` .
<img width="456" alt="截屏2023-01-25 10 06 56" src="https://user-images.githubusercontent.com/32559715/214464651-3c3346ac-982c-4824-a09d-59e7990042ba.png">
It will be better, if we add the shape info before the value. It will look like: `tensor(shape=torch.Size([32, 80, 108]), [[[ -38.3458,  ...`. If this is hard to implement, it will be ok if we just add the shape info before `tenosr`. It will look like: `shape=[32, 80, 108], tensor([[[ -38.3458,  ...` or just `[32, 80, 108], tensor([[[ -38.3458,  ...`. This will be easy to implement because we just add shape str and original repr str.

PR #1661, opened 2024-08-29
to display lentgh/shape, addressing #1525, #1191, #4244, etc.

qige96:
Displays the size of an Sized variable before displaying its value, such as np.ndrray, torch.Tensor. Can benefit ML/AI programming, as well as other programming that works heavily on large collections.


qige96:

rchiodo (changes requested):
pydevd is actually sourced from here:
https://github.com/fabioz/PyDev.Debugger

We just pull it from there and reship it. If you submit there, we should be able to consume this when we pull pydevd again.

qige96:
As requested, I created a [pull request there](https://github.com/fabioz/PyDev.Debugger/pull/288). Thanks for your prompt reply!

Issue #1689, opened 2024-10-09
Support 3.13 debugging


rchiodo:
Upon the release of Python 3.13, debugging a 3.13 app doesn't currently work. There are multiple locations where [thread internals are different](https://github.com/python/cpython/commit/33da0e844c922b3dcded75fbb9b7be67cb013a17#diff-b250d3d0e0d909784a67083e9069ce507513a67e8514a011e2f364027e25e0b0). Pydevd uses these internals, so it needs to be changed to accommodate the changes in CPython. 

Our team is currently investigating and working to support this scenario.

rchiodo:
This just shipped and is available in [Pypi](https://pypi.org/project/debugpy/). It should be available soon in the [debugger extension](https://marketplace.visualstudio.com/items?itemName=ms-python.debugpy) for VS code.

PR #1692, opened 2024-10-09
Get Python 3.13 to work

rchiodo:
There were a number of internal changes in CPython that were preventing 3.13 from working. 

This change fixes all of those spots to use a public API or something that works in all versions.

Addresses https://github.com/microsoft/debugpy/issues/1689


rchiodo:
I believe this also fixes attach? Still verifying, but all the attach tests were passing for me locally.

azure-pipelines/pipelines.yaml
```diff
@@ -65,6 +65,8 @@ jobs:
           python.version: "3.11"
         py312:
           python.version: "3.12"
+        py313:
+          python.version: "3.12"
```
karthiknadig, line 69: This should be `3.13`

karthiknadig (changes requested):
Minor change

azure-pipelines/pipelines.yaml
rchiodo: Oh thanks, totally missed that.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
```diff
@@ -35,7 +35,12 @@ def add_exception_to_frame(frame, exception_info):
 
 
 def remove_exception_from_frame(frame):
-    frame.f_locals.pop("__exception__", None)
+    # In 3.13 frame.f_locals became a proxy for a dict, so we need to copy it to avoid
```
gramster: Grammar? Can't parse this sentence

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
```diff
@@ -140,6 +145,7 @@ def map_columns_to_line(self, original_line: str):
 def _utf8_byte_offset_to_character_offset(s: str, offset: int):
     byte_offset = 0
     char_offset = 0
+    offset = offset if offset is not None else 0
```
gramster, line 148: You could just write:

    offset = offset or 0

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
rchiodo: Sorry I didn't read this after copilot generated it. Should be fixed now.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
rchiodo: I wondered about that. It works that way in TypeScript, wasn't sure if boolean logic worked that way in Python

karthiknadig (approved):

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_constants.py
```diff
@@ -173,9 +173,11 @@ def _current_frames():
 IS_PY310_OR_GREATER = sys.version_info >= (3, 10)
 IS_PY311_OR_GREATER = sys.version_info >= (3, 11)
 IS_PY312_OR_GREATER = sys.version_info >= (3, 12)
+IS_PY313_OR_GREATER = sys.version_info >= (3, 13)
+IS_PY314_OR_GREATER = sys.version_info >= (3, 14)
 
-# Not currently supported in Python 3.12.
-SUPPORT_ATTACH_TO_PID = not IS_PY312_OR_GREATER
+# Not currently supported in Python 3.14.
+SUPPORT_ATTACH_TO_PID = not IS_PY314_OR_GREATER
```
debonte, line 180: Do we know that attach doesn't work in 3.14 for some reason?

rchiodo: It might work, but the code in the attach logic explicitly only allows up to 3.13. This is just matching that logic.

PR #1694, opened 2024-10-10
Directions on how to debug pydevd

rchiodo:
Added some instructions on how one would update and debug pydevd


CONTRIBUTING.md
```diff
@@ -117,9 +117,117 @@ You might need to regenerate the Cython modules after any changes. This can be d
...
+    pydevd_log.debug("Some test logging", frame, etc)
+```
+
+Make sure if you add this in a module that gets `cythonized`, that you turn off `Cython` support as listed above. Otherwise you'll have to regen the C code or you won't actually see your new log output.
+
+#### How to use logs to debug failures
+
+Debugging log failures can be done in multiple ways. 
+
+If you have an existing test failing, you can debug it by running the test with the main branch and comparing the results. To do so you would:
```
debonte, line 187: I'd suggest changing the word "debug" to "investigate". I was expecting that these steps were going to somehow allow me to debug the debugger.
```suggestion
If you have an existing test failing, you can investigate it by running the test with the main branch and comparing the results. To do so you would:
```

debonte (approved):

AdamYoblick (approved):

PR #1704, opened 2024-10-16
Support GraalPy

timfel:
This change seems to be the only thing that keeps debugpy from working on GraalPy on my machine. Would be great if it could be included :-)


rchiodo:
There's some flakiness in the 3.12 tests. They seem to randomly timeout. Hopefully I can get them to pass in this PR:
https://github.com/microsoft/debugpy/pull/1706

And then merge this PR.

Issue #1720, opened 2024-11-04
TypeError: 'None' comparison in collect_try_except_info()


Sorontik:
Hello,

when an exception occured in my Program, i got this output in the debug console:

> 79.61s - Error collecting try..except info from source (src/main.py)
Traceback (most recent call last):
  File "c:\Users\<username>\.vscode\extensions\ms-python.debugpy-2024.12.0-win32-x64\bundled\libs\debugpy\_vendored\pydevd\pydevd.py", line 796, in collect_try_except_info
    if line > max_line:
       ^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'int'

Unfortunately, the Program is rather complex and i was unable to reproduce this error in a simplified Version of my script.

This did not happen previously, so it might be related to the changes in one of the last releases (I'm currently on v2024.12.0 which should be the latest - on a up-to-date VS-Code)

rchiodo:
Thanks for the issue. This was caused by changes in 3.12. Similar to this issue here:
https://github.com/microsoft/debugpy/issues/1708

PR #1721, opened 2024-11-04
In 3.12 findlinestarts seems to return None for line more often

rchiodo:
Fixes https://github.com/microsoft/debugpy/issues/1720

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -791,10 +791,10 @@ class PyDB(object):
                     max_line = -1
                     min_line = sys.maxsize
                     for _, line in dis.findlinestarts(code_obj):
-                        if line > max_line:
+                        if line is not None and line > max_line:
                             max_line = line
 
-                        if line < min_line:
+                        if line is not None and line < min_line:
                             min_line = line
 
                     try_except_infos = [x for x in try_except_infos if min_line <= x.try_line <= max_line]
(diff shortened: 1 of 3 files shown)


debonte (approved):

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -791,10 +791,10 @@ def collect_try_except_info(self, code_obj):
                     max_line = -1
                     min_line = sys.maxsize
                     for _, line in dis.findlinestarts(code_obj):
-                        if line > max_line:
+                        if line is not None and line > max_line:
```
debonte, line 794: I'm confused about how `line` can be `None` since `dis.findlinestarts` checks for `None` before yielding.

rchiodo: I'm not sure what makes you say that? The docs at least say it can be none. dis.findlinestarts is supposed to call: https://docs.python.org/3/reference/datamodel.html#codeobject.co_lines.

I didn't check the source in CPython though

debonte: Oh, I didn't notice that it wasn't our code. I just used "go to def"` and saw this. Turns out this is from my CPython 3.12.7 install.

```python
def findlinestarts(code):
    """Find the offsets in a byte code which are start of lines in the source.

    Generate pairs (offset, lineno)
    """
    lastline = None
    for start, end, line in code.co_lines():
        if line is not None and line != lastline:
            lastline = line
            yield start, line
    return
```

PR #1722, opened 2024-11-05
Build pydevd binaries before running tests

AdamYoblick:
We're removing these binaries from the repo and adding a .gitignore entry for them, since we want to build them ourselves. So we need to make sure we build them before running tests, because the attach tests will fail if those binaries don't exist.


rchiodo (approved):
Looks like the path is off? But looks good otherwise. Not sure why the path is off.

PR #1723, opened 2024-11-05
Remove compiled files so we can generate them ourselves

rchiodo:
Removing all of the compiled dlls for attach so we can generate them ourselves.


.gitignore
```diff
@@ -5,6 +5,10 @@ __pycache__/
 
 # C extensions
 *.so
+*.dylib
```
AdamYoblick, line 8: I'm not sure if there are any other files with these extensions in our repo. Should be consider explicitly excluding these files from `src\debugpy\_vendored\pydevd\pydevd_attach_to_process` ?

AdamYoblick (approved):

.gitignore
rchiodo: There isn't any but I think we shouldn't ever be checking in dlls/exes. Too opaque.

rchiodo:
Can you just submit this Adam? It's not performing the license test for some reason

rchiodo:
Okay the 3.13 tests all passed locally with the same binaries as built by the PR. Submitting.

PR #1725, opened 2024-11-07
Build and publish pydevd binaries

AdamYoblick:
- Build pydevd binaries and publish them as pipeline artifacts
- Download published artifacts in test jobs and copy them to the appropriate location, before tests run
- Move lint and pydevd to their own stages for a cleaner build
- Yaml cleanup

I ran the test pipeline manually pointing at this pr branch, the run is at https://dev.azure.com/debugpy/debugpy/_build/results?buildId=4332&view=results

### DIFF
--- a/azure-pipelines/pipelines.yaml
+++ b/azure-pipelines/pipelines.yaml
@@ -18,7 +18,7 @@ pr:
   drafts: false
 
 variables:
-  architecture: "x64"
+  architecture: x64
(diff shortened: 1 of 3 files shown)


rchiodo (approved):

bschnurr:
we dont need to import the new pipeline template?

AdamYoblick:
That's only required for production pipelines afaik. This is just a PR pipeline. 

Debugpy has different pipelines for CI and PR. It's not how I would have set it up but that's how it was when I joined :)

PR #1744, opened 2024-11-22
Remove stale readme, our CI build compiles all these binaries now

AdamYoblick:


rchiodo (commented):
I feel like the first part of the readme is still valid. It gives directions on how this folder is used.

Issue #460, opened 2020-11-03
Cannot Attach again after disconnect


JohanMabille:
## Environment data

- debugpy version: 1.1.0
- OS and version: Linux 5.4.60-gentoo-x86_64
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.8.6 from Miniconda
- Using VS Code or Visual Studio: No

## Actual behavior

A "disconnect" request does not seem to really terminate the debug adapter as stated [here](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect). It looks like the adapter is still running in the background and that you can reconnect with "initialize" request.

## Expected  behavior

After a "disconnect", the debug adapter should be restarted the same way it was started the first time.

## Steps to reproduce:
1. Execute the following python code in the debugee
```python
import debugpy
debugpy.listen(5416);
```
2. Send "initiliaze" and "attach" requests.
3. After receiving an attach response, send a "disconnect request";
4. Repeat step 1. `debugpy.listen` never returns.

If I skip step 4 and repeat step 2, everything works fine.

int19h:
`listen()` is not supposed to be called multiple times. It starts a debug server in your process, and it's a one-off operation. The server should handle new connections (only one at a time) for as long as the process is running, though.

That said, "disconnect" request indeed doesn't terminate the adapter, and that's by design. The adapter is necessary for the debug server; and the server, once started, cannot be fully excised from the process for various reasons. The intent of the "disconnect" request, in any case, is to signal to the adapter that debugging should be stopped. It handles that accordingly, by disabling tracing (breakpoints etc) as much as Python implementation details allow.

JohanMabille:
Thanks for the quick answer!

I was surprised by this behavior because of the specification of the DAP and its difference with PTVSD, but this seems a reasonable behavior anyway. Do you think it could be worth adding a comment about this difference in this [wiki page](https://github.com/microsoft/debugpy/wiki/Switching-over-from-%60ptvsd%60-to-%60debugpy%60)? I would be glad to do it myself, but I canont open a PR on a waiki page.

int19h:
It should be https://github.com/microsoft/debugpy/wiki/API-Reference - I've added a clarification on `listen()` there.

[...]

PR #1751, opened 2024-12-03
Properly update pydevd._settrace.called

lukejriddle:
`api._settrace.called` is intended to confirm that a process had been traced before. However, returning in a `try` block prevents the subsequent `else` block from executing, so `api._settrace` was not updating this flag.

This change just removes the return. `api._settrace` will still return `None`, as `pydevd.settrace` was returning `None` before.

### DIFF
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -42,7 +42,7 @@ def _settrace(*args, **kwargs):
     # The stdin in notification is not acted upon in debugpy, so, disable it.
     kwargs.setdefault("notify_stdin", False)
     try:
-        return pydevd.settrace(*args, **kwargs)
+        pydevd.settrace(*args, **kwargs)
     except Exception:
         raise
     else:
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -45,11 +45,9 @@ def _settrace(*args, **kwargs):
         pydevd.settrace(*args, **kwargs)
     except Exception:
         raise
-    else:
-        _settrace.called = True
 
 
-_settrace.called = False
+_listen.called = False
 
 
 def ensure_logging():
@@ -78,9 +76,6 @@ def log_to(path):
 
 
 def configure(properties=None, **kwargs):
-    if _settrace.called:
-        raise RuntimeError("debug adapter is already running")
-
     ensure_logging()
     log.debug("configure{0!r}", (properties, kwargs))
 
@@ -104,9 +99,6 @@ def configure(properties=None, **kwargs):
 
 def _starts_debugging(func):
     def debug(address, **kwargs):
-        if _settrace.called:
-            raise RuntimeError("this process already has a debug adapter")
-
         try:
             _, port = address
         except Exception:
@@ -116,7 +108,7 @@ def _starts_debugging(func):
             port.__index__()  # ensure it's int-like
         except Exception:
             raise ValueError("expected port or (host, port)")
-        if not (0 <= port < 2 ** 16):
+        if not (0 <= port < 2**16):
             raise ValueError("invalid port number")
 
         ensure_logging()
@@ -150,10 +142,14 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
     # Errors below are logged with level="info", because the caller might be catching
     # and handling exceptions, and we don't want to spam their stderr unnecessarily.
 
+    if _listen.called:
+        # Multiple calls to listen() cause the debuggee to hang
+        raise RuntimeError("debugpy.listen() has already been called on this process")
+
     if in_process_debug_adapter:
         host, port = address
         log.info("Listening: pydevd without debugpy adapter: {0}:{1}", host, port)
-        settrace_kwargs['patch_multiprocessing'] = False
+        settrace_kwargs["patch_multiprocessing"] = False
         _settrace(
             host=host,
             port=port,
@@ -218,7 +214,10 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
         try:
             global _adapter_process
             _adapter_process = subprocess.Popen(
-                adapter_args, close_fds=True, creationflags=creationflags, env=python_env
+                adapter_args,
+                close_fds=True,
+                creationflags=creationflags,
+                env=python_env,
             )
             if os.name == "posix":
                 # It's going to fork again to daemonize, so we need to wait on it to
@@ -288,6 +287,7 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
         **settrace_kwargs
     )
     log.info("pydevd is connected to adapter at {0}:{1}", server_host, server_port)
+    _listen.called = True
     return client_host, client_port
 
 
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -47,9 +47,6 @@ def _settrace(*args, **kwargs):
         raise
 
 
-_listen.called = False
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

rchiodo (changes requested):
It looks like this is causing attach tests to fail on Linux. Not sure why, but if you follow the directions in the CONTRIBUTING.md, you should be able to run the failing tests yourself.

rchiodo:
Oh we changed the repository to not have the binaries checked in. You have to build pydevd first. Forgot about that. I should update the contributing.md.

Basically:
- `cd src/debugpy/_vendored/pydevd`
- Set PYTHONPATH=.
- Run `python build_tools/build.py`
-

lukejriddle:
Still not able to get the tests working. That build script didn't seem to build the attach .so, so I manually ran `pydevd_attach_to_process/linux_and_mac/compile.sh` as well. This fixed the error of not finding the shared library, but now I get an opaque error about gdb having an exit code of 1.

rchiodo:
> This fixed the error of not finding the shared library, but now I get an opaque error about gdb having an exit code of 1.

That's likely the attach failing. You'll have to add extra logging to the attach code I'm guessing.

rchiodo:
Debugging tests failures is all looking at log files. If there's not enough information, you add more logging. Attach happens in this code here:
https://github.com/microsoft/debugpy/blob/a78e5c28cd3da4b1abd2c289de7070ecb29778b1/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/attach.cpp#L56

You'd add more printfs and then rebuild the .so files again.

lukejriddle:
Okay, figured out why the tests are failing. [In the test](https://github.com/microsoft/debugpy/blob/main/tests/debugpy/test_attach.py#L161), multiple adapters are attached to the same PID as `session1` and `session2`. Before our change, the [check for _settrace.called](https://github.com/microsoft/debugpy/blob/main/src/debugpy/server/api.py#L81) would never pass, allowing this behavior. Now, attaching multiple adapters to the same PID is essentially forbidden.

This is likely how it should work, though, and has even been requested in a [previous issue](https://github.com/microsoft/debugpy/issues/460), since `api.listen` should only be called once-per-process.

I can update the test to expect this failure, if that works with you @rchiodo

rchiodo:
> I can update the test to expect this failure, if that works with you @rchiodo

I believe the test is trying to reattach. We definitely want that to work. Just allowing the failure would be incorrect. Maybe detach has to set the trace flag to false and then the test will pass.

lukejriddle:
@microsoft-github-policy-service agree company="Meta"

lukejriddle:
Okay, how about this? Changed from `_settrace.called` to `_listen.called`, since its the multiple invocations of `listen` that will cause the debuggee to hang; multiple calls to `settrace` don't, as demonstrated by the attach tests.

rchiodo:
I think you need to run ruff? The linter is failing.

Not sure about using listen instead. Sounds plausible, but the old settrace check also made sure that connect wasn't called more than once. Although maybe connect being called more than once is fine-ish? Connect while you're actually debugging and then reconnect after disconnect should be okay, but connect and then connect again isnt.

But the original code was letting that happen anyway so maybe it's fine.

lukejriddle:
Whoops, did a typo, just pushed the fix as well as the lint issues.

As a follow up, I can look into preventing additional `connect` calls to the same process if the debuggee already has an adapter connected. Otherwise behavior is pretty much unchanged other than preventing listen from hanging a process

--
Btw, I'm not sure what's going on with tox/pytest, but there are some tests (like the ones in `test_attach.py`) that pass when run in series using -n0 but fail if taken by a worker. Not sure if it's a config issue on my part

rchiodo:
> Btw, I'm not sure what's going on with tox/pytest, but there are some tests (like the ones in test_attach.py) that pass when run in series using -n0 but fail if taken by a worker. Not sure if it's a config issue on my part

Pytest doesn't play well with the way debugpy is running. Timeouts and such are killing processes before they're done. Mostly only happens on windows though.

lukejriddle:
Not sure about these test failures... `test_attach_pid_client[str]` passes but `[int]` fails. And only in some environments...: `AttributeError: 'GdbRemoveReadlineFinder' object has no attribute 'find_spec'`

lukejriddle:
Hmmm, it looks like it is, from those logs. And we also see the expected `listen_exception` message sent through the backchannel:

```
I+00001.266: listen() failed:
             
               File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\server\api.py", line 130, in debug
                 return func(address, settrace_kwargs, **kwargs)
               File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\server\api.py", line 144, in listen
                 raise RuntimeError("debugpy.listen() has already been called on this process")
             RuntimeError: debugpy.listen() has already been called on this process
             
             Stack where logged:
               File "C:\Users\VssAdministrator\AppData\Local\Temp\pytest-of-unknown\pytest-0\popen-gw0\test_multiple_listen_raises_ex0\code_to_debug.py", line 12, in <module>
                 debugpy.listen(address=(host, port))
               File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\public_api.py", line 31, in wrapper
                 return wrapped(*args, **kwargs)
               File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\server\api.py", line 132, in debug
                 log.reraise_exception("{0}() failed:", func.__name__, level="info")
               File "D:\a\1\s\.tox\py39\lib\site-packages\debugpy\common\log.py", line 222, in reraise_exception
                 _exception(format_string, *args, **kwargs)
             

D+00001.266: backchannel <-- listen_exception
```

If this backchannel message weren't received we'd see an AssertionError, not `worker 'gw5' crashed while running 'tests/debugpy/test_attach.py::test_multiple_listen_raises_exception[program]'`

lukejriddle:
I can't repro these issues on Linux py3.10 or Mac py3.13. Can I add similar skip rules like the other tests in this file? https://github.com/microsoft/debugpy/blob/main/tests/debugpy/test_attach.py#L156

rchiodo:
> I can't repro these issues on Linux py3.10 or Mac py3.13. Can I add similar skip rules like the other tests in this file? https://github.com/microsoft/debugpy/blob/main/tests/debugpy/test_attach.py#L156

I think the root cause of those failures is just that pytest is flakey. Next run they'd likely pass. Ideally we'd made all of these things more deterministic.

I'd rather not skip them though cause eventually we end up skipping everything. Maybe a retry?

lukejriddle:
Yeah sounds good. Do I have permissions to retry?

rchiodo:
Wait did you change anything with the new double listen test? I think the test has to somehow manage the debugger crashing. I don't see another test doing that but I think it might just work if you put a try except around the code like so:

```python
try:
  with debug.Session() as session:
        session.open_backchannel()
        session.spawn_debuggee(
            [
                code_to_debug,
                host,
                port,
            ]
        )
        session.wait_for_adapter_socket()

        session.expect_server_socket()
        session.connect_to_adapter((host, port))
        with session.request_attach():
            pass
except:
   pass
```

And then assert that the backchannel got the exception.

lukejriddle:
Was able to repro locally. Updated the test to finish the lifecycle of the adapter, went from failing to passing so should hopefully work in the pipeline

rchiodo (approved):

rchiodo:
Thanks for the work :)

PR #1779, opened 2024-12-12
Remove bytecode distinfo causing PyPi publish errors and ignore it from now on

AdamYoblick:
This dist-info is causing ESRPRelease to publish to the `bytecode` project instead of the `debugpy` one. We need to get rid of it and ignore it so future `subrepo` commands from pydevd don't bring it back.


rchiodo (approved):

AdamYoblick:
The test failures have nothing to do with my changes, so I'm going to merge in.

PR #1797, opened 2025-01-06
Fix pyqt6 loader on windows

rchiodo:
Was testing matplotlib after a submission this morning and I needed this change for the QT backend to work.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydev_ipython/qt_loaders.py
+++ b/src/debugpy/_vendored/pydevd/pydev_ipython/qt_loaders.py
@@ -145,25 +145,28 @@ def has_binding(api):
             return True
 
     except ModuleNotFoundError:
-        from importlib import machinery
+        try:
+            from importlib import machinery
 
-        # importing top level PyQt4/PySide module is ok...
-        mod = __import__(module_name)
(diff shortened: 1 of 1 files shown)


src/debugpy/_vendored/pydevd/pydev_ipython/qt_loaders.py
```diff
@@ -145,25 +145,28 @@ def has_binding(api):
...
-
-        # ...importing submodules is not
-        loader_details = (machinery.ExtensionFileLoader, machinery.EXTENSION_SUFFIXES)
-        submod_finder = machinery.FileFinder(mod.__path__[0], loader_details)
-        submod_check = (
-            submod_finder.find_spec("QtCore") is not None
-            and submod_finder.find_spec("QtGui") is not None
-            and submod_finder.find_spec("QtSvg") is not None
-        )
+        try:
```
rchiodo, line 148: This code was trying to load `PyQt4` twice, when `PyQt4` doesn't even exist anymore (at least on windows)

rchiodo: If it fails to load the main module, it should just return False.

Issue #1765, opened 2024-12-11
Stepping in a multithreaded application jumps to other threads


rchiodo:
### Discussed in https://github.com/microsoft/debugpy/discussions/1761

<div type='discussions-op-text'>

<sup>Originally posted by **pub-pub** December 10, 2024</sup>
i posted my initial issue [here](https://github.com/microsoft/vscode-python-debugger/issues/515) and it was successfully resolved.

now i'm using `"steppingResumesAllThreads": false` setting for debugging. but this solutions has a drawback, it pauses all other threads.

is it possible to add one more setting that will keep debugger Stepping inside my initial thread without pausing all other threads?</div>

rchiodo:
Here's some source code to reproduce the problem:

```python
import threading
import time
import requests

def request_get(url):
    # return "abc"
    with requests.get(url) as data:
        return data.text

def watcher():
    while True:
        print(1)
        time.sleep(3)

def worker ():
    while True:
        print(len(request_get("https://dns.google//")))
        time.sleep(5)

threading.Thread(daemon=True, target=watcher).start()
t1 = threading.Thread(target=worker)
t1.start()
t2 = threading.Thread(target=worker)
t2.start()
t1.join()
t2.join()
```

Set a breakpoint on line 11 (the print(1)) in the watcher and step. It will eventually end up in the other thread

xmatthias:
I'm encountering the same - though in a pretty large codebase where so far i didn't take the time to create a reproduction.

It's not just "on step" for me, either - also the initial breakpoint may stop - shows you that code for a second - and then (as other threads stop) it'll jump to other parts of the code (other files, running in other threads), pulling me away from the code i wanted to look at / debug, having me go back to the original. 
it's happening on step too - but it's not limited to "step" as far as i did experience.

it's driving me nuts if I'm honest - and I'm pretty certain that it wasn't the case a few months ago (I'm working on this same, multithreaded, partially async codebase since about 5 years now).

tbrisker-april:
I've also been encountering the same issue recently, and it seems to be a regression introduced in the recent months, in the past there was no issue debugging a multithreaded app without the debugger jumping to other threads.

rchiodo:
This might be a problem only related to 3.12 and above. #1792 indicated it started with the changes for sys.monitoring. I can't reproduce the issue with 3.11.

rchiodo:
[...]

PR #1798, opened 2025-01-07
Fix multithreading stepping in 3.12 and later

rchiodo:
Fix for https://github.com/microsoft/debugpy/issues/1765

After adding a bunch of logging I found that one thread would send a breakpoint event, 'return' events were still being fired for other threads. When the 'breakpoint' thread suspended itself and all other threads, it sets the 'step' mode for all other threads to step into, so when those return events came in, we were getting other threads in step mode. This makes the debugger jump all over the place when stepping (as all threads look like they're stepping).

The fix here was to just ignore any events after a thread is suspended. Other threads shouldn't be sending a separate stopped event after they've been suspended.

Added a test to pydevd to verify this works.

### DIFF
--- a/src/debugpy/.vscode/settings.json
+++ b/src/debugpy/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "."
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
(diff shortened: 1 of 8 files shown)


src/debugpy/.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
AdamYoblick, line 1: Is this file supposed to be in source control? It might be, I just wasn't sure since it was in the `.vscode` directory

AdamYoblick (approved):

src/debugpy/.vscode/settings.json
rchiodo: It can be. It's common settings. It makes debugpy tests runnable from the test runner in VS code.

Issue #1301, opened 2023-05-26
Path mappings not working with mixed windows-style/linux path slash


gionapaolini:
I have a launch.json configuration to attach the debugger to a remote service.
This used to work for a long time, not sure when it stopped working exactly.
```
{
            "name": "Service1",
            "type": "python",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 6005
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceRoot}/Source",
                    "remoteRoot": "/app"
                }
            ],
            "logToFile": true,
            "justMyCode": false
        }
      ...
}
```
Gives error:
```
pydev debugger: unable to find translation for: "c:\path\to\workspace\Source\...\app.py" in ["C:\path\to\workspace/Source/","C:\path\to\workspace/Source"] (please revise your path mappings).
```

The local root is translated into 
`C:\path\to\workspace/Source/` (notice the forward slashes at the end, and capital C )
While the pydev debugger is looking for 
`c:\path\to\workspace\Source\...\app.py`

I tried several ways to make it work, by replacing the ${workspaceRoot} variable with the value with lower c, and correct slashes. 
But the only way I managed to make it work is to map **exactly** the full path of the specific file I want to debug. That's the only way I could see and hit the breakpoint on vscode window.
```
            "pathMappings": [
                {
                    "localRoot": "c:\\path\\to\\workspace\\Source\\...\\app.py",
                    "remoteRoot": "/app/../app.py"
                }
            ]
```

There are other ways I have tried, many variations with the slashes, but none of them worked. Also sometimes I don't receive the error above, but rather only `Breakpoint in file that does not exist`.

Is this a vscode issue or debugpy issue?

gramster:
@karthiknadig, does this seem like debugpy or PVSC?

karthiknadig:
For this can you try adding `"clientOS": "windows"` to the configuration? This can occur when debugpy is not able to identify the client OS from the server side.

karthiknadig:
If you are using the latest version of python extension which is 2023.8.* and debugpy 1.6.7 and you are seeing this issue let us know. It could also be a bug in Python extension where the client OS is not being explictly sent to debugpy.

andreeas26:
> If you are using the latest version of python extension which is 2023.8.* and debugpy 1.6.7 and you are seeing this issue let us know. It could also be a bug in Python extension where the client OS is not being explictly sent to debugpy.

[...]

PR #1803, opened 2025-01-09
Allow relative paths in the 'remotePath' for a launch.json

rchiodo:
Partial fix for https://github.com/microsoft/debugpy/issues/1301

Users in issue 1301 were attempting to do something like so:

```jsonc
"pathMappings": [
                {
                    "localRoot": "${workspaceFolder}",
                    "remoteRoot": "." // <-- '.' is not supported here.
                }
            ]
```

In pydevd this isn't supported. The remote and local roots have to be fully specified. This is a pain in a shared situation though as you can pass environment variables for the local path but not the remote.

This change allows the user to have a relative (to the current working directory) for the remote path.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_file_utils.py
+++ b/src/debugpy/_vendored/pydevd/pydevd_file_utils.py
@@ -653,12 +653,6 @@ def _fix_path(path, sep, add_end_sep=False):
 
     if sep != "/":
         path = path.replace("/", sep)
-
-    if path.startswith("."):
-        # We need the full path if this relative because all comparisons below
-        # check if the path is substring of another
-        path = os.path.realpath(path)
-
     return path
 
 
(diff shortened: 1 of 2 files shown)


src/debugpy/_vendored/pydevd/pydevd_file_utils.py
```diff
@@ -747,6 +753,16 @@ def setup_client_server_paths(paths):
         initial_paths_with_end_sep.append((path0, path1))
         paths_from_eclipse_to_python_with_end_sep.append((_normcase_from_client(path0), normcase(path1)))
 
+    if DEBUG_CLIENT_SERVER_TRANSLATION:
+        pydev_log.critical(
```
AdamYoblick, line 757: This is crazy to read 😆

AdamYoblick (approved):

Issue #1633, opened 2024-07-22
interactive plot window not showing under wayland in debug mode for matplotlib=3.9


suxpert:
## Environment data
- debugpy version: 1.8.2
- OS and version: OpenSUSE Tumbleweed (20240716), KDE Plasma 6 under wayland.
- Python version (& distribution if applicable, e.g. Anaconda): python 3.11.9 from conda-forge (mamba)
- Using VS Code or Visual Studio:
VS Code with Python Debugger v2024.8.0

## Actual behavior
Try debugging the following code:
```python
import matplotlib.pyplot as plt

plt.plot([1,2,3,3])
plt.show()
pass
```

If the break point is at `pass`, where Debugger skipped entering the interactive mode, the figure is shown normally.
But if there are break points earlier, e.g., add a break point at `plt.plot...`, then debug and go though until we stop at `pass`,
the figure won't be shown, with a wayland icon/window showing in the taskbar, but an invisible figure window.

Turing on interactive mode manually makes the plot invisible too (set break point at `pass` and debug):
```python
import matplotlib.pyplot as plt

plt.ion()
plt.plot([1,2,3,3])
plt.show()
pass
```

## Expected  behavior
We expect the interactive plot window be shown as a normal window and response to resize/maximize etc.

## Steps to reproduce:
1. Create an environment with matplotlib=3.9: `mamba create -n 'test' matplotlib=3.9 qt-wayland`
2. Enabling wayland support via setting `QT_QPA_PLATFORM=wayland;xcb` in `/etc/environment` or `launch.json`
3. launch VS code and try debugging the previous code snippets
4. You may see the behavior described earlier, with visible figure window when interactive mode is off, or misbehavior when interactive mode is on.

## Additional information:
1. If we down-grade matplotlib to 3.8, this issue do not exist. (see #1606)
2. Might be the same issue as #1623, don't know if is related to #858 et al.,
3. Don't know if this is wayland specific, or also affect X11 (not tested). But note that under wayland, there exists a warning: `qt.qpa.wayland: Wayland does not support QWindow::requestActivate()` when `plt.show()`.

rchiodo:
There's specific code in pydevd (the base of debugpy) that likely is out of date with matplotlib: https://github.com/fabioz/PyDev.Debugger/blob/54f8db469c2d8c31635db7cc59a10bc8dab76057/pydevd.py#L695

mariuswallraff:
[...]

PR #1811, opened 2025-01-15
Fix matplotlib backend computation

rchiodo:
Addresses https://github.com/microsoft/debugpy/issues/1633

Matplotlib in 3.9 changed how they identify their backend to not be case sensitive anymore. They were always returning a lower case backend name. 

Our internal code relies on the casing to find the appropriate gui. I modified the code to turn the non cased version into the cased version so the rest of our code works again.

Tested with Matplotlib 3.10 and 3.8

While running pydevd tests also noticed a problem with the fix I made for just my code. Fixed that fix as well so that all the pydevd tests pass again and the just my code fix still applies.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
@@ -624,10 +624,10 @@ def _get_func_code_info(code_obj, frame_or_depth) -> FuncCodeInfo:
             frame = frame_or_depth
         assert frame.f_code is code_obj
 
-    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, py_db.get_use_libraries_filter())
+    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, True)
 
     if py_db.is_files_filter_enabled:
-        func_code_info.always_filtered_out = func_code_info.filtered_out_force_checked
+        func_code_info.always_filtered_out = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, False)
         if func_code_info.always_filtered_out:
             _code_to_func_code_info_cache[code_obj] = func_code_info
             return func_code_info
@@ -1163,7 +1163,7 @@ def _return_event(code, instruction, retval):
             or (
                 info.pydev_step_cmd == CMD_STEP_INTO_MY_CODE
                 and frame.f_back is not None
-                and not py_db.apply_files_filter(frame.f_back, frame.f_back.f_code.co_filename, py_db.get_use_libraries_filter())
+                and not py_db.apply_files_filter(frame.f_back, frame.f_back.f_code.co_filename, True)
             )
         ):
             _show_return_values(frame, retval)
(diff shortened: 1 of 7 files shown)


src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -624,10 +624,10 @@ def _get_func_code_info(code_obj, frame_or_depth) -> FuncCodeInfo:
             frame = frame_or_depth
         assert frame.f_code is code_obj
 
-    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, py_db.get_use_libraries_filter())
+    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, True)
```
rchiodo, line 627: This needs to be True as this flag is used elsewhere. However `always_filtered_out` should be determined by whether or not the file is filtered (which it isn't when justMyCode = false)

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -1163,7 +1163,7 @@ def _return_event(code, instruction, retval):
             or (
                 info.pydev_step_cmd == CMD_STEP_INTO_MY_CODE
                 and frame.f_back is not None
-                and not py_db.apply_files_filter(frame.f_back, frame.f_back.f_code.co_filename, py_db.get_use_libraries_filter())
+                and not py_db.apply_files_filter(frame.f_back, frame.f_back.f_code.co_filename, True)
```
rchiodo, line 1166: This needed to go back to true. My original change was incorrect.

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -624,10 +624,10 @@ def _get_func_code_info(code_obj, frame_or_depth) -> FuncCodeInfo:
             frame = frame_or_depth
         assert frame.f_code is code_obj
 
-    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, py_db.get_use_libraries_filter())
+    func_code_info.filtered_out_force_checked = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, True)
 
     if py_db.is_files_filter_enabled:
-        func_code_info.always_filtered_out = func_code_info.filtered_out_force_checked
+        func_code_info.always_filtered_out = py_db.apply_files_filter(frame, func_code_info.abs_path_filename, False)
```
rchiodo, line 630: always_filtered_out needs to recheck that the file is actually always filtered out. justMyCode would have to be true or it would have to match the exclusion filters if it's not in the user's project.

src/debugpy/_vendored/pydevd/pydev_ipython/matplotlibtools.py
```diff
@@ -47,9 +62,14 @@ def find_gui_and_backend():
     matplotlib = sys.modules["matplotlib"]
     # WARNING: this assumes matplotlib 1.1 or newer!!
     backend = matplotlib.rcParams["backend"]
+
+    # Translate to the real case as in 3.9 the case was forced to lowercase
+    # but our internal mapping is in the original case.
+    realcase_backend = lowercase_convert.get(backend, backend)
```
rchiodo, line 68: This is what fixes the problem with matplotlib 3.9 and above.

Issue #1783, opened 2024-12-15
Remote attach debug server not dying upon killing the attached process


zbs:
Type: <b>Bug</b>

# Behaviour

Extremely frequently the VS Code remote attach debugger will not die when I kill the `python -m debugpy ...` process that is attaching to the port 5678. Until I restart the IDE I cannot start debugging again, because otherwise it will complain that "Python debugger is already debugging another instance". My colleaques haven experiened this issue as well, and it seems to be a recent regression (last couple of weeks).

As I wrote this, I cannot click the "Create on GitHub" button because VSCode is seeming unresponsive. I resolved by restarting the IDE.

## Steps to reproduce:

1. Use the below config for debugger
2. Run `python -m debugpy --wait-for-client --listen 5678 /path/to/module.py`
3. Run debugger
4. Either detach debugger or CTRL-C the process on command line

# Diagnostic data
<details>

<summary><code>launch.json</code> configuration
</summary>

<p>

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [

        {
            "name": "Python: Remote Attach",
            "type": "python",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 5678
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}",
                    "remoteRoot": "."
                }
            ],
            "justMyCode": false
        }
    ]
}

```

</p>
</details>

<details>

<summary>Output for <code>Python</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

<details>

<summary>Output for <code>Python Debugger</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python Debugger</code>)
</summary>

<p>

```
XXX
```

</p>
</details>

Extension version: 2024.15.2024121301
VS Code version: Code 1.95.3 (f1a4fb101478ce6ec82fe9627c43efbf9e98c813, 2024-11-13T14:50:04.152Z)
OS version: Windows_NT x64 10.0.22631
Modes:
[...]

PR #1859, opened 2025-03-06
Ensure Debugpy closes connection after a disconnect response

AleksMat:
Until now, the process of closing the connection between Debugpy and a client (e.g. VSCode) would be the following:

1. a client sends a disconnect request,
2. Debugpy sends back a disconnect response acknowledging the disconnect,
3. the client sends an empty line,
4. Debugpy waits for an empty line and only then closes connection ([link to code](https://github.com/microsoft/debugpy/blob/v1.8.13/src/debugpy/common/messaging.py#L171)).

The problem is that, unlike VSCode, not all clients send this empty line. Because of that, when a user shuts down the debugger, the Debugpy's process won't die. It keeps waiting for that empty line to close the connection and at the same time occupying the communication port with the client. I believe this is the reason for the bug reported in #1783.

AFAIK, sending an empty line is not a part of the DAP protocol. Therefore, Debugpy shouldn't rely on it and should instead proactively close connection once it sends a disconnect response to the client.

This PR aims to fix this.

### DIFF
--- a/src/debugpy/common/messaging.py
+++ b/src/debugpy/common/messaging.py
@@ -1223,14 +1223,6 @@ class JsonMessageChannel(object):
             yield seq
             self.stream.write_json(message)
 
-        # Close connection after Debugpy acknowledges a disconnect request.
-        if (
-            message.get("type") == "response"
-            and message.get("command") == "disconnect"
-            and message.get("success", False)
-        ):
-            self.stream.close()
-
     def send_request(self, command, arguments=None, on_before_send=None):
         """Sends a new request, and returns the OutgoingRequest object for it.
 
--- a/src/debugpy/common/messaging.py
+++ b/src/debugpy/common/messaging.py
@@ -1223,6 +1223,14 @@ class JsonMessageChannel(object):
             yield seq
             self.stream.write_json(message)
 
+        # Close connection after Debugpy acknowledges a disconnect request.
+        if (
+            message.get("type") == "response"
+            and message.get("command") == "disconnect"
+            and message.get("success", False)
+        ):
+            self.stream.close()
+
     def send_request(self, command, arguments=None, on_before_send=None):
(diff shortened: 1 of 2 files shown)


AleksMat:
Note: the current implementation of the fix might be a bit naive but I'm happy to iterate on the solution with some guidance.

src/debugpy/common/messaging.py
```diff
@@ -1223,6 +1223,14 @@ def _send_message(self, message):
             yield seq
             self.stream.write_json(message)
 
+        # Close connection after Debugpy acknowledges a disconnect request.
```
rchiodo, line 1226: I think this should happen in the disconnect_request handler instead.

This code:
```python
    @message_handler
    def disconnect_request(self, request):
        # If user specifically requests to disconnect, it means that they don't want
        # debug session auto-restart kicking in.
        self.restart_requested = False

        terminate_debuggee = request("terminateDebuggee", bool, optional=True)
        if terminate_debuggee == ():
            terminate_debuggee = None
        self.session.finalize('client requested "disconnect"', terminate_debuggee)
        request.respond({})

        if self.using_stdio:
            # There's no way for the client to reconnect to this adapter once it disconnects
            # from this session, so close any remaining server connections.
            servers.stop_serving()
            log.info("{0} disconnected from stdio; closing remaining server connections.", self)
            for conn in servers.connections():
                try:
                    conn.channel.close()
                except Exception:
                    log.swallow_exception()

```

Something like so at the end:

```python
        # Close the client channel since we disconnected
        # from the client.
        try:
            self.channel.close()
        except Exception:
            log.swallow_exception(level="warning")
```

AleksMat: Thanks, this solution makes more sense. I made the change.

rchiodo (approved):

rchiodo:
Thanks for the submission, Win 3.13 failures aren't related so just going to merge.

Issue #1863, opened 2025-03-07
test_targets test seems to consistently fail on Win 3.13


rchiodo:
Recent PRs have the same errors:

![Image](https://github.com/user-attachments/assets/dd0122ca-e957-428f-959d-086188ec9b50)

All of the tests in test_targets consistently have this same sort of assertion.

PR #1864, opened 2025-03-07
Try making the test_cli tests more robust on Python 3.13

rchiodo:
I've been playing around with Copilot and it suggested these changes might fix the consistent problem on 3.13 with the test_targets failing.

Fixes https://github.com/microsoft/debugpy/issues/1863

### DIFF
--- a/tests/debugpy/server/test_cli.py
+++ b/tests/debugpy/server/test_cli.py
@@ -50,7 +50,10 @@ def cli(pyfile):
         }
 
         # Serialize the command line args and the options to stdout
-        os.write(1, pickle.dumps([sys.argv[1:], options]))
+        serialized_data = pickle.dumps([sys.argv[1:], options])
+        os.write(1, serialized_data)
+        # Ensure all data is written before process exits
+        sys.stdout.flush()
 
     def parse(args):
         log.debug("Parsing argv: {0!r}", args)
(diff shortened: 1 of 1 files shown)


rchiodo:
Wow, I guess copilot was right. They all passed :)

tests/debugpy/server/test_cli.py
```diff
@@ -50,21 +50,35 @@ def cli_parser():
...
+                    log.debug("Failed to deserialize output: {0}, Output was: {1!r}", e, output)
+                    raise
             except subprocess.CalledProcessError as exc:
-                raise pickle.loads(exc.output)
+                log.debug("Process exited with code {0}. Output: {1!r}, Error: {2!r}", 
+                         exc.returncode, exc.output, exc.stderr)
+                try:
+                    raise pickle.loads(exc.output)
+                except Exception as e:
+                    log.debug("Failed to deserialize error output: {0}, Output was: {1!r}", e, exc.output)
```
debonte, line 80: Isn't this `except` clause always going to get hit because of the `raise` within the `try`?

rchiodo: Yeah maybe it did that in order to log it? It's doing the same thing as before as it reraises it, but now it's logging it too.

Okay yeah that doesn't make any sense. I'll remove that part.

Issue #1857, opened 2025-03-05
Global variables set from "Variables" moved to locals on Step


Andrej730:
## Environment data

- debugpy version: 1.8.13
- OS and version: Windows 11
- Python version (& distribution if applicable, e.g. Anaconda): 3.12.7
- Using VS Code or Visual Studio: VS Code 1.93.1

## Actual behavior

https://github.com/user-attachments/assets/9ec3ca67-de5c-4ab7-a35b-905840013559

## Expected  behavior

To either prohibit editing global variable in that case as it's not editable or actually edit it.

## Steps to reproduce:

1. Set breakpoint at `b = 35` in the snippet below and run it.
2. Attach to Python using debugger.
3. Change `c` value in `Variables` section to `1000`.
4. Step Over (F10), global variable is moved to locals and global variable value is reset.

What I've also noticed that if on step 3 I run `globals()['c']` in Debug Console, it prints `0` and Variables are reset as we would be on the step 4.

```python
c = 0

def main():
    a = 25

    import debugpy
    debugpy.listen(5678)
    debugpy.wait_for_client()

    b = 35
    print(a + b + c)

if __name__ == "__main__":
    main()

```

rchiodo:
I believe this is by design. It sets the value of 'c' in the local frame. This code here:
https://github.com/microsoft/debugpy/blob/96de4376a5e61046d4348f3ff7ae73f076805649/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py#L522

If you want to force an eval to change a global, you have to do something like so:

```
globals()['c'] = 1000
```

rchiodo:
I agree it's confusing. Right now the logic for eval doesn't take into account the parent of the value.

We do get the 'parent' for the variable though, so maybe we can tell that it's coming from the global list. 

Reopening as an enhancement request.

rchiodo:
Here's the request in this example:

```
D+00009.609: /handling #22 request "setVariable" from Client[1]/
             Server[1] <-- {
                 "seq": 24,
                 "type": "request",
                 "command": "setVariable",
                 "arguments": {
                     "variablesReference": 4,
                     "name": "c",
                     "value": "1000"
                 }
             }
```

And you can see in the scopes request, that variables reference 4 is 'globals':

```
D+00003.891: Server[1] --> {
                 "pydevd_cmd_id": 502,
                 "seq": 52,
                 "type": "response",
                 "request_seq": 20,
                 "success": true,
[...]

PR #1869, opened 2025-03-10
Account for the scope when changing variables

rchiodo:
In this issue:
https://github.com/microsoft/debugpy/issues/1857

The user is trying to assign the global 'c' to a new value while inside of a function. Since the variable is listed under 'globals', you'd think this would just work.

Internally we don't actually assign it to globals unless the user types `globals()['c']=value`, but that doesn't really make sense from the variables panel.

I found that the scope of the variable is passed in the `setVariables` request, so we can use that to determine if the variable was originally a global or not and assign it to the correct scope.

I tested this with Pydevd's test automation and by using VS code.

Addresses https://github.com/microsoft/debugpy/issues/1857

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -934,7 +934,7 @@ def internal_change_variable_json(py_db, request):
         )
         return
 
-    child_var = variable.change_variable(arguments.name, arguments.value, py_db, fmt=fmt)
+    child_var = variable.change_variable(arguments.name, arguments.value, py_db, fmt=fmt, scope=scope)
 
     if child_var is None:
         _write_variable_response(py_db, request, value="", success=False, message="Unable to change: %s." % (arguments.name,))
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
@@ -199,7 +199,7 @@ class PluginManager(object):
 
         return None
 
-    def change_variable(self, frame, attr, expression):
+    def change_variable(self, frame, attr, expression, scope=None):
         for plugin in self.active_plugins:
             ret = plugin.change_variable(frame, attr, expression, self.EMPTY_SENTINEL)
             if ret is not self.EMPTY_SENTINEL:
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
@@ -201,7 +201,7 @@ class PluginManager(object):
 
     def change_variable(self, frame, attr, expression, scope=None):
         for plugin in self.active_plugins:
-            ret = plugin.change_variable(frame, attr, expression, self.EMPTY_SENTINEL)
+            ret = plugin.change_variable(frame, attr, expression, self.EMPTY_SENTINEL, scope)
             if ret is not self.EMPTY_SENTINEL:
                 return ret
 
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
@@ -200,7 +200,7 @@ class _ObjectVariable(_AbstractVariable):
 
         return children_variables
 
-    def change_variable(self, name, value, py_db, fmt=None):
+    def change_variable(self, name, value, py_db, fmt=None, scope: ScopeRequest | None=None):
         children_variable = self.get_child_variable_named(name)
         if children_variable is None:
             return None
@@ -257,7 +257,6 @@ class _FrameVariable(_AbstractVariable):
 
     def change_variable(self, name, value, py_db, fmt=None, scope: ScopeRequest | None=None):
         frame = self.frame
-
         pydevd_vars.change_attr_expression(frame, name, value, py_db, scope=scope)
         return self.get_child_variable_named(name, fmt=fmt, scope=scope)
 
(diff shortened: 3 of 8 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
```diff
@@ -934,7 +934,7 @@ def internal_change_variable_json(py_db, request):
         )
         return
 
-    child_var = variable.change_variable(arguments.name, arguments.value, py_db, fmt=fmt)
+    child_var = variable.change_variable(arguments.name, arguments.value, py_db, fmt=fmt, scope=scope)
```
rchiodo, line 937: This is where the setVariable request is handled. I added the scope to the change_variable

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
```diff
@@ -199,7 +199,7 @@ def exception_break(self, py_db, frame, thread, arg, is_unwind=False):
 
         return None
 
-    def change_variable(self, frame, attr, expression):
+    def change_variable(self, frame, attr, expression, scope=None):
         for plugin in self.active_plugins:
             ret = plugin.change_variable(frame, attr, expression, self.EMPTY_SENTINEL)
```
debonte, line 204: Should the `scope` be passed along to the plugins within the loop?

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
```diff
@@ -200,7 +200,7 @@ def get_children_variables(self, fmt=None, scope=None):
 
         return children_variables
 
-    def change_variable(self, name, value, py_db, fmt=None):
+    def change_variable(self, name, value, py_db, fmt=None, scope: Union[ScopeRequest, None]=None):
```
debonte, line 203: Can you use `Optional[Foo]` instead of `Union[Foo, None]`? https://docs.python.org/3/library/typing.html#typing.Optional

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
```diff
@@ -595,11 +596,15 @@ def method():
         del frame
 
 
-def change_attr_expression(frame, attr, expression, dbg, value=SENTINEL_VALUE):
+def change_attr_expression(frame, attr, expression, dbg, value=SENTINEL_VALUE, /, scope: Union[ScopeRequest, None]=None):
     """Changes some attribute in a given frame."""
     if frame is None:
         return
 
+    if scope is not None:
```
debonte, line 604: I would've expected `if scope:` rather than `if scope is not None:`. Is that just the convention in debugpy?

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
```diff
@@ -595,11 +596,15 @@ def method():
...
+        assert isinstance(scope, ScopeRequest)
```
debonte, line 605: Why is the `assert` needed?

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
```diff
@@ -608,13 +613,15 @@ def change_attr_expression(frame, attr, expression, dbg, value=SENTINEL_VALUE):
             if result is not dbg.plugin.EMPTY_SENTINEL:
                 return result
 
-        if attr[:7] == "Globals":
-            attr = attr[8:]
+        if attr[:7] == "Globals" or scope == "globals":
+            attr = attr[8:] if attr.startswith("Globals") else attr
```
debonte, line 617: Given that the `attr[8:]` range starts at 8, is there one more required character in the `"Globals"` prefix that should be added to the `startswith` check?

src/debugpy/_vendored/pydevd/pydevd_plugins/django_debug.py
```diff
@@ -427,7 +427,7 @@ def __init__(self, frame, original_filename, lineno, f_locals):
         self.f_trace = None
 
 
-def change_variable(frame, attr, expression, default):
+def change_variable(frame, attr, expression, default, scope=None):
     if isinstance(frame, DjangoTemplateFrame):
         result = eval(expression, frame.f_globals, frame.f_locals)
         frame._change_variable(attr, result)
```
debonte, line 433: I don't know what these plug-ins are doing, but should they be passing on the `scope` value?

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_plugin_utils.py
rchiodo: Yes I should probably do that. They don't use it, but makes sense

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
rchiodo: Yeah that works too. I'll change it. THere was already another spot that did that.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
rchiodo: Yeah it's used a lot

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
rchiodo: I just copied this code from elsewhere, but it protects the `scope = scope.scope` to make sure the item is a scoperequest

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
rchiodo: I don't think so as the if line above doesn't check for that. I didn't add this part so I'm not sure what the 8th character is supposed to be

src/debugpy/_vendored/pydevd/pydevd_plugins/django_debug.py
rchiodo: Yeah I suppose they should. That one is using the code I wrote.

debonte (approved):

PR #1870, opened 2025-03-12
add debugpy-adapter entrypoint

TheJJ:
so the adapter can also be launched inside a venv even though debugpy is not installed in the venv.

ref #1868

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -196,6 +196,11 @@ if __name__ == "__main__":
         has_ext_modules=lambda: True,
         cmdclass=cmds,
         # allow the user to call "debugpy" instead of "python -m debugpy"
-        entry_points={"console_scripts": ["debugpy = debugpy.server.cli:main"]},
+        entry_points={
+            "console_scripts": [
+                "debugpy = debugpy.server.cli:main",
+                "debugpy-adapter = debugpy.adapter.__main__:main",
+            ],
+        },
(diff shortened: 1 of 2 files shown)


TheJJ:

rchiodo:
Yeah I agree. I'll submit.

rchiodo (approved):

rchiodo:
Can you update the wiki with this extra supported command line?

TheJJ:
I don't think I have the permissions to edit the wiki? And should the `-m debugpy` invocations described there also be replaced by the entry-point? I'd of course leave a note that running as module still works.

TheJJ:
No, this doesn't solve the venv-usage yet, I fear, but is a step towards it. What the launcher now allows is launching despite being in a venv (where debugpy isn't installed). What remains to be done is that debugpy now is aware of the venv, I think.

One idea would be adding the venv site package path to the path, and removing system site packages if the venv config disallows system-site-packages.

What do you think?

rchiodo:
Sorry but I'm confused. I thought you were doing the opposite. Debugpy can already be run outside of a venv. It's actually how VS code itself works.

TheJJ:
Yes, that's what I'm trying to achieve, but using the adapter. How is debugpy launched in vscode so this works properly?

PR #1885, opened 2025-04-02
feat: add monkey_patch_os for posix_spawnp

lachalek:
Add monkey_patch_os for posix_spawnp. Currently only an existing monkey patch for posix_spawn, but would like to add one for posix_spawnp as well. This would allow more users to use debugpy for debugging applications w/ multiprocess architecture.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
+++ b/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
@@ -13,7 +13,6 @@ from _pydevd_bundle.pydevd_constants import (
     DebugInfoHolder,
     PYDEVD_USE_SYS_MONITORING,
     IS_PY313_OR_GREATER,
-    IS_PY38_OR_GREATER,
 )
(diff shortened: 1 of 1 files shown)


src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -13,6 +13,7 @@
     DebugInfoHolder,
     PYDEVD_USE_SYS_MONITORING,
     IS_PY313_OR_GREATER,
+    IS_PY38_OR_GREATER,
```
rchiodo, line 16: We don't actually support PY38 anymore, so this shouldn't be necessary

lachalek: Thanks! Removed check for `IS_PY38_OR_GREATER`

rchiodo (approved):

rchiodo:
Thanks for the PR, windows test failures can be ignored.

PR #1892, opened 2025-05-14
Do not write to stdout on debugging

vnmabus:
Write to stderr instead of stdout when changing Matplotlib interactivity. Otherwise the doctests will fail when debugging them.

I tested this by changing it manually on my VSCode instance 😁.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydev_ipython/matplotlibtools.py
+++ b/src/debugpy/_vendored/pydevd/pydev_ipython/matplotlibtools.py
@@ -140,11 +140,11 @@ def activate_matplotlib(enable_gui_function):
     if is_interactive:
         enable_gui_function(gui)
         if not matplotlib.is_interactive():
-            sys.stdout.write("Backend %s is interactive backend. Turning interactive mode on.\n" % backend)
+            sys.stderr.write("Backend %s is interactive backend. Turning interactive mode on.\n" % backend)
         matplotlib.interactive(True)
     else:
         if matplotlib.is_interactive():
-            sys.stdout.write("Backend %s is non-interactive backend. Turning interactive mode off.\n" % backend)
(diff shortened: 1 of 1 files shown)


rchiodo:
I think this is okay, but what is failing with writing to stdout?

src/debugpy/_vendored/pydevd/pydev_ipython/matplotlibtools.py
```diff
@@ -140,11 +140,11 @@ def activate_matplotlib(enable_gui_function):
     if is_interactive:
         enable_gui_function(gui)
         if not matplotlib.is_interactive():
-            sys.stdout.write("Backend %s is interactive backend. Turning interactive mode on.\n" % backend)
+            sys.stderr.write("Backend %s is interactive backend. Turning interactive mode on.\n" % backend)
```
rchiodo, line 143: I'm thinking this shouldn't write to either. It should probably write to the log instead. 

That would be something like this instead:

```suggestion
            pydev_log.debug("Backend %s is interactive backend. Turning interactive mode on.\n" % backend)
```

rchiodo: Of course after importing pdev_log. Maybe that causes some import problems though and it's why this code was writing to stdout.

vnmabus: I changed it to `stderr` because I saw it writing to `stderr` in other places in the same file, but logging is fine to me, as long as it does not end in `stdout`.

rchiodo: Yeah not sure why it's writing to stderr instead of log elsewhere. I believe this code was originally a port from somewhere else and maybe whomever ported it didn't bother to make it write to the log instead.

Issue #1252, opened 2023-03-28
[Feature request] Support connect to ipv6 address


Jarvan-Wang:
debugpy works well with ipv4,
now I are trying to use ipv6 , and obtain :
`Error: invalid --connect <address>: invalid port number`

I tried to modify debugpy/server/cli.py to fix address parse error:
```
def set_address(mode):
    def do(arg, it):
        if options.address is not None:
            raise ValueError("--listen and --connect are mutually exclusive")

        # It's either host:port, or just port.
        value = next(it)
        def get_host_port(address):
            parts = address.split(":")
            host = ":".join(parts[:-1])
            if host == "": 
                host = "127.0.0.1"
            port = parts[-1]
            return host, port
        host, port = get_host_port(value)
        #host, sep, port = value.partition(":")
        #if not sep:
        #    host = "127.0.0.1"
        #    port = value
        try:
            port = int(port)
        except Exception:
            port = -1
        if not (0 <= port < 2 ** 16):
            raise ValueError("invalid port number")

        options.mode = mode
        options.address = (host, port)

    return do
```

rerun and obtain:
```
  File "debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py", line 491, in start_client
    s.connect((host, port))
socket.gaierror: [Errno -9] Address family for hostname not supported
```

gattytto:
+1 I need this feature please, I don't know why it doesn't work anymore I used to debug my apps inside containers with ipv6 only scenarios some months ago but something changed in latest versions that will make debugpy only listen on 0.0.0.0 even if there's no ipv4 in the container

gattytto:
the relevant code is [here](https://github.com/microsoft/debugpy/blob/6e8e5becb251defdf4b7a878ec091ac7a91425d1/src/debugpy/common/sockets.py#L50)
```python
def _new_sock():
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.IPPROTO_TCP)

```

this needs to be extended so when address is not ipv4 but rather ipv6, it needs to be something like:
```python
def _new_sock_v6():
    sock = socket.socket(socket.AF_INET6...```

rchiodo:
If you're willing to try, we accept PRs. Sounds like you found where the problem is.

gattytto:
I don't think I'm that good of a dev but I will give it a try xD

gattytto:
[...]

PR #1896, opened 2025-05-22
IPv6 support

wh1ter4bb1t-js:
Add Support for IPv6 addresses

This PR addresses https://github.com/microsoft/debugpy/issues/1252.

This change tries to better generalize the host address in order to support both IPv4 and IPv6 address families

In this PR the debug workflows have changed in the following manner:

* When passing the full `{host}:{port}` path, the host can now utilize `[::1]` or any other IPv6 address
* The default cases will check if IPv4 or IPv6 is available:
    - if IPv4 is unavailable then the default host will be `[::1]`
    - if IPv4 is available or IPv6 is also unavailable the host will default back to `127.0.0.1`
   
#### Nuances:

* `getsockname` will always return >= 2 (2 in the case of IPv4 and 4 in the case of IPv6) so to ensure that there is no complicated branching logic we always take a [:2] slice.
* since the majority of the cases that are being updated here is in order to replace `127.0.0.1` to either `127.0.0.1` _or_  `::1` we opted to refer to the host when utilizing these values as localhost

### DIFF
--- a/src/debugpy/adapter/servers.py
+++ b/src/debugpy/adapter/servers.py
@@ -395,7 +395,7 @@ def serve(host="127.0.0.1", port=0):
     global listener
     listener = sockets.serve("Server", Connection, host, port)
     sessions.report_sockets()
-    return listener.getsockname()
+    return listener.getsockname()[:2]
 
 
 def is_serving():
@@ -475,7 +475,7 @@ def dont_wait_for_first_connection():
 
 
 def inject(pid, debugpy_args, on_output):
-    host, port = listener.getsockname()
+    host, port = listener.getsockname()[:2]
 
     cmdline = [
         sys.executable,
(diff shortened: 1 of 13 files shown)


wh1ter4bb1t-js:
@microsoft-github-policy-service agree company="Google"

src/debugpy/adapter/servers.py
```diff
@@ -395,7 +395,7 @@ def serve(host="127.0.0.1", port=0):
     global listener
     listener = sockets.serve("Server", Connection, host, port)
     sessions.report_sockets()
-    return listener.getsockname()
+    return listener.getsockname()[:2]
```
rchiodo, line 398: I suppose the socket.getsockname is well defined, but it feels a little strange to me to embed this [:2] in every spot this is called. Might it be better to have a function that wraps this and the function is called everywhere?

rchiodo (approved):

rchiodo:
Looks good. Thanks for the submission.

src/debugpy/adapter/servers.py
wh1ter4bb1t-js: sg. Thanks

PR #1915, opened 2025-06-24
dap: Ensure cwd is present in runInTerminal request

osiewicz:
Per https://microsoft.github.io/debug-adapter-protocol/specification, `runInTerminalRequestArguments.cwd` is a mandatory property (although it can be left empty).

### DIFF
--- a/src/debugpy/adapter/launchers.py
+++ b/src/debugpy/adapter/launchers.py
@@ -148,6 +148,7 @@ def spawn_debuggee(
                 "title": console_title,
                 "args": cmdline,
                 "env": env,
+                "cwd": ""
             }
             if cwd is not None:
                 request_args["cwd"] = cwd


rchiodo:
Is this breaking something else? I'd be wary of making it '' or setting it at all given that VS code works with it without the parameter.

osiewicz:
Yes, for one we've had to work around it in Zed in our deserialization code for that request type.
Edit: At a cursory glance, Helix might break with unfilled CWD as well:
https://github.com/helix-editor/helix/blob/master/helix-dap/src/types.rs#L756
Haven't tested that though.

rchiodo:
I think VS code is using it here:
https://github.com/microsoft/vscode/blob/d8ffc540d19fd4f936c99dd1cc38a64d3a28a079/src/vs/workbench/api/node/extHostDebugService.ts#L103, so when it's undefined it may have different behavior than if it's just empty. It'd need to be tested.

rchiodo:
It looks like it makes the tests fail at least:

```
D+00000.080: Client[1] --> {
                 "seq": 3,
                 "type": "response",
                 "request_seq": 7,
                 "success": false,
                 "command": "runInTerminal",
                 "message": "[Errno 2] No such file or directory: ''"
             }
```

So the tests would have to be updated to ignore an empty cwd.

PR #1917, opened 2025-06-26
Fix attach to process on arm64 Mac.

ConradIrwin:
Before this change it was not possible to attach to a process on arm64 mac.

The primary issue was that we weren't building the attach.dylib for all targets; but even once we did that we had to ensure that we were exiting successfully after injecting into the process.

We pulled in the compile changes from https://github.com/JetBrains/intellij-community/commit/2e7109f6fe37cf3ba74778b72bfbe9ddeb8ca4b4

Co-authored-by: @artemmukhin <Artem.Mukhin@jetbrains.com>
Co-authored-by: Cole Miller <cole@zed.dev>

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/compile_mac.sh
+++ b/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/compile_mac.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
-set -euxo pipefail
+set -e
+SRC="$(dirname "$0")/.."
 
-clang++ -fPIC -D_REENTRANT -std=c++11 -arch arm64 -c -o attach_arm64.o attach.cpp
-clang++ -dynamiclib -nostartfiles -arch arm64 -o attach_arm64.dylib attach_arm64.o -lc
-rm attach_arm64.o
+g++ -fPIC -D_REENTRANT -std=c++11 -D_FORTIFY_SOURCE=2 -arch arm64 -c -o "$SRC/attach_arm64.o" "$SRC/linux_and_mac/attach.cpp"
+g++ -dynamiclib -nostartfiles -arch arm64 -o "$SRC/attach_arm64.dylib" "$SRC/attach_arm64.o" -lc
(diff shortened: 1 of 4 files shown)


ConradIrwin:
@microsoft-github-policy-service agree company="Zed Industries"

src/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/compile_mac.sh
```diff
@@ -1,4 +1,14 @@
-set -e
-SRC="$(dirname "$0")/.."
-g++ -fPIC -D_REENTRANT -std=c++11 -D_FORTIFY_SOURCE=2 -arch x86_64 -c $SRC/linux_and_mac/attach.cpp -o $SRC/attach_x86_64.o
-g++ -dynamiclib -nostartfiles -arch x86_64 -lc $SRC/attach_x86_64.o -o $SRC/attach_x86_64.dylib
+#!/bin/bash
+set -euxo pipefail
+
+clang++ -fPIC -D_REENTRANT -std=c++11 -arch arm64 -c -o attach_arm64.o attach.cpp
```
rchiodo, line 4: I assume clang supports D_FORTIFY_SOURCE?. We have internal requirements on this flag being set or Microsoft security won't let us ship this binary. Can you try adding it back?

rchiodo: And why use clang instead of g++?

ConradIrwin: The build script was pulled from https://github.com/JetBrains/intellij-community/commit/2e7109f6fe37cf3ba74778b72bfbe9ddeb8ca4b4. I've pushed up a new version that is more like the original.

rchiodo (approved):

rchiodo:
Thanks for the submission :). Should ship in our next release. Sometime before 3.14 goes GA I think.

ConradIrwin:
Great, thank you for the super fast turnaround.

Issue #1918, opened 2025-07-03
New child process spawned with debugpy --connect not receiving breakpoints


jborean93:
## Environment data

- debugpy version:  1.8.14
- OS and version: Fedora 42
- Python version (& distribution if applicable, e.g. Anaconda): CPython 3.11.13
- Using VS Code or Visual Studio: VS Code

## Actual behavior

Due to a complicated debugging environment I am trying to setup a launch configuration which spawns a Python process which in turn setups up an interactive terminal where new processes spawned in that terminal attaches to the VSCode client running the original launch request. For example I have this launch configuration

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Python Debugger: Current File",
            "type": "debugpy",
            "request": "launch",
            "program": "debugpy_test.py",
            "console": "integratedTerminal",
            "logToFile": true
        },
    ]
}
```

`debugpy_test.py` is the following:

```python
import os
import subprocess

from debugpy.server import cli as debugpy_cli

def main() -> None:
    host, addr = debugpy_cli.options.address
    access_token = debugpy_cli.options.adapter_access_token

    with open('child_test.py', 'w') as f:
        f.write("import os\nimport debugpy\n\nprint(f'PID={os.getpid()} - debugpy.is_client_connected={debugpy.is_client_connected()}')\na = 1\nb = ''")

    print(f"Launched PID {os.getpid()}\nSet breakpoint on line 5 and run\npython -m debugpy --connect {host}:{addr} --adapter-access-token {access_token} --log-to-stderr --wait-for-client child_test.py")
    subprocess.run(
        ['bash'],
        env=os.environ | {
            'DEBUGPY_RUNNING': 'false',
            'PYDEVD_DISABLE_FILE_VALIDATION': '1',
        })


if __name__ == '__main__':
    main()
```

When running it will print out a message like the following

> Launched PID 103924
> Set breakpoint on line 5 and run
> python -m debugpy --connect 127.0.0.1:50865 --adapter-access-token ... --log-to-stderr --wait-for-client child_test.py

The `child_test.py` is created by `debugpy_test.py` with the following content

```python
import os
import debugpy

print(f'PID={os.getpid()} - debugpy.is_client_connected={debugpy.is_client_connected()}')
a = 1
b = ''
```

[...]

Issue #1876, opened 2025-03-19
Step Over, Step Into, not working when debugging Subprocesses


maor-outrival:
When debugging a process in python, the program halts on breakpoints yet only the Continue button works. Pressing Step Over or Step into just Continues to the next breakpoint.

I have tried many launch.json configs and the only thing that resolved my issue was downgrading to v2024.10.0.

here is my config just in case:

        {
            "name": "Python Debugger: Current File",
            "type": "debugpy",
            "request": "launch",
            "program": "app/main.py",
            "console": "integratedTerminal",
            "env": {
                "PYTHONPATH": "${workspaceFolder}"
            },
            "justMyCode": true,
            "subProcess": true,
            "steppingResumesAllThreads": false
        },

SobhanMP:
It's broken, see the following 

```python

import time
import multiprocessing as mp

def f():
    for i in range(10):
        time.sleep(1)
        print(i)

if __name__ == "__main__":
    p = mp.Process(target=f)
    p.start()
    p.join()

```

![Image](https://github.com/user-attachments/assets/e4806ec7-e1f0-4f75-a76a-ca208fa84328)

rchiodo:
Seems to work fine for me. Although I'm on windows. 

https://github.com/user-attachments/assets/f6b6eab0-1617-4d60-b605-08c77e955e3a

rchiodo:
I can reproduce on Linux though.

rmpIGN:
I'm facing the same issue. It appeared when I upgraded Ubuntu from 22.04 to 24.04, but I can't be sure whether it's related.

Fortunately, I found a temporary fix: downgrading ms-code.debugpy to version 2024.10.0.

rchiodo:
Sounds like this is related to sys monitoring support in Debugpy, which we switched to after 2024.10. Additionally, if it works with 3.11 or earlier (with the latest debugpy), that would also mean it's a sys monitoring issue.

jborean93:
Can confirm that downgrading to Python 3.11 works around this issue and 3.12+ is affected. Happy to test anything out or help in whatever way I can to try and fix this issue. Would love to be able to use the latest Python version on my projects that have subprocesses.

jborean93:
I've started digging into this a big more and I think I can see the cause behind it but not 100% sure. The underlying issues seems to be when the `fork` subprocess method is used (default for `multiprocessing.Process` on Linux) and not `spawn`.

[...]

PR #1920, opened 2025-07-03
Add parent-session-pid argument

jborean93:
Add the ability to specify the parent process id when connecting a new DAP server to the client. This value is used instead of the actual process' parent id so that it can be associated with a specific debug session even if it hasn't been spawned directly by that parent process.

Fixes: https://github.com/microsoft/debugpy/issues/1918

I can update the wiki pages with the new argument/kwarg once this is merged.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -2947,6 +2947,7 @@ def settrace(
     client_access_token=None,
     notify_stdin=True,
     protocol=None,
+    ppid=0,
     **kwargs,
 ):
     """Sets the tracing function with the pydev debug function and initializes needed facilities.
@@ -3006,6 +3007,11 @@ def settrace(
         When using in Eclipse the protocol should not be passed, but when used in VSCode
         or some other IDE/editor that accepts the Debug Adapter Protocol then 'dap' should
         be passed.
+
+    :param ppid:
+        Override the parent process id (PPID) for the current debugging session. This PPID is
+        reported to the debug client (IDE) and can be used to act like a child process of an
+        existing debugged process without being a child process.
     """
     if protocol and protocol.lower() == "dap":
         pydevd_defaults.PydevdCustomization.DEFAULT_PROTOCOL = pydevd_constants.HTTP_JSON_PROTOCOL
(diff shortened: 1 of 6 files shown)


src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -3088,6 +3096,7 @@ def _locked_settrace(
                 "port": int(port),
                 "multiprocess": patch_multiprocessing,
                 "skip-notify-stdin": not notify_stdin,
+                pydevd_constants.ARGUMENT_PPID: ppid,
```
rchiodo, line 3099: Shouldn't this only be set if ppid is not zero?

rchiodo (approved):

jborean93:
I've also opened a PR on PyDevD upstream with the changes I've made in the vendored copy https://github.com/fabioz/PyDev.Debugger/pull/309.

rchiodo:
Thanks for the PR.

rchiodo:
A test would be nice for this option, but it might be hard to get the intermediate process to launch debugpy in the test.

This file would be the likely spot for a test:
tests\debugpy\server\test_cli.py

jborean93:
I'll try and add one, I'm not going to lie the workflow for this scenario is complex but I'll see how I go :)

jborean93:
I've added a test for this scenario that I believe covers what this is trying to solve. The `wait_for_next_subprocess` call verifies that the Python grandchild process is associated with the parent's debug session through the `--parent-session-pid` argument. It spawns the grandchild process through a shell ensuring that it doesn't inherit the session just by being the child of the launch Python process.

rchiodo (approved):
Thanks for the PR. LGTM

rchiodo:
I'm hoping to release before 3.14 goes GA. We still need to do some more testing/work around 3.14, but waiting for it to settle down a bit.

jborean93:
Appreciate the guidance and review. Hoping to solve this one today or tomorrow as well https://github.com/microsoft/debugpy/issues/1876#issuecomment-3043654636. We are about to bump our minimum dep to 3.12+ so I can no longer rely on using 3.11 to debug our forked worker properly.

jborean93:
Sorry I know I said I was going to update the wiki with the new CLI arg and API kwarg but it looks like you need to be a maintainer for this repo to make changes there.

rchiodo:
> Sorry I know I said I was going to update the wiki with the new CLI arg and API kwarg but it looks like you need to be a maintainer for this repo to make changes there.

Oh thanks for checking. I can do that then.

PR #1921, opened 2025-07-08
Fix debugger stepping actions in forked process

jborean93:
Fix the debugger stepping state when debugging a process that has been forked from the main process. The new sys.monitoring mechanism didn't fully clear the thread local storage after a fork leading to a state where the forked child process tracked the wrong thread information and was never updated on the latest continue action.

Fixes: https://github.com/microsoft/debugpy/issues/1876

I'm not sure what the common practice is for updating the Cython/c files. The diff is pretty large but that's mostly due to line and auto formatting changes that my Cython seems to do. Happy to take them out and let you update those files if you prefer.


rchiodo (approved):

jborean93:
Not sure why that test failed, I'm assuming it's just a bit flaky. I've pushed a commit that adds a test for this scenario as well.

jborean93:
I'll push one more commit to hopefully reduce the diff, looks like Cython's auto generated .c files have trailing spaces so I'll temporarily stop my editor from removing them to avoid troubles in the future.

jborean93:
PR also opened to fix upstream https://github.com/fabioz/PyDev.Debugger/pull/310.

rchiodo:
Thanks for the fix @jborean93

PR #1928, opened 2025-07-15
Expose CLI Options through public API

jborean93:
Expose a public API that can retrieve the processed CLI options for the current process launched through the debugpy CLI. This enables code to be able to retrieve options like the port and adapter access token to be used for launching their own child process' that can be debugged.

### DIFF
--- a/tests/debugpy/test_cli_args.py
+++ b/tests/debugpy/test_cli_args.py
@@ -15,13 +15,14 @@ def test_cli_options_with_no_debugger():
 def test_cli_options_under_file_connect(pyfile, target, run):
     @pyfile
     def code_to_debug():
+        import dataclasses
         import debugpy
 
         import debuggee
         from debuggee import backchannel
 
         debuggee.setup()
-        backchannel.send(debugpy.get_cli_options())
+        backchannel.send(dataclasses.asdict(debugpy.get_cli_options()))
 
     with debug.Session() as session:
         backchannel = session.open_backchannel()


rchiodo (approved):

jborean93:
I'll look at the CI failures and try and fix them up but I just wanted to float an alternative idea to see if you prefer that.

Instead of exposing the CLI options like this an alternative would be to extract the debugging information at runtime from pydevd instead. That way it would also work for debugging sessions started with something like

```python
# src/debugpy/server/api.py

@dataclasses.dataclass(frozen=True)
class DebugInfo:
    client: str
    port: int
    access_token: str | None


def get_debug_info():
    ensure_logging()
    log.debug("get_debug_info()")

    info = pydevd.SetupHolder.setup

    return DebugInfo(
        client=info['client']
        port=info['port']
        access_token=info['access-token']
    )
```

The advantages here is that it only exposes what you want to expose from pydevd and works for both a session started from the CLI as well as through the API. The downside is you are potentially at the mercy of exposing how pydevd works.

rchiodo:
> The advantages here is that it only exposes what you want to expose from pydevd and works for both a session started from the CLI as well as through the API. The downside is you are potentially at the mercy of exposing how pydevd works.

I think I prefer your original idea. I'd guess that somebody starting using the API would know what parameters they passed to it (so they wouldn't need this api?) And not super happy about pydevd knowledge bleeding upwards.

jborean93:
Thanks, looking forward to the next release!

_Edit: just noticed 1.8.15 was just pushed out, will have to wait for the next one._

PR #1939, opened 2025-08-05
pydevd: Fix up prefix of attach shared library for Windows

osiewicz:
Follow-up to #1917, which changed the prefix for Windows. The crux of that contribution was about enabling attaching on Sillicon Mac (in fact, it came from my colleagues at Zed Industries). This however broke .dll lookup per https://github.com/zed-industries/zed/pull/35640#issuecomment-3155624377

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_tracing.py
+++ b/src/debugpy/_vendored/pydevd/pydevd_tracing.py
@@ -250,7 +250,9 @@ def get_python_helper_lib_filename():
         else:
             suffix = suffix_32
 
-        if IS_WINDOWS or IS_MAC:  # just the extension changes
+        if IS_WINDOWS:  # just the extension changes
+            prefix = "attach_"
+        elif IS_MAC:
             prefix = "attach"
             suffix = ""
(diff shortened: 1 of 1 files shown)


rchiodo (changes requested):
I'm wondering why this wasn't caught in any of the tests? Shouldn't an attach test have been failing?

osiewicz:
That I don't know, but there were some CI failures in #1917. Whether that's relevant - I'm not sure.
https://github.com/microsoft/debugpy/runs/44871322864

rchiodo:
Hmm, not sure why but either works for me locally. Oh no wait, it's only a problem on 3.11 and earlier. Sys monitoring doesn't use this method.

rchiodo (approved):
Thanks, I verified this change is required for attach in 3.11 and below on Windows.

osiewicz:
Thanks for a quick turnaround - and sorry for messing up. A member of our community told me  VSC's Python plugin is not using latest version of Debugpy just yet, so hopefully it won't have any consequences.

Issue #1943, opened 2025-08-12
`os.__file__` not always set


timrid:
## Environment data

- debugpy version: 1.8.16
- OS and version: Windows 11
- Python version: 3.13
- Using VS Code or Visual Studio: VSCode
- briefcase version: 0.3.24

## Actual behavior
I am trying to run

```
import debugpy
debugpy.listen(5678)
```

in an application that is bundled with [`briefcase`](https://github.com/beeware/briefcase) but I get the following error message:

```
Traceback (most recent call last):
  File "asyncio\events.py", line 89, in _run
  File "...\build\helloworld\windows\app\src\app_packages\toga\app.py", line 655, in _startup
    self.startup()
    ~~~~~~~~~~~~^^
  File "...\build\helloworld\windows\app\src\app\helloworld\app.py", line 218, in startup
    debugpy.listen(5678)
    ~~~~~~~~~~~~~~^^^^^^
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\public_api.py", line 44, in wrapper
    from debugpy.server import api
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\server\__init__.py", line 7, in <module>
    import debugpy._vendored.force_pydevd  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\_vendored\force_pydevd.py", line 44, in <module>
    preimport('pydevd', [
    ~~~~~~~~~^^^^^^^^^^^^
        '_pydev_bundle',
        ^^^^^^^^^^^^^^^^
    ...<5 lines>...
        'pydevd',
        ^^^^^^^^^
    ])
    ^^
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\_vendored\__init__.py", line 126, in preimport
    import_module(name)
    ~~~~~~~~~~~~~^^^^^^
  File "importlib\__init__.py", line 88, in import_module
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\_vendored\pydevd\pydevd.py", line 39, in <module>
    import pydevd_file_utils
  File "...\build\helloworld\windows\app\src\app_packages\debugpy\_vendored\pydevd\pydevd_file_utils.py", line 98, in <module>
    _library_dir = _get_library_dir()
[...]

PR #1944, opened 2025-08-12
Check if `os.__file__` is available before using it

timrid:
Fixes https://github.com/microsoft/debugpy/issues/1943

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_file_utils.py
+++ b/src/debugpy/_vendored/pydevd/pydevd_file_utils.py
@@ -89,10 +89,13 @@ def _get_library_dir():
 
     if library_dir is None or not os_path_exists(library_dir):
         if hasattr(os, "__file__"):
+            # "os" is a frozen import an thus "os.__file__" is not always set.
+            # See https://github.com/python/cpython/pull/28656
             library_dir = os.path.dirname(os.__file__)
-
-    if library_dir is None:
-        library_dir = ""  # not possible to detect library_dir
+        else:
+            # "threading" is not a frozen import an thus "threading.__file__" is always set.
+            import threading
+            library_dir = os.path.dirname(threading.__file__)
 
     return library_dir
 


timrid:

src/debugpy/_vendored/pydevd/pydevd_file_utils.py
```diff
@@ -88,7 +88,11 @@ def _get_library_dir():
                 break
 
     if library_dir is None or not os_path_exists(library_dir):
-        library_dir = os.path.dirname(os.__file__)
+        if hasattr(os, "__file__"):
+            library_dir = os.path.dirname(os.__file__)
+
+    if library_dir is None:
+        library_dir = ""  # not possible to detect library_dir
```
rchiodo, line 95: Does the debugger still work in this scenario? You might be able to do something like this:

```python
    import sysconfig
    library_dir = sysconfig.get_paths()["stdlib"])
```

rchiodo: At least locally that gives me the same path as `os.__filename__`

timrid: My understandig of `def _get_library_dir()` is, that it is a function to search for the `site-packages` directory, which is normally set via `sysconfig.get_path(purelib)`. But if that folder does not exist, as a fallback a folder called `site-packages` is searched in sys.path. And it that foulder is not found the fallback of the fallback is to use `os.__file__`. But even if `os.__file__` is a valid path, it normally does not point to the `site-packages` path.

So there are the following solutions:
1. just leave `library_dir` empty, as I did
2. use `threading.__file__` as a fallback, of the fallback, of the fallback ;) This is also set when using `briefcase` because it is not a frozen import. It points to `...\build\helloworld\windows\app\src\python313.zip\threading.pyc`. So it also would not point to the correct `site-packages` location, but at least it does not throw an exception.

Both solutions are working for me.

rchiodo: Yeah how bout we try `os.path.dirname(threading.__file__)` if library_dir is None (and threading has the `__file__` attribute)

rchiodo: That should be enough safety checks. Or at least put in a comment here saying why library_dir doesn't need to be set.

rchiodo (approved):

src/debugpy/_vendored/pydevd/pydevd_file_utils.py
timrid: I added `threading.__file__` as last fallback if `os.__file__` is not available. That should always work, as threading.__file__` is not an frozen import.

PR #1961, opened 2025-09-28
Resolve generic collections

### LINKED ISSUE #1959: Add fundamental types to resolvers
seems like the actual types that define Python behavior like `collections.abc.Mapping` are missing from the default resolvers, and instead some random collection of builtin types is there:

https://github.com/microsoft/debugpy/blob/6cbdf8767e4c88dfaedf3db7b09ce2781496fc51/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_xml.py#L42

I say “random” because `MappingProxyType`, `KeysView`, … are just as fundamental Python types that are used whenever something as basic as a class is encountered so I can’t think of a reason why they shouldn’t be there when `dict` is.

But as initially said: Python’s behavior is actually defined by `collections.abc.*` classes, so instead of having random builtin sequences like `list` in the defaults, why not just use `collections.abc.{Sequence,Mapping}` directly?

PS: #825 was closed because its title asked about this being possible. I know it’s possible and ask for it to be changed instead.

flying-sheep:
Fixes #1959

TODO:
- [ ] are the tests in `src/debugpy/_vendored/pydevd/tests_python/` run at all?
- [ ] maybe needs changes because `User{Dict,List}` also have a field

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_resolver.py
@@ -238,9 +238,9 @@ def _does_obj_repr_evaluate_to_obj(obj):
 
 
 # =======================================================================================================================
-# DictResolver
+# MappingResolver
 # =======================================================================================================================
-class DictResolver:
+class MappingResolver:
     sort_keys = not IS_PY36_OR_GREATER
 
     def resolve(self, dct, key):
@@ -452,7 +452,7 @@ class ForwardInternalResolverToObject:
         return var.resolve(attribute)
 
 
-class TupleResolver:  # to enumerate tuples and lists
+class SequenceResolver:
     def resolve(self, var, attribute):
         """
         :param var: that's the original object we're dealing with.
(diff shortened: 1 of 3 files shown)


rchiodo:
>  are the tests in src/debugpy/_vendored/pydevd/tests_python/ run at all?

They're not run as part of the normal PR process, well because changes to Pydevd should generally be made upstream first. This hasn't been happening lately though. I usually run the Pydevd tests manually myself.

rchiodo:
The directions here should outline how to run the Pydevd tests:

https://github.com/microsoft/debugpy/blob/main/CONTRIBUTING.md#updating-pydevd

flying-sheep:
so they aren’t run by your CI?

other question: are any failures here related to my change?

flying-sheep:
so they aren’t run by your CI?

other question: are any failures here related to my change?

rchiodo:
> so they aren’t run by your CI?

No they're not. You have to run them manually. 

> other question: are any failures here related to my change?

I don't believe so. There's something going on with thread_count at the moment. Not sure what, but there's been intermittent failures with that test.

PR #1989, opened 2026-01-10
Add Devcontainer Spec

rameshvarun:
I'm taking a stab at adding UNIX domain socket support to debugpy. Since I prefer to work in devcontainers, I added a spec.

Feel free to reject this MR if you don't think it's appropriate. Also let me know if I should add in all Python versions so the whole test matrix can be run.

Once the container is started, the tests can be run using `python -m tox -e py314 --develop`


rameshvarun:

Issue #1530, opened 2024-03-03
Easy Post Mortem Function


Sharpz7:
Hey Folks, feel free to close if this is not appropriate. 

I am aware this has been brought up in different forms (https://github.com/microsoft/debugpy/issues?q=is%3Aissue+post+mortem+is%3Aclosed), but looking through the solutions, I never saw anything as simple as https://docs.python.org/3/library/pdb.html#pdb.post_mortem

So I created https://github.com/SharpSet/afterthought, which works pretty much as a direct equivalent to pdb.post_mortem, and uses debugpy behind the scenes with less than 30 lines of boilerplate.

```python
import afterthought


def bombs():
    a = []
    print(a[0])


if __name__ == "__main__":
    try:
        bombs()
    except Exception as e:
        afterthought.debug(e)
```

I think it would make sense for this to exist within debugpy directly. If this is not of interest, I still would like to keep this issue here so people can see how I handled it. To the best of my knowledge I couldn't find anything this simple.

Thanks :))

nshepperd:
I've been using the following to trigger postmortem debugging when debugging code in the vscode debugger:

```py
import sys
import threading


def maybe_debugpy_postmortem(excinfo):
    """Make the debugger enter and stop at a raised exception.

    excinfo: A (type(e), e, e.__traceback__) tuple. See sys.exc_info()
    """
    exc_type, exc_value, exc_tb = excinfo

    try:
        import debugpy
        import pydevd  # type: ignore
    except ImportError:
        # If pydevd isn't available, no debugger attached; do nothing.
        return

    if not debugpy.is_client_connected():
        return

    py_db = pydevd.get_global_debugger()
    thread = threading.current_thread()
    additional_info = py_db.set_additional_thread_info(thread)
    additional_info.is_tracing += 1
    try:
        py_db.stop_on_unhandled_exception(py_db, thread, additional_info, excinfo)
    finally:
        additional_info.is_tracing -= 1


def debugpy_pm_tb():
    maybe_debugpy_postmortem(sys.exc_info())
```

[...]

PR #1996, opened 2026-02-01
Add trigger_exception_handler() for post-mortem debugging of caught exceptions

nshepperd:
Adds a convenient trigger_exception_handler() function (previously named post_mortem(); renamed per review discussion) that can be called to immediately trigger the graphical postmortem debugger given any exc_info structure. We need to suspend sys.monitoring callbacks while doing this, because the built in reentrancy protection doesn't apply when pydevd suspends directly from normal code instead of a monitoring callback. Functions were added in _pydevd_sys_monitoring to do that.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
@@ -1,6 +1,6 @@
 from _pydev_bundle import pydev_log
 from _pydevd_bundle import pydevd_import_class
-from _pydevd_bundle.pydevd_frame_utils import add_exception_to_frame, remove_exception_from_frame
+from _pydevd_bundle.pydevd_frame_utils import exception_on_frame
 from _pydev_bundle._pydev_saved_modules import threading
 
 
@@ -157,8 +157,7 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
     frames_byid = dict([(id(frame), frame) for frame in frames])
     # Attach __exception__ so conditions/expressions can reference it; always
     # detach as the thread may keep running (do_stop re-adds it while suspended).
-    add_exception_to_frame(user_frame, arg)
-    try:
+    with exception_on_frame(user_frame, arg):
         if exception_breakpoint.condition is not None:
             eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
             if not eval_result:
@@ -166,8 +165,6 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
 
         if exception_breakpoint.expression is not None:
             py_db.handle_breakpoint_expression(exception_breakpoint, additional_info, user_frame)
-    finally:
-        remove_exception_from_frame(user_frame)
 
     try:
         additional_info.pydev_message = exception_breakpoint.qname
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
@@ -1,6 +1,6 @@
 from _pydev_bundle import pydev_log
 from _pydevd_bundle import pydevd_import_class
-from _pydevd_bundle.pydevd_frame_utils import add_exception_to_frame
+from _pydevd_bundle.pydevd_frame_utils import add_exception_to_frame, remove_exception_from_frame
 from _pydev_bundle._pydev_saved_modules import threading
 
 
@@ -159,6 +159,9 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
     if exception_breakpoint.condition is not None:
         eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
         if not eval_result:
+            # The thread may keep running (e.g. via PyDB.trigger_exception_handler),
+            # so don't leave __exception__ in the frame's locals.
+            remove_exception_from_frame(user_frame)
             return
 
     if exception_breakpoint.expression is not None:
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
@@ -155,17 +155,19 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
         return
 
     frames_byid = dict([(id(frame), frame) for frame in frames])
+    # Attach __exception__ so conditions/expressions can reference it; always
+    # detach as the thread may keep running (do_stop re-adds it while suspended).
     add_exception_to_frame(user_frame, arg)
-    if exception_breakpoint.condition is not None:
-        eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
-        if not eval_result:
-            # The thread may keep running (e.g. via PyDB.trigger_exception_handler),
-            # so don't leave __exception__ in the frame's locals.
-            remove_exception_from_frame(user_frame)
-            return
+    try:
+        if exception_breakpoint.condition is not None:
+            eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
+            if not eval_result:
+                return
 
-    if exception_breakpoint.expression is not None:
-        py_db.handle_breakpoint_expression(exception_breakpoint, additional_info, user_frame)
+        if exception_breakpoint.expression is not None:
+            py_db.handle_breakpoint_expression(exception_breakpoint, additional_info, user_frame)
+    finally:
+        remove_exception_from_frame(user_frame)
 
     try:
         additional_info.pydev_message = exception_breakpoint.qname
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
@@ -43,6 +43,32 @@ def remove_exception_from_frame(frame):
         frame.f_locals.pop("__exception__", None)
 
 
+_NO_EXCEPTION = object()
+
+
+class exception_on_frame(object):
+    """Exposes '__exception__' in the frame's locals during a stop, then restores
+    whatever was there before. Restoring by assignment (rather than deleting) avoids
+    a ValueError on 3.14+, where FrameLocalsProxy forbids deleting a real local -- as
+    happens when the user's own code has a variable named '__exception__'."""
+
+    def __init__(self, frame, exception_info):
+        self.frame = frame
+        self.exception_info = exception_info
+
+    def __enter__(self):
+        self._previous = self.frame.f_locals.get("__exception__", _NO_EXCEPTION)
+        add_exception_to_frame(self.frame, self.exception_info)
+        return self
+
+    def __exit__(self, *args):
+        if self._previous is _NO_EXCEPTION:
+            remove_exception_from_frame(self.frame)
+        else:
+            self.frame.f_locals["__exception__"] = self._previous
+        self.frame = None
+
+
 FILES_WITH_IMPORT_HOOKS = ["pydev_monkey_qt.py", "pydev_import_hook.py"]
 
 
--- a/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
@@ -1866,6 +1866,51 @@ def stop_monitoring(all_threads=False):
         thread_info.trace = False
 
 
+# fmt: off
+# IFDEF CYTHON
+# cpdef bint suspend_current_thread_tracing():
+#     cdef ThreadInfo thread_info
+# ELSE
+def suspend_current_thread_tracing():
+# ENDIF
+# fmt: on
+    """
+    Suspends tracing for the current thread and returns the previous state,
+    to be restored with resume_current_thread_tracing().
+    """
+    try:
+        thread_info = _thread_local_info.thread_info
+    except:
+        # Create the ThreadInfo if missing; monitoring events create it with
+        # tracing enabled by default, which would defeat the suspension.
+        thread_info = _get_thread_info(True, 1)
+        if thread_info is None:
+            return False
+    previous_state = thread_info.trace
+    thread_info.trace = False
+    return previous_state
+
+
+# fmt: off
+# IFDEF CYTHON
+# cpdef resume_current_thread_tracing():
+#     cdef ThreadInfo thread_info
+# ELSE
+def resume_current_thread_tracing():
+# ENDIF
+# fmt: on
+    """
+    Resumes tracing for the current thread.
+    """
+    try:
+        thread_info = _thread_local_info.thread_info
+    except:
+        thread_info = _get_thread_info(True, 1)
+        if thread_info is None:
+            return
+    thread_info.trace = True
+
+
 def update_monitor_events(suspend_requested: Optional[bool]=None) -> None:
     """
     This should be called when breakpoints change.
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -2452,9 +2452,6 @@ class PyDB(object):
                 pydev_log.debug("trigger_exception_handler: no user frame found in traceback")
                 return
 
-            # do_stop_on_unhandled_exception doesn't use frames_byid.
-            frames_byid = {id(user_frame): user_frame}
-
         thread = threading.current_thread()
         additional_info = self.set_additional_thread_info(thread)
 
@@ -2472,7 +2469,8 @@ class PyDB(object):
                 if as_uncaught:
                     stop_on_unhandled_exception(self, thread, additional_info, excinfo)
                 else:
-                    self.do_stop_on_unhandled_exception(thread, user_frame, frames_byid, excinfo)
+                    # frames_byid is unused; excinfo's traceback keeps the frames alive.
+                    self.do_stop_on_unhandled_exception(thread, user_frame, None, excinfo)
             finally:
                 additional_info.is_tracing -= 1
         finally:
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -2449,7 +2449,7 @@ class PyDB(object):
                 tb = tb.tb_next
 
             if user_frame is None:
-                pydev_log.debug("trigger_exception_handler: no user frame found in traceback")
+                pydev_log.warn("trigger_exception_handler: no user frame found in traceback")
                 return
 
         thread = threading.current_thread()
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -2443,7 +2443,7 @@ class PyDB(object):
             # Innermost frame that isn't debugger-internal.
             user_frame = None
             while tb is not None:
-                if self.get_file_type(tb.tb_frame) is None:
+                if self.get_file_type(tb.tb_frame) != self.PYDEV_FILE:
                     user_frame = tb.tb_frame
                 tb = tb.tb_next
 
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -2427,6 +2427,58 @@ class PyDB(object):
             remove_exception_from_frame(frame)
             frame = None
 
+    def trigger_exception_handler(self, excinfo, as_uncaught=True):
+        """
+        Triggers post-mortem debugging as if handling an uncaught exception.
+
+        If as_uncaught is True (default), applies the client's uncaught-exception breakpoint
+        filters (no-op if they exclude this exception); if False, stops directly on the deepest
+        non-internal frame of the traceback (no-op if there is none). The traceback has already
+        been unwound, so stepping from the stop behaves like pdb.post_mortem.
+
+        :param excinfo: A tuple of (exc_type, exc_value, exc_traceback).
+        """
+        if not as_uncaught:
+            tb = excinfo[2]
+
+            # Innermost frame that isn't debugger-internal.
+            user_frame = None
+            while tb is not None:
+                if self.get_file_type(tb.tb_frame) is None:
+                    user_frame = tb.tb_frame
+                tb = tb.tb_next
+
+            if user_frame is None:
+                pydev_log.debug("trigger_exception_handler: no user frame found in traceback")
+                return
+
+            # do_stop_on_unhandled_exception doesn't use frames_byid.
+            frames_byid = {id(user_frame): user_frame}
+
+        thread = threading.current_thread()
+        additional_info = self.set_additional_thread_info(thread)
+
+        # PEP 669 protects stops made from inside a monitoring callback against
+        # re-entrancy; a stop from normal user code gets no such protection, so
+        # suspend this thread's tracing until we resume. Side effect: breakpoints
+        # hit by code evaluated in the debug console during this stop won't
+        # trigger on this thread.
+        saved_sys_monitoring_trace = False
+        try:
+            if PYDEVD_USE_SYS_MONITORING:
+                saved_sys_monitoring_trace = pydevd_sys_monitoring.suspend_current_thread_tracing()
+            additional_info.is_tracing += 1
+            try:
+                if as_uncaught:
+                    stop_on_unhandled_exception(self, thread, additional_info, excinfo)
+                else:
+                    self.do_stop_on_unhandled_exception(thread, user_frame, frames_byid, excinfo)
+            finally:
+                additional_info.is_tracing -= 1
+        finally:
+            if saved_sys_monitoring_trace:
+                pydevd_sys_monitoring.resume_current_thread_tracing()
+
     def set_trace_for_frame_and_parents(self, thread_ident: Optional[int], frame, **kwargs):
         disable = kwargs.pop("disable", False)
         assert not kwargs
--- a/src/debugpy/_vendored/pydevd/pydevd.py
+++ b/src/debugpy/_vendored/pydevd/pydevd.py
@@ -99,7 +99,7 @@ from _pydevd_bundle.pydevd_defaults import PydevdCustomization  # Note: import a
 from _pydevd_bundle.pydevd_custom_frames import CustomFramesContainer, custom_frames_container_init
 from _pydevd_bundle.pydevd_dont_trace_files import DONT_TRACE, PYDEV_FILE, LIB_FILE, DONT_TRACE_DIRS
 from _pydevd_bundle.pydevd_extension_api import DebuggerEventHandler
-from _pydevd_bundle.pydevd_frame_utils import add_exception_to_frame, remove_exception_from_frame, short_stack
+from _pydevd_bundle.pydevd_frame_utils import exception_on_frame, short_stack
 from _pydevd_bundle.pydevd_net_command_factory_xml import NetCommandFactory
 from _pydevd_bundle.pydevd_trace_dispatch import (
     trace_dispatch as _trace_dispatch,
@@ -2414,17 +2414,16 @@ class PyDB(object):
     def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
         pydev_log.debug("We are stopping in unhandled exception.")
         try:
-            add_exception_to_frame(frame, arg)
-            self.send_caught_exception_stack(thread, arg, id(frame))
-            try:
-                self.set_suspend(thread, CMD_ADD_EXCEPTION_BREAK)
-                self.do_wait_suspend(thread, frame, "exception", arg, EXCEPTION_TYPE_UNHANDLED)
-            except:
-                self.send_caught_exception_stack_proceeded(thread)
+            with exception_on_frame(frame, arg):
+                self.send_caught_exception_stack(thread, arg, id(frame))
+                try:
+                    self.set_suspend(thread, CMD_ADD_EXCEPTION_BREAK)
+                    self.do_wait_suspend(thread, frame, "exception", arg, EXCEPTION_TYPE_UNHANDLED)
+                except:
+                    self.send_caught_exception_stack_proceeded(thread)
         except:
             pydev_log.exception("We've got an error while stopping in unhandled exception: %s.", arg[0])
         finally:
-            remove_exception_from_frame(frame)
             frame = None
 
     def trigger_exception_handler(self, excinfo, as_uncaught=True):
--- a/src/debugpy/public_api.py
+++ b/src/debugpy/public_api.py
@@ -212,6 +212,44 @@ def trace_this_thread(__should_trace: bool):
     """
 
 
+@_api()
+def trigger_exception_handler(
+    __excinfo: (
+        typing.Tuple[type, BaseException, typing.Any] | BaseException | None
+    ) = None,
+    as_uncaught: bool = True,
+) -> None:
+    """Stops the debugger on an already-caught exception, as if it
+    were unhandled, allowing inspection of the exception and the
+    call stack at the point of failure.
+
+    The exception may be passed as an exception instance or as a
+    (type, value, traceback) tuple; with no argument, the current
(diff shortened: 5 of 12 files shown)


nshepperd:
See https://github.com/microsoft/debugpy/issues/1530.

rchiodo:
Thanks for the submission but I think this would need some sort of automated test before we'd accept it. And maybe a section in the readme.md explaining how to use it.

nshepperd:
Actually I was wondering how to setup tests for this. I figured it out today. Hooray!

rchiodo:
I'm confused about something. pdb.post_mortem attaches the debugger to the process and starts debugging it. This won't. You have to be already attached, right? 

What would be the point of this then? Is it just so you don't have to break on exceptions in the debugger, but instead control when the break happens?

rchiodo:
I think ideally this would actually start VS code and open the debugger. Which might be doable, but would probably have to be configurable somehow.

nshepperd:
Yes, the point is so that you can eg. make a context manager or function decorator to debug exceptions that bubble past it instead of setting bespoke breakpoint, which isn't really directly possible for this. I find something like this indispensable for debugging things like api servers or unit tests (pytest) that have exception handlers that eat exceptions. With this you can just wrap your route handler or set an exception handler with conftest.py and launch the program under the debugger and everything just works.

I think previously debugpy had the 'user uncaught exceptions' breakpoint added for similar use cases, but that's much less specific and in some cases has too many false positives to be useful.

You're right this is a bit different from pdb.post_mortem() in that that actually starts the debugger. I'm not sure about starting vscode though, heh. Maybe it's possible via some sort of vscode:// url? What we could do is at least run debugpy.listen and wait for the debugger to attach? I'm not sure i would actually use that though, usually if i'd want to debug something with vscode I'm already debugging it (eg. I press the 'debug unit test' button) and the issue is just getting the breakpoint.

nshepperd:
Either way each use cases should work, i think. If you wanted to launch vscode and attach the debugger you can use existing debugpy functions for that and then call post_mortem. Or if we include that in the function you can check if debugpy.is_client_connected() before calling it if you don't want that.

rchiodo:
> With this you can just wrap your route handler or set an exception handler with conftest.py and launch the program under the debugger and everything just works.

If this is the intended use case, maybe the name is off then. It's not post_mortem, it's more like 'force_exception_handler' or something like that. I feel like 'post_mortem' is intended to be like pdb means it. It starts a debugger even when it's not already attached.

rchiodo:
Your unit tests also seem to be failing. Not sure if you can see the errors or not

nshepperd:
`trigger_exception_handler()`, maybe?

I see the test failures. Cython issue. I forgot to regenerate the cython files. Confusing because it worked locally.

rchiodo:
`trigger_exception_handler` sounds good to me :)

rchiodo (approved):

rchiodo:
LGTM. I'll get somebody else to take a look.

rchiodo:

heejaechang:

heejaechang (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -2427,6 +2427,51 @@ def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
...
+        try:
+            if as_uncaught:
+                self.stop_on_unhandled_exception(self, thread, additional_info, excinfo)
+            else:
+                self.do_stop_on_unhandled_exception(thread, user_frame, frames_byid, excinfo)
+        finally:
+            if PYDEVD_USE_SYS_MONITORING:
+                if saved_sys_monitoring_trace:
+                    pydevd_sys_monitoring.resume_current_thread_tracing()
+            additional_info.is_tracing -= 1
```
rchiodo: 📍 src/debugpy/_vendored/pydevd/pydevd.py:2455
The Skeptic verified (reproduced with a snippet: `tracing restored? False`) that `suspend_current_thread_tracing()`, `set_additional_thread_info`, and `is_tracing += 1` all run **outside** the `try`. If anything raises in that window — most realistically a `KeyboardInterrupt` during a live debug session, or a failure in `set_additional_thread_info` — the `finally` calling `resume_current_thread_tracing()` never runs, leaving `thread_info.trace = False` permanently and silently killing breakpoints on that thread for the rest of the session. Move the suspend + `is_tracing += 1` setup inside the `try` so the `finally` always restores state.

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -1866,6 +1866,56 @@ def stop_monitoring(all_threads=False):
...
+    Use resume_current_thread_tracing() to restore.
+    """
+    try:
+        thread_info = _thread_local_info.thread_info
+    except:
+        # Create the thread info if it doesn't exist yet: if tracing is not
+        # explicitly disabled here, the first monitoring event on this thread
+        # would create it with tracing enabled.
+        thread_info = _get_thread_info(True, 1)
+        if thread_info is None:
```
rchiodo, line 1887: 📍 src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py:1875
When the `except` branch freshly creates a `ThreadInfo` via `_get_thread_info(True, 1)` (tracing enabled), it still reports `previous_state = True`. The caller's `finally` then sees `saved == True` and re-enables tracing on a thread that was **not** previously traced, contradicting "restore previous state" and potentially causing unexpected breakpoint stops on background threads. When the `ThreadInfo` is created here rather than found, return `False` so the caller skips the resume. Mirror the same fix in the `.pyx` (and regenerate the `.c`).

tests/debugpy/test_trigger_exception_handler.py
```diff
@@ -0,0 +1,205 @@
...
+        exc_info = session.request("exceptionInfo", {"threadId": occ.body['threadId']})
+        assert exc_info == some.dict.containing(
+            {
+                "exceptionId": some.str.matching(r"(.+\.)?ValueError"),
+                "description": "something went wrong",
+                "breakMode": "unhandled",
+            }
+        )
+
+        session.request_continue()
```
rchiodo, line 284: 📍 tests/debugpy/test_trigger_exception_handler.py:1
The PR's core justification — suspending `sys.monitoring` to avoid reentrancy — is untested. None of the 5 tests set a breakpoint condition that re-enters monitoring during the stop, and none assert that breakpoints still fire on the same thread **after** `request_continue()`. If `suspend_current_thread_tracing()` were a no-op (or the resume-leak bugs triggered), every test would still pass. Add a test that hits a real breakpoint on the same thread after `trigger_exception_handler` returns, proving tracing was correctly restored.

src/debugpy/public_api.py
```diff
@@ -212,6 +212,33 @@ def trace_this_thread(__should_trace: bool):
...
+    Example::
+
+        try:
+            risky_operation()
+        except Exception:
+            debugpy.trigger_exception_handler()  # Uses current exception
+            raise
+    """
+
+
```
rchiodo, line 252: 📍 src/debugpy/public_api.py:215
The `as_uncaught` parameter is never documented in the public docstring or README, despite changing a non-obvious, inverted-feeling contract: `True` respects the uncaught-exception breakpoint filters (may not stop), while `False` forces post-mortem regardless. Document both `as_uncaught` semantics and note the `as_uncaught=False` + library-only-traceback case.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -2427,6 +2427,51 @@ def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
...
+            user_frame = None
+            while tb is not None:
+                frame = tb.tb_frame
+                # Skip debugger-internal frames, use last user frame
+                if self.get_file_type(frame) is None:
+                    user_frame = frame
+                frames.append(frame)
+                tb = tb.tb_next
+
+            if user_frame is None:
```
rchiodo, line 2450: 📍 src/debugpy/_vendored/pydevd/pydevd.py:2430
The `as_uncaught=False` path silently no-ops when the traceback contains only library frames (`user_frame` stays `None` → debug log + return). The README/docstring imply `as_uncaught=False` always enters post-mortem, so a caller wrapping a library-only exception gets a surprising silent non-stop. At minimum document this; consider surfacing it more loudly than a debug-level log.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/server/api.py:L396](https://github.com/microsoft/debugpy/blob/92434f46a43292154a6ff10dfff21cf849f4dca2/src/debugpy/server/api.py#L396).

📍 src/debugpy/server/api.py:366
Missing negative-path test coverage for the early returns here: no test covers "no client connected → returns gracefully" or "None arg / no current exception → does nothing". These are cheap to add and lock in the documented no-op semantics.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -2427,6 +2427,51 @@ def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
...
+        if PYDEVD_USE_SYS_MONITORING:
+            saved_sys_monitoring_trace = pydevd_sys_monitoring.suspend_current_thread_tracing()
+        thread = threading.current_thread()
+        additional_info = self.set_additional_thread_info(thread)
+        additional_info.is_tracing += 1
+
+        try:
+            if as_uncaught:
+                self.stop_on_unhandled_exception(self, thread, additional_info, excinfo)
+            else:
```
rchiodo: 📍 src/debugpy/_vendored/pydevd/pydevd.py:2466
`self.stop_on_unhandled_exception(self, thread, additional_info, excinfo)` is correct — line 765 binds the module-level function as an instance attribute, so the explicit `self` is the intended `py_db` first arg. It is NOT a bug, but it reads like a double-`self` mistake and is a maintenance trap a future reader may "fix" and break. Add a one-line comment (e.g. `# instance attr holds a module-level fn; first arg is py_db`).

tests/debugpy/test_trigger_exception_handler.py
```diff
@@ -0,0 +1,205 @@
...
+        occ = session.wait_for_next(
+            Event("stopped") | Event("terminated"),
+        )
+
+        if occ.event == "terminated":
+            pytest.fail("Debuggee exited without hitting breakpoint")
+        
+
+        exc_info = session.request("exceptionInfo", {"threadId": occ.body['threadId']})
+        assert exc_info == some.dict.containing(
```
rchiodo, line 162: 📍 tests/debugpy/test_trigger_exception_handler.py:152
The stray `print("About to call trigger_exception_handler with excinfo")` in `test_trigger_exception_handler_excinfo` is leftover debugging. Remove it.

rchiodo:
Overall this is a solid, well-tested addition. The main theme worth addressing before/after merge is robustness of the sys.monitoring tracing state restoration: the suspend + is_tracing setup happens outside the try/finally, the 'previous state' value can be misleading for freshly-created ThreadInfo, and the core reentrancy-suspension behavior is not directly exercised by a test. Also a couple of documentation gaps around as_uncaught. None are blocking.

rchiodo (approved):
Approved via Review Center.

rchiodo:
Okay sorry, I had our automated tools run over it. It found some things to look at if you can. Thanks for the work :)

rchiodo:
You also need to do this before we can accept it:

```
@microsoft-github-policy-service agree [company="{your company}"]
```

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -1866,6 +1866,51 @@ def stop_monitoring(all_threads=False):
...
+    """
+    try:
+        thread_info = _thread_local_info.thread_info
+    except:
+        # Create the ThreadInfo if missing; monitoring events create it with
+        # tracing enabled by default, which would defeat the suspension.
+        thread_info = _get_thread_info(True, 1)
+        if thread_info is None:
+            return False
+    previous_state = thread_info.trace
```
rchiodo, line 1889: When the `except` branch creates a fresh `ThreadInfo` via `_get_thread_info(True, 1)` (tracing enabled by default), the function still falls through to `return previous_state` (= True). The caller's `finally` then re-enables tracing on a thread that was never previously traced, contradicting the documented "returns the previous tracing state" contract. This path is hard to hit in the current flow (explicitly-untraced threads already own a `ThreadInfo` and take the `try` branch), so it's low severity — but the fix is trivial: `return False` from the `except` branch after creating the info. Mirror the change in `_pydevd_sys_monitoring_cython.pyx` and regenerate the `.c`.

tests/debugpy/test_trigger_exception_handler.py
```diff
@@ -0,0 +1,284 @@
...
+
+        import debugpy
+
+        def risky_operation():
+            raise ValueError("something went wrong")  # @raise
+
+        try:
+            risky_operation()
+        except ValueError:
+            debugpy.trigger_exception_handler()
```
rchiodo, line 190: `test_trigger_exception_handler_restores_tracing` is a good guard that tracing is restored after the stop, but it proves *restoration* only — a no-op `suspend_current_thread_tracing()` would still pass it, leaving the PR's central justification (suspending `sys.monitoring` to avoid PEP 669 reentrancy during the stop) unverified. Consider adding a test that re-enters monitoring *during* the stop (e.g. a conditional exception breakpoint whose condition itself hits another breakpoint) so the suppression itself is asserted.

rchiodo:
Looks good overall — approving. Two minor, non-blocking notes below (a trivial self-consistency fix in the new suspend helper, and an optional test to directly exercise the sys.monitoring suppression). Note that several first-pass findings were already addressed in the current diff (the suspend/resume setup is inside the try with saved state initialized to False, and as_uncaught is documented in the public docstring).

rchiodo (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
```diff
@@ -159,6 +159,9 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
     if exception_breakpoint.condition is not None:
         eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
         if not eval_result:
+            # The thread may keep running (e.g. via PyDB.trigger_exception_handler),
+            # so don't leave __exception__ in the frame's locals.
+            remove_exception_from_frame(user_frame)
```
rchiodo: The added `remove_exception_from_frame(user_frame)` only runs on the `if not eval_result:` path. `add_exception_to_frame` ran unconditionally, so if `handle_breakpoint_condition`/`handle_breakpoint_expression` propagates a `BaseException` (KeyboardInterrupt/SystemExit — ordinary `Exception`s are swallowed by those helpers), `__exception__` stays in the frame's locals. This is newly observable because the thread now survives the stop. Pair the attach/detach in a `try/finally` (or clean up on all early-return paths, including the `expression` branch below this hunk) so `__exception__` never lingers in a live frame.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -2427,6 +2427,58 @@ def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
...
+
+            # Innermost frame that isn't debugger-internal.
+            user_frame = None
+            while tb is not None:
+                if self.get_file_type(tb.tb_frame) is None:
+                    user_frame = tb.tb_frame
+                tb = tb.tb_next
+
+            if user_frame is None:
+                pydev_log.debug("trigger_exception_handler: no user frame found in traceback")
```
rchiodo: The comment says `do_stop_on_unhandled_exception` doesn't use `frames_byid`, yet a meaningful-looking `frames_byid = {id(user_frame): user_frame}` is still constructed and passed. This misleads future readers. Pass `{}`/`None` while keeping the comment, or drop the comment if the dict is in fact consumed.

rchiodo:
Looks good overall — solid implementation and test coverage. Left a few minor, non-blocking notes; safe to merge once you've glanced at them.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/pydevd.py
nshepperd: Rearranged the try/catch to fix this.

src/debugpy/public_api.py
nshepperd: fixed

src/debugpy/_vendored/pydevd/pydevd.py
nshepperd: padding None now. it's indeed unused.

tests/debugpy/test_trigger_exception_handler.py
```diff
@@ -0,0 +1,284 @@
...
+            pytest.fail("Debuggee exited without hitting breakpoint")
+
+        exc_info = session.request("exceptionInfo", {"threadId": occ.body["threadId"]})
+        assert exc_info == some.dict.containing(
+            {
+                "exceptionId": some.str.matching(r"(.+\.)?ValueError"),
+                "description": "something went wrong",
+                "breakMode": "unhandled",
+            }
+        )
```
rchiodo, line 48: 📍 tests/debugpy/test_trigger_exception_handler.py:40

The Rules reviewer flagged that four tests (`_basic`, `_basic_with_exception`, `_excinfo`, `_not_as_uncaught`) assert with `some.dict.containing({...})` and `some.str.matching(r"(.+\.)?ValueError")` without justification, violating `docs/pylancewiki/review/tests-no-partial-asserts.md` (partial-dict/substring matches require a comment explaining why the result cannot be made deterministic). This is a confirmed rule violation and is non-downgradeable. These `some.*` matchers are debugpy's idiomatic harness, and `exceptionId` legitimately varies across Python versions (`ValueError` vs `builtins.ValueError`) — so the minimal compliant fix is to add a short comment stating why exact equality isn't possible (and tighten to `==` on the parts of the `exceptionInfo` response that are deterministic).

[verified]

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -1866,6 +1866,51 @@ def stop_monitoring(all_threads=False):
...
+    Suspends tracing for the current thread and returns the previous state,
+    to be restored with resume_current_thread_tracing().
+    """
+    try:
+        thread_info = _thread_local_info.thread_info
+    except:
+        # Create the ThreadInfo if missing; monitoring events create it with
+        # tracing enabled by default, which would defeat the suspension.
+        thread_info = _get_thread_info(True, 1)
+        if thread_info is None:
```
rchiodo, line 1887: 📍 src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py:1875

Unresolved from the prior round (kept as warning). When the `except` branch creates a fresh `ThreadInfo` via `_get_thread_info(True, 1)` (trace defaults to `True`), the function still returns `previous_state = True`, contradicting the docstring's "returns the previous state" for a thread that had no prior state. As the Advocate/Architect traced, the naive `return False` is NOT a clean fix (it strands the thread at `trace=False` or defeats suppression), so reconcile the **docstring/contract** instead — e.g. document that a freshly-created info reports `True` by design and `False` only when the thread cannot be resolved. Mirror any wording change in `_pydevd_sys_monitoring_cython.pyx`.

[verified]

src/debugpy/server/api.py
```diff
@@ -364,3 +365,47 @@ def trace_this_thread(should_trace):
...
+        )
+
+    exctype, value, tb = excinfo
+    if exctype is None or value is None or tb is None:
+        log.debug("trigger_exception_handler() ignored - no exception info")
+        return
+
+    if not (
+        isinstance(exctype, type)
+        and isinstance(value, BaseException)
```
rchiodo, line 396: 📍 src/debugpy/server/api.py:380

Unresolved from the prior round. The identical multi-line `"excinfo must be an exception instance or a (type, value, traceback) tuple ..."` message is raised by both the tuple-shape guard and the element-type guard. When the second guard fires the tuple shape is already correct, so the message is misleading. Hoist it into a single constant, or give the element-type guard a distinct message (e.g. reporting the actual element types).

[verified]

tests/debugpy/test_trigger_exception_handler.py
```diff
@@ -0,0 +1,284 @@
...
+            expected_frames=[some.dap.frame(code_to_debug, line="bp")],
+        )
+        session.request_continue()
+
+
+def test_trigger_exception_handler_no_op_paths(pyfile, target, run):
+    """Bad input raises ValueError; calling with no current exception is a no-op."""
+
+    @pyfile
+    def code_to_debug():
```
rchiodo, line 213: 📍 tests/debugpy/test_trigger_exception_handler.py:190

Unresolved coverage gap. `test_trigger_exception_handler_restores_tracing` proves tracing is *restored* after the stop, but a no-op `suspend_current_thread_tracing()` would still pass it — leaving the PR's central justification (suppressing PEP 669 reentrancy *during* the stop) unasserted. Consider adding a test that re-enters monitoring while the stop is in progress (e.g. a conditional exception breakpoint whose condition itself hits another breakpoint) so the suppression itself is validated.

[verified]

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
```diff
@@ -1866,6 +1866,51 @@ def stop_monitoring(all_threads=False):
...
+            return False
+    previous_state = thread_info.trace
+    thread_info.trace = False
+    return previous_state
+
+
+# fmt: off
```
rchiodo, line 1894: 📍 src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py:1887
The `except` branch still creates `thread_info` via `_get_thread_info(True, 1)` (default `trace=True`) and returns `previous_state = True`, contradicting the docstring's "returns the previous state" for a thread that was never traced (Skeptic, Architect, and Rules all re-confirm this). **However**, the Advocate's caller-trace shows the literal `return False` requested last round is not a clean fix: returning `False` *before* setting `thread_info.trace = False` skips suppression during the stop (defeating the PR's PEP-669 purpose on this path), while returning `False` *after* suspending would leave the fresh thread permanently at `trace=False` (a suppression leak). The current behavior — suppress during the stop, then resume to the thread's `True` default — is actually functionally correct. Resolve the contract honestly: either clarify the docstring that for a freshly-created `ThreadInfo` "previous state" is the thread's default (enabled), or, if a fresh thread should end untraced, suspend during the stop AND return `False` so resume is skipped. Mirror whatever you choose in the `.pyx` and regenerate the `.c`.

[verified]

src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
nshepperd: I think how it is currently is actually correct. True is the default state (for everything except pydevd-daemon threads), in the sense that every other monitoring event sets thread_info like that if it doesn't exist. So that's what we should restore.

src/debugpy/_vendored/pydevd/pydevd.py
nshepperd: I think this is fine. The meaning of "library frames" is very narrow here (a static list of stdlib files) and this basically would never happen. I'll make it a warn() log though.

tests/debugpy/test_trigger_exception_handler.py
nshepperd: Most of the tests already fail if suspend/resume is commented out due to LINE events occurring inside the stop corrupting debugpy state (which is why i originally implemented all this), so there's no need for a specific reentrancy test really.

src/debugpy/server/api.py
nshepperd: seems to be a pylance thing

nshepperd:
I think this is good now. Thank Claude for letting me stop procrastinating on this. I'll get CLA permission from my employer.

rchiodo:
Looks good overall. A couple of low-priority, non-blocking notes on a docstring/contract mismatch and a duplicated error message; safe to merge.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
```diff
@@ -155,14 +155,19 @@ def stop_on_unhandled_exception(py_db, thread, additional_info, arg):
...
+    try:
+        if exception_breakpoint.condition is not None:
+            eval_result = py_db.handle_breakpoint_condition(additional_info, exception_breakpoint, user_frame)
+            if not eval_result:
+                return
+
+        if exception_breakpoint.expression is not None:
+            py_db.handle_breakpoint_expression(exception_breakpoint, additional_info, user_frame)
+    finally:
+        remove_exception_from_frame(user_frame)
```
rchiodo: On Python 3.14, an existing `__exception__` local makes `remove_exception_from_frame()` raise `ValueError`, aborting the default stop and potentially masking the caught exception after an unconditional stop. Preserve and restore any previous binding rather than unconditionally removing it, with regression tests for both `as_uncaught` modes.

src/debugpy/_vendored/pydevd/pydevd.py
```diff
@@ -2427,6 +2427,56 @@ def do_stop_on_unhandled_exception(self, thread, frame, frames_byid, arg):
...
+            # Innermost frame that isn't debugger-internal.
+            user_frame = None
+            while tb is not None:
+                if self.get_file_type(tb.tb_frame) is None:
+                    user_frame = tb.tb_frame
+                tb = tb.tb_next
+
+            if user_frame is None:
+                pydev_log.warn("trigger_exception_handler: no user frame found in traceback")
+                return
```
rchiodo, line 2452: Using `get_file_type(...) is None` excludes `LIB_FILE` frames as well as debugger-internal frames; a library-only traceback therefore never stops despite `as_uncaught=False`. Exclude only `PYDEV_FILE` frames so unconditional mode can select ordinary library frames.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
nshepperd: Oh, you're right. It does fail on 3.14. I've fixed this with a context manager that restores the previous state, using it in both places.

src/debugpy/_vendored/pydevd/pydevd.py
nshepperd: Sure.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_breakpoints.py
nshepperd: Interestingly, Claude says this is also a problem on HEAD. It shows me this repro testcase for the userUnhandled exception handler, which seemingly hangs the unit test unless the context manager/restore method is used there:

```py
import pytest

from tests import debug
from tests.patterns import some
from tests.timeline import Event


def test_uu_repro(pyfile, target, run):
    @pyfile
    def code_to_debug():
        import threading

        import debuggee

        debuggee.setup()

        def thread_target():
            __exception__ = "sentinel"  # real local in the user boundary frame
            raise ValueError("boom")  # @raise

        t = threading.Thread(target=thread_target)
        t.start()
        t.join()
        print("main continued after thread")  # @done

    with debug.Session() as session:
        session.config["justMyCode"] = True
        session.expected_exit_code = some.int
        with run(session, target(code_to_debug)):
            session.request(
                "setExceptionBreakpoints", {"filters": ["userUnhandled"]}
            )
            session.set_breakpoints(code_to_debug, ["done"])

        session.wait_for_stop("exception")
        session.request_continue()
        session.wait_for_next(Event("exited") | Event("terminated"))
```

rchiodo: The tests do run in 3.14 as far as I know, so it should be working in main

nshepperd: Right, but you didn't have a unit test that puts an existing local named `__exception__`, as far as i can see.

nshepperd:
@microsoft-github-policy-service agree company="Anlatan"

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
```diff
@@ -43,6 +43,32 @@ def remove_exception_from_frame(frame):
...
+
+
+class exception_on_frame(object):
+    """Exposes '__exception__' in the frame's locals during a stop, then restores
+    whatever was there before. Restoring by assignment (rather than deleting) avoids
+    a ValueError on 3.14+, where FrameLocalsProxy forbids deleting a real local -- as
+    happens when the user's own code has a variable named '__exception__'."""
+
+    def __init__(self, frame, exception_info):
+        self.frame = frame
```
rchiodo, line 56: Hmm, I'm wondering if we can just stick the exception on the frame as an attribute. Modifying the locals is usually not so great because then the user sees it in the locals list (and the potential conflict). Although it did that before. But that might explain the error you were talking about with 3.14.

Something like this instead:

```python
self.frame.__stored_exception = exception_info
```

Then removing it would just be deleting that attribute.

nshepperd: TBH I'm not sure what `__exception__` is used for. I assumed the point was for users to see it in the locals list (eg. vscode debugging terminal). Or maybe it's a legacy thing used by Eclipse?

rchiodo: Looking at it, it seems it takes special precautions to remove it from the locals just for that reason. So that users don't see it. I think we can probably just change addExceptionToFrame and removeExceptionFromFrame by just adding an attribute on it. Then we wouldn't have to special case locals.

nshepperd: That can't be the case because in actual fact I do see it when i use the debugger (with vanilla debugpy release). For example, in vscode's python debugger, `__exception__` appears as "special variable" under the locals in the Variables pane. And afaict it's not used for anything internally, so if we were going to make it a frame attribute (no longer visible externally) we might as well delete it entirely. The purpose of the cleanup after the stop is so when the thread continues running (eg. userUnhandled exception path where it's not fatal) it doesn't clutter your locals during unrelated later stops.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
```diff
@@ -43,6 +43,32 @@ def remove_exception_from_frame(frame):
...
+    def __enter__(self):
+        self._previous = self.frame.f_locals.get("__exception__", _NO_EXCEPTION)
+        add_exception_to_frame(self.frame, self.exception_info)
+        return self
+
+    def __exit__(self, *args):
+        if self._previous is _NO_EXCEPTION:
+            remove_exception_from_frame(self.frame)
+        else:
+            self.frame.f_locals["__exception__"] = self._previous
```
heejaechang, line 68: If a stopped user frame declares `__exception__` but has not assigned it before raising, `f_locals.get()` returns `_NO_EXCEPTION`. `add_exception_to_frame()` then binds that real fast local, and `__exit__` takes the delete path; deleting a real local through Python 3.14's `FrameLocalsProxy` raises `ValueError`. Handle declared-but-unbound locals without deleting them, and add coverage for that case.

rchiodo (approved):

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py
rchiodo: Oh yeah it does seem it's actually on purpose.

rchiodo: Like that's its only purpose. To expose exception when stopping

rchiodo:
Thanks for the PR. I'm going to merge.

Issue #1999, opened 2026-02-24
debugpy keeps objects alive via the frame object


maxbachmann:
I am unsure what exactly changes since we didn't use debugpy in around a year. In the meantime we also switched from Python 3.11 to Python 3.13.
Our code base relies on objects being destroyed e.g. so signals are disconnected when debugging our code base.

When attempting to use debugpy again with our recent versions we experience crashes because the objects are kept alive via a frame object and so the signals aren't disconnected. Presumably this is caused by changes like the python 3.12 low impact monitoring. 

Is it known what causes this + what would be the best way to avoid this?

rchiodo:
Do you have some code that reproduces the issue? It's likely our eval that's adding a reference to something. 3.12 changed how everything worked.

maxbachmann:
Not right now but I can try to write a smaller reproducer.

maxbachmann:
[signal.py](https://github.com/user-attachments/files/25752106/signal.py)
[mySignal.cpp](https://github.com/user-attachments/files/25752330/mySignal.cpp)

I finally managed to make a reproducer for my issue I can share

The repo includes a `signal.py` to debug and a `mySignal.cpp` for a C++ signal that is required to trigger the issue. 
To compile the C++ signal on Linux I did run
```
pip install pybind11
c++ -O3 -Wall -shared -std=c++11 -fPIC $(python3 -m pybind11 --includes) mySignal.cpp -o mySignal$(python3-config --extension-suffix)
```

Running signal.py without debugpy on python3.11/3.13 just prints switch twice and then exits.
Running it under debugpy (python3.11) in vs code also  just prints switch twice and then exits.
Running it under debugpy (python3.13) in vs code triggers the assert in line 72.

This occurs because the reference is kept alive -> the signal continues to call it when I already expect it to be dead.
I don't have any breakpoints or watched variables.

rchiodo:
Thanks for the repro. Might be a while before somebody gets to this though. 

I was going to try this. It should theoretically show where the references are coming from.

```
import objgraph
objgraph.show_backrefs(obj, max_depth=5)
```

maxbachmann:
I did slighly extend using

```py
def main():
    a = StateMachine(GameState)
    a.execute()
    import weakref

    state = weakref.ref(a.state)
    while not shutdown:
        nextFrame()
        b = state()
        if b is not None:
            import objgraph
            objgraph.show_backrefs([b], max_depth=5)
        b = None

    a.leave()
```

[...]

PR #2008, opened 2026-03-07
avoid strong reference to exceptions during unwind

maxbachmann:
For now this is only a draft of a fix for https://github.com/microsoft/debugpy/issues/1999 to discuss the approach. This attempts to add a tag to exceptions during the unwind that can be used to detect similar exceptions without holding a strong reference. If for some reason no attribute can be added it falls back to a strong reference on the exception.

I didn't test this yet since I first wanted to discuss the approach. I would still prefer to just check the frame during the unwind to see whether the frame should trigger the break. However it sounds like for some unknown reason the frame cache was necessary.


maxbachmann:
As reference: This is the alternative approach that doesn't have any side effects https://github.com/microsoft/debugpy/pull/2009 but where it was unclear why the cache was needed.

I am fine with either approach since they should both fix our problem :)

rchiodo:
No, I know the exception is there for a reason. Not tracking it is going to break something. Maybe the current pydevd tests cover that scenario and that's why I didn't need to add any. 

The fix in this PR seems good to me. No worries about the exception not being tracked and it adds a hidden member to an exception object.

maxbachmann:
btw is there anything special I have to do when generating the cython files? Because they include changes that certainly do not stem from me. I did push them.

rchiodo:
> btw is there anything special I have to do when generating the cython files? Because they include changes that certainly do not stem from me. I did push them.

Yeah something is off. I'll try on monday with your change and see if I get the same results. 

I'm using Windows with MSVC to generate them.

maxbachmann:
I did generate them on Linux

maxbachmann:
I did now test this on our code base and it fixes our crashes under the debugger. So this is ready for review for you next week

PR #2018, opened 2026-04-07
Fix evaluation of variables from chained exception frames

pdepetro:
When stopped on a chained exception, selecting a frame from the cause/context exception in the call stack would show "Unable to find thread to evaluate variable reference." for every variable. This happened because chained exception frames were included in the stackTrace DAP response but were never registered in the SuspendedFramesManager for variable lookup.

Modified _FramesTracker.track() to always walk the chained_frames_list and register those frames in the same tracker, enabling variable evaluation for chained exception frames.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
@@ -404,6 +404,21 @@ class _FramesTracker(object):
 
                 self._frame_id_to_main_thread_id[frame_id] = thread_id
 
+            # Also track frames from chained exceptions (e.g. __cause__ / __context__)
+            # so that variable evaluation works for chained exception frames displayed
+            # in the call stack.
+            chained = getattr(frames_list, 'chained_frames_list', None)
+            while chained is not None and len(chained) > 0:
+                for frame in chained:
+                    frame_id = id(frame)
+                    self._frame_id_to_frame[frame_id] = frame
+                    _FrameVariable(self.py_db, frame, self._register_variable)
+                    self._suspended_frames_manager._variable_reference_to_frames_tracker[frame_id] = self
+                    frame_ids_from_thread.append(frame_id)
+
+                    self._frame_id_to_main_thread_id[frame_id] = thread_id
+                chained = getattr(chained, 'chained_frames_list', None)
+
             frame = None
 
     def untrack_all(self):
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
@@ -394,15 +394,17 @@ class _FramesTracker(object):
 
             frame_ids_from_thread = self._thread_id_to_frame_ids.setdefault(coroutine_or_main_thread_id, [])
 
-            self._thread_id_to_frames_list[coroutine_or_main_thread_id] = frames_list
-            for frame in frames_list:
+            def _register_frame(frame):
                 frame_id = id(frame)
                 self._frame_id_to_frame[frame_id] = frame
                 _FrameVariable(self.py_db, frame, self._register_variable)  # Instancing is enough to register.
                 self._suspended_frames_manager._variable_reference_to_frames_tracker[frame_id] = self
                 frame_ids_from_thread.append(frame_id)
-
                 self._frame_id_to_main_thread_id[frame_id] = thread_id
+
+            self._thread_id_to_frames_list[coroutine_or_main_thread_id] = frames_list
+            for frame in frames_list:
+                _register_frame(frame)
 
             # Also track frames from chained exceptions (e.g. __cause__ / __context__)
             # so that variable evaluation works for chained exception frames displayed
@@ -410,13 +412,7 @@ class _FramesTracker(object):
             chained = getattr(frames_list, 'chained_frames_list', None)
             while chained is not None and len(chained) > 0:
                 for frame in chained:
-                    frame_id = id(frame)
-                    self._frame_id_to_frame[frame_id] = frame
-                    _FrameVariable(self.py_db, frame, self._register_variable)
-                    self._suspended_frames_manager._variable_reference_to_frames_tracker[frame_id] = self
-                    frame_ids_from_thread.append(frame_id)
-
-                    self._frame_id_to_main_thread_id[frame_id] = thread_id
+                    _register_frame(frame)
                 chained = getattr(chained, 'chained_frames_list', None)
 
             frame = None
--- a/src/debugpy/_vendored/pydevd/tests_python/test_debugger_json.py
+++ b/src/debugpy/_vendored/pydevd/tests_python/test_debugger_json.py
@@ -926,6 +926,80 @@ def test_case_throw_exc_reason(case_setup_dap):
         writer.finished_ok = True
 
 
+def test_case_chained_exception_variables(case_setup_dap, pyfile):
+    """
+    When stopped on a chained exception, variable evaluation must work for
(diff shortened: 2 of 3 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
```diff
@@ -404,6 +404,21 @@ def track(self, thread_id, frames_list, frame_custom_thread_id=None):
 
                 self._frame_id_to_main_thread_id[frame_id] = thread_id
 
+            # Also track frames from chained exceptions (e.g. __cause__ / __context__)
```
rchiodo, line 409: Copilot generated:
-419

**Missing cycle detection on chained-frame walk (Medium-High).** The `while chained is not None` loop follows `chained_frames_list` links without tracking visited nodes. If the exception chain is cyclic (possible pre-3.11 via `__context__`, or via manual `__cause__` assignment), this hangs the debugger. CPython's own `traceback.py` uses a `_seen` set for this reason. Suggested fix:
```python
seen = set()
chained = getattr(frames_list, 'chained_frames_list', None)
while chained is not None and id(chained) not in seen and len(chained) > 0:
    seen.add(id(chained))
    ...
    chained = getattr(chained, 'chained_frames_list', None)
```
Even if `create_frames_list_from_traceback` currently prevents cycles, this is cheap defense-in-depth.

rchiodo, line 409: Copilot generated:
-419

**Duplicated frame registration logic (Medium).** The 5-line per-frame registration block (`_frame_id_to_frame`, `_FrameVariable`, `_variable_reference_to_frames_tracker`, `frame_ids_from_thread.append`, `_frame_id_to_main_thread_id`) is duplicated verbatim from the primary loop ~10 lines above. Since this duplication is being *introduced* in this PR, now is the right time to extract a `_register_frame(self, frame, thread_id, frame_ids_from_thread)` helper called from both loops. This eliminates drift risk if the registration contract changes.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
```diff
@@ -404,6 +404,21 @@ def track(self, thread_id, frames_list, frame_custom_thread_id=None):
...
+            # so that variable evaluation works for chained exception frames displayed
+            # in the call stack.
+            chained = getattr(frames_list, 'chained_frames_list', None)
```
rchiodo, line 412: Copilot generated:
**Potential issue with synthetic separator frames (Low).** Chained exception stacks may contain synthetic separator frames (e.g., `[Chained Exc: ...]`). If these appear in `chained_frames_list`, calling `_FrameVariable(self.py_db, frame, ...)` on a non-real frame could raise or produce nonsensical variable listings. If separator objects do appear here, they should be skipped or handled gracefully.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
```diff
@@ -404,6 +404,21 @@ def track(self, thread_id, frames_list, frame_custom_thread_id=None):
...
+            while chained is not None and len(chained) > 0:
+                for frame in chained:
+                    frame_id = id(frame)
```
rchiodo: Copilot generated:
**`frame_custom_thread_id` not considered for chained frames (Low-Medium).** The original primary-frame tracking code has `frame_custom_thread_id` logic, but the new chained-frame block unconditionally uses `thread_id`. If `frame_custom_thread_id` remaps the thread identity in some contexts, chained frames would map to the wrong thread. Worth verifying whether `frame_custom_thread_id` is ever non-None when chained exceptions are present.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
pdepetro: @rchiodo there is already cycle detection in create_frames_list_from_traceback(); do you still want me to add it again here?

rchiodo: Sorry I'm not seeing it there? This logic?
https://github.com/microsoft/debugpy/blob/6d265e0eda5be94f6c0f1e736eb651b680a7ba63/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py#L372

AFAICT it:
- Adds the topmost frame to the list
- goes through all the frames below that one and stick them in a list
- reverses the list

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
pdepetro: @rchiodo Where are the synthetic separator frames added? I think the "[Chained Exc:" bit is added a prefix to the method name when displaying the frames.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
pdepetro: memo is passed into create_frames_list_from_exception_cause() and used for cycle detection.

https://github.com/microsoft/debugpy/blob/6d265e0eda5be94f6c0f1e736eb651b680a7ba63/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py#L424-L429

https://github.com/microsoft/debugpy/blob/6d265e0eda5be94f6c0f1e736eb651b680a7ba63/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py#L321-L325

rchiodo: Ah that makes sense. Yeah seems good enough.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
pdepetro: I don't think this can happen when chained exceptions are present.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
rchiodo: Here:

https://github.com/microsoft/debugpy/blob/6d265e0eda5be94f6c0f1e736eb651b680a7ba63/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame_utils.py#L338

rchiodo: That doesn't seem like extra frames though. That's just a wrapper. I'm thinking copilot might be incorrect here. Synthetic frames might be added when responding to the DAP request, not here.

rchiodo: Yeah here. https://github.com/microsoft/debugpy/blob/6d265e0eda5be94f6c0f1e736eb651b680a7ba63/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_net_command_factory_xml.py#L213

I think we can ignore this comment too.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_suspended_frames.py
rchiodo: Yeah I agree. Resolving.

rchiodo (approved):

rchiodo:
Thanks for the PR

PR #2024, opened 2026-05-12
Skip uninitialized __slots__ in variable resolver

pdepetro:
Ignore unused slots instead of showing the attribute with an exception.


rchiodo (approved):
Thanks for the submission

PR #2026, opened 2026-05-12
Prevent invalid scopes request from crashing debug session

pdepetro:
If an unmapped or non-numeric frameId is sent in a scopes request, int() raises a ValueError that terminates the debug session. Wrap the scope construction in a try/except so that invalid frameIds return an empty scopes list instead, which is valid per the DAP spec.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
@@ -1109,10 +1109,13 @@ class PyDevJsonCommandProcessor(object):
         frame_id = request.arguments.frameId
 
         variables_reference = frame_id
-        scopes = [
-            Scope("Locals", ScopeRequest(int(variables_reference), "locals"), False, presentationHint="locals"),
-            Scope("Globals", ScopeRequest(int(variables_reference), "globals"), False),
-        ]
+        try:
+            scopes = [
+                Scope("Locals", ScopeRequest(int(variables_reference), "locals"), False, presentationHint="locals"),
+                Scope("Globals", ScopeRequest(int(variables_reference), "globals"), False),
+            ]
+        except Exception:
+            scopes = []
         body = ScopesResponseBody(scopes)
         scopes_response = pydevd_base_schema.build_response(request, kwargs={"body": body})
         return NetCommand(CMD_RETURN, 0, scopes_response, is_json=True)
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
@@ -1114,7 +1114,11 @@ class PyDevJsonCommandProcessor(object):
                 Scope("Locals", ScopeRequest(int(variables_reference), "locals"), False, presentationHint="locals"),
                 Scope("Globals", ScopeRequest(int(variables_reference), "globals"), False),
             ]
+        except (ValueError, TypeError):
+            pydev_log.info("Failed to create scopes for frame id: %s", frame_id)
+            scopes = []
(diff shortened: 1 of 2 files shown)


src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -1109,10 +1109,13 @@ def on_scopes_request(self, py_db, request):
         frame_id = request.arguments.frameId
 
         variables_reference = frame_id
-        scopes = [
-            Scope("Locals", ScopeRequest(int(variables_reference), "locals"), False, presentationHint="locals"),
-            Scope("Globals", ScopeRequest(int(variables_reference), "globals"), False),
-        ]
+        try:
+            scopes = [
+                Scope("Locals", ScopeRequest(int(variables_reference), "locals"), False, presentationHint="locals"),
```
rchiodo, line 1114: Copilot generated:
`except Exception` is too broad. The PR description identifies `ValueError` from `int()` as the failure mode — catch `(ValueError, TypeError)` instead. The current catch will silently swallow any future bug in `Scope()` or `ScopeRequest()` constructors (e.g., `AttributeError`, `KeyError`), making those bugs invisible. All three technical reviewers flagged this independently. `[unverified]`

[verified]

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py
```diff
@@ -1109,10 +1109,13 @@ def on_scopes_request(self, py_db, request):
...
+                Scope("Globals", ScopeRequest(int(variables_reference), "globals"), False),
```
rchiodo, line 1115: Copilot generated:
The `except` block has no logging. When this triggers in production, there's zero diagnostic trail — the user sees an empty Variables panel with no explanation. Add at minimum `pydev_log.info('Invalid frameId %r in scopes request', variables_reference)` so developers can trace why variables aren't showing. The Advocate notes the codebase has precedent for logging in exception handlers (e.g., line 534 uses `pydev_log.exception`). `[unverified]`

[verified]

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_process_net_command_json.py
```diff
@@ -0,0 +1,32 @@
...
+    assert scopes[0]["presentationHint"] == "locals"
+    assert scopes[1]["name"] == "Globals"
+
+
+def test_on_scopes_request_with_invalid_frame_id():
+    from _pydevd_bundle._debug_adapter import pydevd_schema, pydevd_base_schema
+    from _pydevd_bundle.pydevd_process_net_command_json import PyDevJsonCommandProcessor
+
+    processor = PyDevJsonCommandProcessor(pydevd_base_schema.from_json)
+    request = pydevd_schema.ScopesRequest(
```
rchiodo, line 24: Copilot generated:
-26

`frameId="not_a_number"` tests a scenario that likely can't occur via the DAP protocol (which defines `frameId` as integer). The more realistic failure mode is a valid integer that references a stale/dead frame (e.g., `frameId=999999`). Consider adding a test for that case — it would exercise a more likely production failure path. `[unverified]`

[verified]

pdepetro: 999999 is a valid int, so it wouldn't exercise what we're preventing here, which would be a non-integer value.

pdepetro: The specific issue we've seen is that the client can send a scopes request using a DAP id that corresponds to a thread instead of a frame. The DAP to id translation converts that to something like `pid_1996270_id_140192610058896` which raises an exception when passed to int().

rchiodo (approved):
Approved via Review Center.

PR #2028, opened 2026-05-28
Fix TSA #2816216: suppress Flawfinder false positive on Cython DIGIT_PAIRS_8 memcpy

StellaHuang95:
## Issue

Internal Flawfinder compliance scan flagged a `memcpy` in the Cython-generated `pydevd_cython.c`.

- Work item: [TSA #2816216](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2816216/)
- Rule: `FlawFinder/buffer/memcpy` — *"Does not check for buffer overflows when copying to destination (CWE-120)"*
- File: `src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c`
- Line: 45636
- AB#2816216

## What Flawfinder reported

```c
memcpy(dpos, DIGIT_PAIRS_8 + digit_pos * 2, 2);
```

inside the Cython 3.x integer-to-string helper `__Pyx____Pyx_PyUnicode_From_int` (case `'o'` of the format-char switch).

Flawfinder flags every `memcpy()` call by default because it can't statically prove the size argument is safe. In this case all three arguments are provably bounded:

1. **Destination is in bounds.** The buffer is a local stack array declared a few lines above: `char digits[sizeof(int)*3+2]`. Each iteration does `dpos -= 2` *before* the 2-byte copy, and the enclosing `do/while` loop divides `remaining` by 64 each pass, so it iterates at most ⌈log₆₄(INT_MAX)⌉ times — well within the allocated buffer.
2. **Source is in bounds.** `DIGIT_PAIRS_8` is a 128-byte compile-time constant table (64 pairs × 2 ASCII chars). `digit_pos = abs(remaining % 64)`, so the offset `digit_pos * 2` is in `[0, 126]` — the 2-byte read is always inside the table.
3. **Size is a compile-time literal `2`.** No way for it to be wrong.

This is Cython runtime boilerplate that mirrors the same pattern used in CPython itself.

## Fix

## Verification

| Run | Hits at line 45636 |
|---|---|
| Default | **0** ✅ |
| `flawfinder --neverignore` (suppressions disabled) | 1 (the originally reported warning reappears) |

## Risk

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
@@ -45633,7 +45633,7 @@ static CYTHON_INLINE PyObject* __Pyx____Pyx_PyUnicode_From_int(int value, Py_ssi
             digit_pos = abs((int)(remaining % (8*8)));
             remaining = (int) (remaining / (8*8));
             dpos -= 2;
-            memcpy(dpos, DIGIT_PAIRS_8 + digit_pos * 2, 2);
+            memcpy(dpos, DIGIT_PAIRS_8 + digit_pos * 2, 2); /* Flawfinder: ignore */
             last_one_off = (digit_pos < 8);
             break;
         case 'd':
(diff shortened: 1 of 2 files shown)


rchiodo (approved):
Approved via Review Center.

StellaHuang95:
Pushed 272f6e52 to fix a `SyntaxError: '(' was never closed` that snuck in via the merge from main — the closing `)` of the DIGIT_PAIRS_8 `.replace(...)` call was dropped when the `read<end` `.replace(...)` block was spliced in (same pattern as the issue caught in #2029). `py_compile` now passes.

rchiodo (approved):
Approved via Review Center.

PR #2029, opened 2026-05-28
Fix TSA #2816217: suppress Flawfinder false positive on Cython JoinPyUnicode memcpy

StellaHuang95:
## Issue

Internal Flawfinder compliance scan flagged a `memcpy` in the Cython-generated `_pydevd_sys_monitoring_cython.c`.

- Work item: [TSA #2816217](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2816217/)
- Rule: `FlawFinder/buffer/memcpy` — *"Does not check for buffer overflows when copying to destination (CWE-120)"*
- File: `src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring_cython.c`
- Line: 38188
- AB#2816217

## What Flawfinder reported

```c
memcpy((char *)result_udata + (char_pos << kind_shift),
       udata,
       (size_t)(ulength << kind_shift));
```

inside the Cython 3.x string-join helper `__Pyx_PyUnicode_Join` (Cython's implementation of `"".join([...])`).

Flawfinder flags every `memcpy()` because the size argument here is a variable (`ulength << kind_shift`) rather than a constant. But the surrounding code makes the copy provably safe:

1. **Destination is pre-allocated to the exact total length.** A few lines above:
   ```c
   result_uval = PyUnicode_New(result_ulength, max_char);
   result_udata = PyUnicode_DATA(result_uval);
   ```
   `result_ulength` is the sum of all input lengths, computed by the caller.

2. **Overflow is explicitly checked before each `memcpy`.** Two lines above the flagged line:
   ```c
   if (unlikely((PY_SSIZE_T_MAX >> kind_shift) - ulength < char_pos))
       goto overflow;
   ```
   This guards against `char_pos + ulength` exceeding the addressable range. If it would, the code skips the copy and reports an `OverflowError`.

3. **`char_pos` accumulates the same `ulength` values that sized the buffer**, so by induction `char_pos + ulength ≤ result_ulength` at every iteration.

This mirrors CPython's own `str.join()` implementation in `Objects/unicodeobject.c`.

## Fix

## Verification

| Run | Hits at line 38188 |
|---|---|
| Default | **0** ✅ |
| `flawfinder --neverignore` (suppressions disabled) | 1 (the originally reported warning reappears) |

## Risk

### DIFF
--- a/src/debugpy/_vendored/pydevd/setup_pydevd_cython.py
+++ b/src/debugpy/_vendored/pydevd/setup_pydevd_cython.py
@@ -177,6 +177,17 @@ def build_extension(dir_name, extension_name, target_pydevd_name, force_cython, 
                 c_file_contents = c_file_contents.replace(r"_pydevd_bundle\\pydevd_cython.pxd", "_pydevd_bundle/pydevd_cython.pxd")
                 c_file_contents = c_file_contents.replace(r"_pydevd_bundle\\pydevd_cython.pyx", "_pydevd_bundle/pydevd_cython.pyx")
 
+                # Suppress Flawfinder false positive (CWE-120) in the Cython 3.x
+                # `__Pyx_PyUnicode_Join` boilerplate: the destination `result_uval` was just
+                # allocated via `PyUnicode_New(result_ulength, max_char)`, and the immediately
+                # preceding `(PY_SSIZE_T_MAX >> kind_shift) - ulength < char_pos` check guards
+                # against char_pos+ulength overflow before the memcpy. The size argument is
+                # `ulength << kind_shift` which is bounded by the pre-allocated buffer length.
+                c_file_contents = c_file_contents.replace(
+                    "            memcpy((char *)result_udata + (char_pos << kind_shift), udata, (size_t) (ulength << kind_shift));\n",
+                    "            memcpy((char *)result_udata + (char_pos << kind_shift), udata, (size_t) (ulength << kind_shift)); /* Flawfinder: ignore */\n",
+                )
+
                 with open(c_file, "w") as stream:
                     stream.write(c_file_contents)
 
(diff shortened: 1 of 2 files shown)


rchiodo (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/setup_pydevd_cython.py
```diff
@@ -177,6 +177,15 @@ def build_extension(dir_name, extension_name, target_pydevd_name, force_cython,
                 c_file_contents = c_file_contents.replace(r"_pydevd_bundle\\pydevd_cython.pxd", "_pydevd_bundle/pydevd_cython.pxd")
                 c_file_contents = c_file_contents.replace(r"_pydevd_bundle\\pydevd_cython.pyx", "_pydevd_bundle/pydevd_cython.pyx")
 
+                # Suppress Flawfinder false positive (CWE-120) in the Cython 3.x
+                # `__Pyx_PyUnicode_Join` boilerplate: the destination `result_uval` was just
+                # allocated via `PyUnicode_New(result_ulength, max_char)`, and the immediately
+                # preceding `(PY_SSIZE_T_MAX >> kind_shift) - ulength < char_pos` check guards
```
rchiodo, line 183: Copilot generated:
-185

**Critical: Missing closing `)` — SyntaxError.** The new `.replace(` call has two string arguments ending with a trailing comma but no closing parenthesis. Python will attempt to parse the subsequent comment block and next assignment as additional arguments, raising `SyntaxError` at parse time. This breaks the entire Cython build script.

Fix: add `)` after the second argument:
```python
c_file_contents = c_file_contents.replace(
    "            memcpy((char *)result_udata + (char_pos << kind_shift), udata, (size_t) (ulength << kind_shift));\n",
    "            memcpy((char *)result_udata + (char_pos << kind_shift), udata, (size_t) (ulength << kind_shift)); /* Flawfinder: ignore */\n",
)
```
All four reviewers flagged this independently — structurally confirmed from the diff.

[verified]

StellaHuang95: Good catch — confirmed `SyntaxError: '(' was never closed` at parse time. Root cause was the merge from main: the closing `)` of the JoinPyUnicode `.replace(...)` got eaten when the new `read<end` `.replace(...)` block was spliced in, so the second call was being parsed as continued positional args to the first.

Fixed in 6bebc9d1 by adding the missing `)` (and a blank line) to cleanly separate the two calls. `py_compile` now passes.

rchiodo (approved):
Approved via Review Center.

bschnurr (approved):

rchiodo (approved):

PR #2032, opened 2026-05-28
Fix TSA #2816220: suppress Flawfinder false positive on Cython read-loop iterator (pydevd_cython.c)

StellaHuang95:
## Issue

Internal Flawfinder compliance scan flagged a `for` loop containing the identifier `read` in the Cython-generated `pydevd_cython.c`.

- Work item: [TSA #2816220](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2816220/)
- Rule: `FlawFinder/buffer/read` — *"Check buffer boundaries if used in a loop including recursive loops (CWE-120, CWE-20)"*
- File: `src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c`
- Line: 51788
- AB#2816220

## What Flawfinder reported

```c
__Pyx_InterpreterIdAndModule *read  = data->table;
__Pyx_InterpreterIdAndModule *write = data->table;
__Pyx_InterpreterIdAndModule *end   = read + data->count;
for (; read<end; ++read) {                  // ← flagged
    if (read->module) {
        write->id     = read->id;
        write->module = read->module;
        ++write;
    }
}
```

This is the Cython 3.x runtime boilerplate function `__Pyx_State_ConvertFromInterpIdAsIndex` (per-interpreter module-state compaction). The same function is emitted into every Cython-generated `.c` file.

Flawfinder's `buffer/read` rule fires whenever the identifier `read` appears inside a loop, on the assumption it refers to the POSIX `read()` syscall (which can return fewer bytes than requested). Here `read` is a **variable name** — specifically a pointer iterator walking `data->table` — not a syscall.

The loop is provably bounded:
- `end = read + data->count` defines the upper bound.
- The loop condition `read < end` guarantees `read` never advances past the last valid slot.
- `write` only advances on successful copies and always trails `read`, so `write <= read <= end` is invariant.

## Fix

## Verification

| Run | Hits at line 51788 |
|---|---|
| Default | **0** ✅ |
| `flawfinder --neverignore` (suppressions disabled) | 2 (the originally reported warning reappears — one hit per `read` token on that line) |

## Risk

## Related

The same Cython boilerplate appears in two sibling generated files and is tracked separately:
- TSA #2816218 → `pydevd_frame_evaluator.c`
- TSA #2816219 → `_pydevd_sys_monitoring_cython.c`

Each is fixed in its own PR.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_cython.c
@@ -51785,7 +51785,7 @@ static void __Pyx_State_ConvertFromInterpIdAsIndex(__Pyx_ModuleStateLookupData *
     __Pyx_InterpreterIdAndModule *read = data->table;
     __Pyx_InterpreterIdAndModule *write = data->table;
     __Pyx_InterpreterIdAndModule *end = read + data->count;
-    for (; read<end; ++read) {
+    for (; read<end; ++read) { /* Flawfinder: ignore */
         if (read->module) {
             write->id = read->id;
             write->module = read->module;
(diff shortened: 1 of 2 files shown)


rchiodo (approved):
Approved via Review Center.

PR #2040, opened 2026-06-19
Fix GraalPy compatibility

steve-s:
Fixes `debugpy` compatibility with GraalPy, an alternative Python implementation.

### DIFF
--- a/src/debugpy/adapter/__main__.py
+++ b/src/debugpy/adapter/__main__.py
@@ -12,6 +12,7 @@ import sys
 # WARNING: debugpy and submodules must not be imported on top level in this module,
 # and should be imported locally inside main() instead.
 
+_CAN_DAEMONIZE = os.name == "posix" and hasattr(os, "fork")
 
 def main():
     args = _parse_argv(sys.argv)
@@ -31,12 +32,13 @@ def main():
         if os.name == "posix":
             # On POSIX, we need to leave the process group and its session, and then
             # daemonize properly by double-forking (first fork already happened when
-            # this process was spawned).
+            # this process was spawned). Some POSIX runtimes, such as GraalPy, do not
+            # implement fork(), so in that case we settle for a single detached child.
             # NOTE: if process is already the session leader, then
             # setsid would fail with `operation not permitted`
             if os.getsid(os.getpid()) != os.getpid():
                 os.setsid()
-            if os.fork() != 0:
+            if _CAN_DAEMONIZE and os.fork() != 0:
                 sys.exit(0)
 
         for stdio in sys.stdin, sys.stdout, sys.stderr:
--- a/src/debugpy/adapter/__main__.py
+++ b/src/debugpy/adapter/__main__.py
@@ -12,7 +12,6 @@ import sys
 # WARNING: debugpy and submodules must not be imported on top level in this module,
 # and should be imported locally inside main() instead.
 
-_CAN_DAEMONIZE = os.name == "posix" and hasattr(os, "fork")
 
 def main():
     args = _parse_argv(sys.argv)
@@ -38,7 +37,7 @@ def main():
             # setsid would fail with `operation not permitted`
             if os.getsid(os.getpid()) != os.getpid():
                 os.setsid()
-            if _CAN_DAEMONIZE and os.fork() != 0:
+            if hasattr(os, "fork") and os.fork() != 0:
                 sys.exit(0)
 
         for stdio in sys.stdin, sys.stdout, sys.stderr:
--- a/src/debugpy/launcher/debuggee.py
+++ b/src/debugpy/launcher/debuggee.py
@@ -58,9 +58,12 @@ def spawn(process_name, cmdline, env, redirect_output):
         else:
             kwargs = {}
 
-        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
-            # GraalPy does not support running code between fork and exec
+        # GraalPy does not support running code between fork and exec, but supports the
+        # process_group argument for Popen
+        if sys.platform != "win32" and sys.implementation.name == "graalpy":
+            kwargs.update(process_group=0)
 
+        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
             def preexec_fn():
                 try:
                     # Start the debuggee in a new process group, so that the launcher can
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -17,6 +17,7 @@ from pydevd_file_utils import absolute_path
 from debugpy.common.util import hide_debugpy_internals
 
 _tls = threading.local()
+_CAN_DAEMONIZE = os.name == "posix" and hasattr(os, "fork")
 
 # TODO: "gevent", if possible.
 _config = {
@@ -218,11 +219,13 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 creationflags=creationflags,
                 env=python_env,
             )
-            if os.name == "posix":
+            if _CAN_DAEMONIZE:
                 # It's going to fork again to daemonize, so we need to wait on it to
-                # clean it up properly.
+                # clean it up properly. If we did not fork, we cannot take this path
+                # because it cannot perform that extra fork; waiting there would just
+                # preserve the broken assumption that a daemonized grandchild exists.
                 _adapter_process.wait()
-            else:
+            elif os.name != "posix":
                 # Suppress misleading warning about child process still being alive when
                 # this process exits (https://bugs.python.org/issue38890).
                 _adapter_process.returncode = 0
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -17,7 +17,6 @@ from pydevd_file_utils import absolute_path
 from debugpy.common.util import hide_debugpy_internals
 
 _tls = threading.local()
-_CAN_DAEMONIZE = os.name == "posix" and hasattr(os, "fork")
 
 # TODO: "gevent", if possible.
 _config = {
@@ -219,7 +218,7 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 creationflags=creationflags,
                 env=python_env,
             )
-            if _CAN_DAEMONIZE:
+            if os.name == "posix" and hasattr(os, "fork"):
                 # It's going to fork again to daemonize, so we need to wait on it to
                 # clean it up properly. If we did not fork, we cannot take this path
                 # because it cannot perform that extra fork; waiting there would just
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -225,7 +225,7 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 # because it cannot perform that extra fork; waiting there would just
                 # preserve the broken assumption that a daemonized grandchild exists.
                 _adapter_process.wait()
-            elif os.name != "posix":
+            else:
                 # Suppress misleading warning about child process still being alive when
                 # this process exits (https://bugs.python.org/issue38890).
                 _adapter_process.returncode = 0


rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/server/api.py:L233](https://github.com/microsoft/debugpy/blob/9eb0cde1fcabe7a2ef2f014d6aaa0144f3a46894/src/debugpy/server/api.py#L233).

📍 src/debugpy/server/api.py:222
GraalPy falls through **both** branches here: `_CAN_DAEMONIZE` is `False` yet `os.name == "posix"` is `True`, so neither `if` nor `elif os.name != "posix"` executes. Because the `__main__.py` change means the GraalPy adapter no longer double-forks, it is a long-lived *direct* child just like the Windows case and needs the same treatment. As written, GraalPy will miss `returncode = 0` and re-emit the bpo-38890 "child still alive" warning. Fix: key the second branch on child lifetime, not OS name — use a plain `else:` (i.e. `not _CAN_DAEMONIZE`, covering Windows AND GraalPy) that sets `returncode = 0`.

src/debugpy/launcher/debuggee.py
```diff
@@ -58,9 +58,12 @@ def spawn(process_name, cmdline, env, redirect_output):
         else:
             kwargs = {}
 
-        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
-            # GraalPy does not support running code between fork and exec
+        # GraalPy does not support running code between fork and exec, but supports the
+        # process_group argument for Popen
+        if sys.platform != "win32" and sys.implementation.name == "graalpy":
```
rchiodo, line 63: 📍 src/debugpy/launcher/debuggee.py:62
The `process_group` keyword for `subprocess.Popen` was only added in Python 3.11. A GraalPy build targeting a pre-3.11 stdlib (e.g. GraalPy 23.x → 3.10) will raise `TypeError: got an unexpected keyword argument 'process_group'`, which is caught and surfaced as "Couldn't spawn debuggee" — making launch fail entirely rather than degrading. Guard it, e.g. `if ... graalpy and sys.version_info >= (3, 11): kwargs.update(process_group=0)`, or wrap in `try/except TypeError`.

rchiodo:
Solid GraalPy compatibility work overall. The main blocker is the `if _CAN_DAEMONIZE / elif os.name != "posix"` structure in api.py, which silently drops GraalPy through both branches. Keying the second branch on child lifetime (`else:`) rather than OS name fixes it.

src/debugpy/launcher/debuggee.py
steve-s: This would also require adding another branch to `kill` so that we use `os.kill` and not `os.killpg` for older GraalPy versions. Moreover, support for current and older GraalPy releases depends on https://github.com/fabioz/PyDev.Debugger/pull/325, and unless that's merged and propagated to debugpy I think it's better to keep the changes in this file simple - it will work on GraalPy master (and some future release) for which fixes in https://github.com/fabioz/PyDev.Debugger/pull/325 aren't necessary.

steve-s:
Thank you for the review! I addressed the blocker in the last commit. I will have to figure out CLA with my company.

rchiodo:
Small, well-scoped GraalPy compatibility fix. Note: the api.py daemonize change correctly routes GraalPy into the plain `else` branch (same path as Windows), so it still receives `returncode = 0` and the don't-terminate registration — no missed-warning regression. Looks good to merge.

rchiodo (approved):
Approved via Review Center.

src/debugpy/adapter/__main__.py
```diff
@@ -31,12 +32,13 @@ def main():
...
             # daemonize properly by double-forking (first fork already happened when
-            # this process was spawned).
+            # this process was spawned). Some POSIX runtimes, such as GraalPy, do not
+            # implement fork(), so in that case we settle for a single detached child.
             # NOTE: if process is already the session leader, then
             # setsid would fail with `operation not permitted`
             if os.getsid(os.getpid()) != os.getpid():
                 os.setsid()
-            if os.fork() != 0:
+            if _CAN_DAEMONIZE and os.fork() != 0:
```
rchiodo: The extra variable outside of the if statement seems like overkill. Why not just have this:

```
if hasattr(os, "fork") and os.fork() != 0:
```

src/debugpy/server/api.py
```diff
@@ -218,9 +219,11 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 creationflags=creationflags,
                 env=python_env,
             )
-            if os.name == "posix":
+            if _CAN_DAEMONIZE:
```
rchiodo: Same thing here. It would be

```
if os.name == "posix" and hasattr(os, "fork")
```

Makes for fewer code changes.

src/debugpy/adapter/__main__.py
```diff
@@ -31,12 +31,13 @@ def main():
...
             # daemonize properly by double-forking (first fork already happened when
-            # this process was spawned).
+            # this process was spawned). Some POSIX runtimes, such as GraalPy, do not
+            # implement fork(), so in that case we settle for a single detached child.
             # NOTE: if process is already the session leader, then
             # setsid would fail with `operation not permitted`
             if os.getsid(os.getpid()) != os.getpid():
                 os.setsid()
-            if os.fork() != 0:
+            if hasattr(os, "fork") and os.fork() != 0:
```
rchiodo, line 40: 📍 src/debugpy/adapter/__main__.py:38
This PR adds a `hasattr(os, "fork")` probe but leaves the sibling syscalls `os.getsid`/`os.setsid` (a few lines up in the same `if os.name == "posix":` block) gated only on `os.name == "posix"`. A forkless-POSIX runtime that also lacks `setsid` would raise `AttributeError`/`OSError` here and crash *before* reaching the new fork guard — defeating the PR's goal. Please confirm GraalPy exposes `setsid`/`getsid`, or guard with `hasattr(os, "setsid")` / wrap in `try/except (AttributeError, OSError)`.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

src/debugpy/adapter/__main__.py
steve-s: yes, GraalPy exposes those functions

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/launcher/debuggee.py:L36](https://github.com/microsoft/debugpy/blob/81d199b958126344bd74dbb36325c79772f0270a/src/debugpy/launcher/debuggee.py#L36).

📍 src/debugpy/launcher/debuggee.py:33

This file detects the fork-less runtime by **identity** (`sys.implementation.name == "graalpy"`), while `adapter/__main__.py` and `server/api.py` detect it by **capability** (`hasattr(os, "fork")`). They encode the same concept two ways, so a future fork-less POSIX runtime would be routed correctly by the other two files but fall into the `!= 'graalpy'` branch here and attempt `os.setpgrp()` between fork and exec — exactly what the comment says is impossible. Consider unifying all three sites behind one shared predicate (e.g. a `supports_fork()` helper).

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/launcher/debuggee.py:L36](https://github.com/microsoft/debugpy/blob/81d199b958126344bd74dbb36325c79772f0270a/src/debugpy/launcher/debuggee.py#L36).

📍 src/debugpy/launcher/debuggee.py:31

The original `preexec_fn` does two things: `os.setpgrp()` (own process group, required so the launcher's `os.killpg(process.pid, ...)` targets the whole tree) **and** `tcsetpgrp(...)` (make that group the terminal foreground). `process_group=0` faithfully replaces only the first — the kill-tree invariant is preserved, but the foreground-terminal handling is silently dropped for GraalPy. Interactive terminal I/O on a real TTY may differ. At minimum, expand the comment to state the foreground-group behavior is intentionally not replicated and why.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/launcher/debuggee.py:L36](https://github.com/microsoft/debugpy/blob/81d199b958126344bd74dbb36325c79772f0270a/src/debugpy/launcher/debuggee.py#L36).

📍 src/debugpy/launcher/debuggee.py:33

`process_group=0` is a `Popen` keyword added in CPython 3.11. On a GraalPy release whose stdlib tracks 3.8/3.10, `Popen(..., process_group=0)` raises `TypeError: unexpected keyword argument`, which would make launching any debuggee impossible — the opposite of this PR's goal. Consider feature-detecting (`"process_group" in inspect.signature(subprocess.Popen).parameters`) with a fallback, or documenting the minimum supported GraalPy version.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/server/api.py:L55](https://github.com/microsoft/debugpy/blob/81d199b958126344bd74dbb36325c79772f0270a/src/debugpy/server/api.py#L55).

📍 src/debugpy/server/api.py:49

This `hasattr(os, "fork")` is evaluated in the **server** interpreter, but the process that actually decides to fork/daemonize is the **adapter** (spawned separately, with a potentially user-configured `python`). In a CPython-server + GraalPy-adapter combination the server sees `fork`, calls `_adapter_process.wait()`, but the adapter never forks/exits, so `listen()` can hang. Consider basing this decision on whether the *adapter* will daemonize rather than on the server's own `os.fork`.

heejaechang:
Looks good overall and the approach is reasonable. The main theme of the feedback: the fork-less runtime is detected two different ways (`sys.implementation.name == 'graalpy'` vs `hasattr(os, 'fork')`) — worth unifying behind one predicate — and please confirm `process_group=0` is available on your minimum supported GraalPy (it's a CPython 3.11+ Popen kwarg) and that dropping the `tcsetpgrp` foreground handling is acceptable.

heejaechang (approved):
Approved via Review Center.

rchiodo:
@steve-s you need to sign the CLA in order for us to merge this. Otherwise everything else looks good to me.

src/debugpy/launcher/debuggee.py
```diff
@@ -58,9 +58,12 @@ def spawn(process_name, cmdline, env, redirect_output):
         else:
             kwargs = {}
 
-        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
-            # GraalPy does not support running code between fork and exec
+        # GraalPy does not support running code between fork and exec, but supports the
+        # process_group argument for Popen
```
heejaechang, line 62: 📍 src/debugpy/launcher/debuggee.py:62
The `process_group` argument to `subprocess.Popen` was only added in Python 3.11. On GraalPy builds targeting CPython ≤ 3.10 (e.g. GraalPy 23.x / 24.0), `kwargs.update(process_group=0)` causes `Popen` to raise `TypeError`, which is caught at the spawn site and re-raised as `MessageHandlingError` — so **every debuggee launch fails**. This is a regression: pre-PR, GraalPy hit neither branch and spawned fine (only `kill()` was unreliable). Guard the kwarg, e.g. `if sys.version_info >= (3, 11): kwargs.update(process_group=0)` (or check `inspect.signature(subprocess.Popen).parameters`), with a documented fallback for older runtimes.

src/debugpy/launcher/debuggee.py
heejaechang, line 63: 📍 src/debugpy/launcher/debuggee.py:60
`process_group=0` reproduces only the `os.setpgrp()` half of the original `preexec_fn`; it does not replicate the `os.tcsetpgrp()` call that makes the debuggee the foreground process group of the controlling terminal. Interactive GraalPy debuggees that read stdin or use job control may behave differently (e.g. SIGTTIN/SIGTTOU on terminal access). This is likely acceptable since the block is best-effort, but add a one-line comment noting the lost foreground-terminal behavior is a known, intentional limitation.

heejaechang:
Main concern: the `process_group=0` kwarg needs a version/availability guard so older GraalPy (pre-3.11) runtimes don't break on every launch. One secondary note about lost foreground-terminal (tcsetpgrp) behavior is worth documenting.

heejaechang:
Overall the GraalPy compatibility approach is reasonable, but please guard the `process_group=0` usage for Python versions where `subprocess.Popen` actually supports it (3.11+) so older GraalPy/CPython-3.10-based builds don't fail to spawn debuggees.

steve-s:
Thank you for the reviews! I'll get back to this once I figure out the CLA with my company.

rchiodo (approved):
Approved via Review Center.

heejaechang:
Solid GraalPy compatibility work overall. The main blocker is the unconditional `process_group=0` kwarg, which is only valid on Python 3.11+; please guard it so older GraalPy runtimes don't regress.

heejaechang:

src/debugpy/launcher/debuggee.py
heejaechang, line 62: `process_group=0` requires the `subprocess.Popen(process_group=...)` parameter, which only exists in Python 3.11+. GraalPy has shipped 3.8/3.10-compatible releases; on such a build this raises `TypeError`, and `spawn()` re-raises it as `MessageHandlingError`, so every launch fails — converting the fork fix into a spawn crash. Guard it (e.g. check `"process_group" in inspect.signature(subprocess.Popen).parameters`) or wrap in try/except with a graceful fallback. At minimum confirm the minimum supported GraalPy exposes `process_group`.

src/debugpy/adapter/__main__.py
heejaechang, line 40: Only `os.fork` was capability-guarded, but a fresh adapter subprocess is normally not a session leader, so `os.getsid(os.getpid())` and `os.setsid()` execute *before* the new `hasattr(os, "fork")` check is reached. If GraalPy omits `setsid`/`getsid` (the same class of low-level POSIX primitive that motivated the fork guard), this raises `AttributeError` and the adapter dies before daemonizing. Consider guarding consistently, e.g. `if hasattr(os, "setsid") and os.getsid(...) != ...`.

src/debugpy/launcher/debuggee.py
```diff
@@ -58,9 +58,12 @@ def spawn(process_name, cmdline, env, redirect_output):
...
+        if sys.platform != "win32" and sys.implementation.name == "graalpy":
+            kwargs.update(process_group=0)
 
+        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
```
heejaechang, line 66: The two `if sys.platform != "win32" and sys.implementation.name == / != "graalpy"` blocks are mutually-exclusive arms of one decision, but written as independent top-level `if`s they hide that relationship and re-evaluate `sys.platform != "win32"` twice. Consider collapsing into a single `if sys.platform != "win32":` with an inner `if/else` on the implementation, so the either/or intent is visible and a future edit to one arm prompts an edit to the other.

heejaechang:
Solid GraalPy compatibility fix. Two things worth a look before merge: confirm the minimum supported GraalPy exposes `subprocess.Popen(process_group=...)` (3.11+), and consider guarding `setsid`/`getsid` as consistently as `fork`.

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:

heejaechang (approved):
Approved via Review Center.

src/debugpy/launcher/debuggee.py
rchiodo, line 63: Released GraalPy 23.1.3 and 24.0.2 reject `Popen(process_group=...)` with `TypeError`, preventing debuggee launch. Capability-check this argument and preserve the existing fallback and termination behavior for unsupported releases.

src/debugpy/server/api.py
```diff
@@ -218,9 +218,11 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
                 creationflags=creationflags,
                 env=python_env,
             )
-            if os.name == "posix":
+            if os.name == "posix" and hasattr(os, "fork"):
```
rchiodo, line 221: This checks the server interpreter's `os.fork`, but `_config["python"]` can launch the adapter under a different runtime. A CPython parent launching a GraalPy adapter will still call `wait()`, even though that adapter cannot fork and remains running, so startup can block indefinitely. Determine this behavior from the adapter runtime or otherwise handle the mixed-interpreter path.

StellaHuang95:

StellaHuang95 (approved):
Approved via Review Center.

src/debugpy/launcher/debuggee.py
```diff
@@ -58,9 +58,12 @@ def spawn(process_name, cmdline, env, redirect_output):
         else:
             kwargs = {}
 
-        if sys.platform != "win32" and sys.implementation.name != 'graalpy':
-            # GraalPy does not support running code between fork and exec
+        # GraalPy does not support running code between fork and exec, but supports the
+        # process_group argument for Popen
+        if sys.platform != "win32" and sys.implementation.name == "graalpy":
+            kwargs.update(process_group=0)
```
StellaHuang95, line 64: > **Warning** · Non-blocking recommendation

`process_group=0` replaces only the new-process-group portion of the CPython `preexec_fn`; it does not perform the foreground-terminal handoff. Please clarify that terminal foreground/job-control behavior is unsupported on GraalPy so this is not mistaken for full parity.

StellaHuang95, line 64: > **Warning** · Non-blocking recommendation

`process_group` requires Python 3.11+. A GraalPy build targeting an older Python version will fail every launch with an unsupported Popen keyword. Please guard this path or document the required GraalPy/Python version.

src/debugpy/server/api.py
```diff
@@ -218,9 +218,11 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
...
                 # It's going to fork again to daemonize, so we need to wait on it to
-                # clean it up properly.
+                # clean it up properly. If we did not fork, we cannot take this path
+                # because it cannot perform that extra fork; waiting there would just
+                # preserve the broken assumption that a daemonized grandchild exists.
```
StellaHuang95, line 225: > **Warning** · Non-blocking recommendation

This wait guard must remain synchronized with the adapter daemonization behavior in `adapter/__main__.py`: waiting is correct only when the adapter forks and its spawned parent exits. Please add a cross-reference so future changes do not introduce a hang or unreaped child.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

src/debugpy/server/api.py
StellaHuang95, line 225: > **Warning** · Non-blocking recommendation

Please add regression coverage for the forkless POSIX path, verifying that `listen()` reaches endpoint acceptance without blocking when the adapter cannot daemonize via `fork()`.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

bschnurr (approved):
Approved via Review Center.

steve-s:
I am closing this PR until I figure out the CLA signature.

PR #2041, opened 2026-06-24
Add Dependabot configuration for pip and GitHub Actions

rchiodo:
## Summary

Adds a `.github/dependabot.yml` to enable automated dependency and security (CVE) update pull requests.

This addresses the recurring Component Governance / CVE alerts for the repository by letting Dependabot keep dependencies current and raise security update PRs automatically.

## Ecosystems covered

| Ecosystem | Path | Covers |
|-----------|------|--------|
| `pip` | `/tests` | `tests/requirements.txt` (django, flask, gevent, numpy, requests, etc.) |
| `github-actions` | `/` | Actions used by the workflows (e.g. `actions/checkout`, `github/codeql-action`, `actions/stale`, `actions/github-script`) |

- Weekly schedule, grouped PRs to reduce noise, labels, and a 10 open-PR limit per ecosystem.

## Notes

- This file enables **version updates**. For Dependabot to also open PRs specifically for existing CVE alerts, **Dependabot security updates** must be enabled in repo Settings -> Code security.
- Vendored dependencies under `src/debugpy/_vendored/` and the git components in `cgmanifest.json` (pydevd, bytecode, winappdbg, scandir) are **not** managed by Dependabot and still require manual updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,38 @@
+# Dependabot configuration for debugpy.
+#
+# Keeps dependencies up to date and automatically opens pull requests for
+# security (CVE) and version updates. Security updates are always enabled for
+# every ecosystem Dependabot can detect; the entries below also enable regular
+# version updates so dependencies don't drift far enough to accumulate
+# vulnerabilities in the first place.
+#
+# Docs: https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+version: 2
+updates:
+  # Python packages used to run and exercise the test suite.
+  - package-ecosystem: "pip"
+    directory: "/tests"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "python"
+    groups:
+      python-test-dependencies:
+        patterns:
+          - "*"
+
+  # GitHub Actions used by the workflows in .github/workflows.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "github-actions"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,13 @@
 # Dependabot configuration for debugpy.
 #
-# Keeps dependencies up to date and automatically opens pull requests for
-# security (CVE) and version updates. Security updates are always enabled for
-# every ecosystem Dependabot can detect; the entries below also enable regular
-# version updates so dependencies don't drift far enough to accumulate
+# This file configures version updates: Dependabot opens pull requests to keep
+# dependencies current so they don't drift far enough to accumulate
 # vulnerabilities in the first place.
+#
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,38 @@
+# Dependabot configuration for debugpy.
+#
+# Keeps dependencies up to date and automatically opens pull requests for
+# security (CVE) and version updates. Security updates are always enabled for
+# every ecosystem Dependabot can detect; the entries below also enable regular
+# version updates so dependencies don't drift far enough to accumulate
```
heejaechang: The header comment states "Security updates are always enabled for every ecosystem Dependabot can detect," but this file only configures **version updates**. Dependabot security (CVE) updates require the separate toggle in repo Settings → Code security, as the PR description correctly notes. Consider softening this committed comment to match that caveat so a future maintainer doesn't assume CVE PRs are already flowing.

.github/dependabot.yml
```diff
@@ -0,0 +1,38 @@
...
+updates:
+  # Python packages used to run and exercise the test suite.
+  - package-ecosystem: "pip"
+    directory: "/tests"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "python"
```
heejaechang: The custom labels `python` and `github-actions` will be silently ignored by Dependabot if they don't already exist in the repo (only `dependencies` is auto-created). Either pre-create these labels or drop them to avoid a no-op.

heejaechang:
Config looks sound and safe to merge. A couple of minor, non-blocking notes inline (misleading header comment about security updates, and custom labels that may need to be pre-created).

heejaechang (approved):
Approved via Review Center.

.github/dependabot.yml
rchiodo: Good catch - softened the header comment in 8d24d2cf to make clear this file only configures version updates, and that CVE security updates require the separate Settings -> Code security toggle.

.github/dependabot.yml
rchiodo: Dropped the custom 'python' and 'github-actions' labels in 8d24d2cf, keeping only the auto-created 'dependencies' label to avoid the silent no-op.

.github/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+updates:
+  # Python packages used to run and exercise the test suite.
+  - package-ecosystem: "pip"
+    directory: "/tests"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+    groups:
```
heejaechang, line 23: The `patterns: ["*"]` grouping collapses every dependency into one all-or-nothing grouped PR. Given the PR's stated CVE-remediation goal, a single breaking major bump (e.g. `numpy`, `gevent`, `django`) can wedge the entire group and delay a security-relevant bump (e.g. `requests`) swept into it. Consider carving high-risk majors out of the catch-all group, or rely on Dependabot's ungrouped security PRs once that setting is enabled. Not a blocker.

heejaechang:
Config looks sound and safe to merge. One optional note about the catch-all grouping is worth considering, but it's non-blocking.

heejaechang (approved):
Approved via Review Center.

PR #2045, opened 2026-06-24
Pin secure test dependency floors and drop EOL Python 3.9/3.8

rchiodo:
## Summary

Supersedes #2043. Adds the secure minimum-version floors for test dependencies that Component Governance flags as vulnerable, and drops Python 3.9/3.8 (both EOL) so those floors are installable.

The pinned versions in #2043 — `pytest>=9.0.3`, `requests>=2.33.0`, `urllib3>=2.7.0` — all require **Python >=3.10**, so on the 3.9 CI legs `pip install -r tests/requirements.txt` failed with `No matching distribution found for pytest>=9.0.3` (every 9.x release is `Requires-Python >=3.10`). Since Python 3.9 (and 3.8) are past end-of-life, the right fix is to stop testing/claiming support for them rather than weaken the security floors.

## CVEs addressed

| CVE | Package | Severity | Vulnerable | Fixed in |
|-----|---------|----------|-----------|----------|
| CVE-2026-44431 / CVE-2026-44432 | urllib3 | High | 2.6.0 – 2.6.x | 2.7.0 |
| CVE-2026-25645 | requests | Medium | < 2.33.0 | 2.33.0 |
| CVE-2025-71176 | pytest | Medium | <= 9.0.2 | 9.0.3 |

## Changes

- `tests/requirements.txt`: `pytest>=9.0.3`, `requests>=2.33.0`, `urllib3>=2.7.0` (explicit floor since it's pulled in transitively by `requests`).
- `azure-pipelines/pipelines.yaml`: remove `py39` from the Linux/macOS/Windows test matrices, remove the `py39_32` (32-bit) leg, and bump the Lint stage to Python 3.10.
- `tox.ini`: drop `py38`/`py39` from `envlist` and the per-env commands.
- `setup.py`: `python_requires=">=3.10"` and remove the 3.8/3.9 classifiers.

## Notes

- 32-bit Windows testing is dropped (it only existed on the removed 3.9 leg).
- The vendored components under `src/debugpy/_vendored/pydevd/` are out of scope here.

### DIFF
--- a/azure-pipelines/pipelines.yaml
+++ b/azure-pipelines/pipelines.yaml
@@ -213,9 +213,6 @@ stages:
         matrix:
           py310:
             python.version: 3.10
-          py310_32:
-            python.version: 3.10
-            architecture: x86
           py311:
             python.version: 3.11
           py312:
--- a/azure-pipelines/pipelines.yaml
+++ b/azure-pipelines/pipelines.yaml
@@ -213,6 +213,9 @@ stages:
         matrix:
           py310:
             python.version: 3.10
+          py310_32:
+            python.version: 3.10
+            architecture: x86
           py311:
             python.version: 3.11
           py312:
--- a/azure-pipelines/pipelines.yaml
+++ b/azure-pipelines/pipelines.yaml
@@ -29,7 +29,7 @@ stages:
           vmImage: ubuntu-latest
 
         variables:
-          python.version: "3.9"
+          python.version: "3.10"
 
         steps:
 
@@ -125,8 +125,6 @@ stages:
 
       strategy:
         matrix:
-          py39:
-            python.version: 3.9
           py310:
             python.version: 3.10
           py311:
@@ -170,8 +168,6 @@ stages:
 
       strategy:
         matrix:
-          py39:
-            python.version: 3.9
           py310:
             python.version: 3.10
           py311:
@@ -215,11 +211,6 @@ stages:
 
       strategy:
         matrix:
-          py39:
-            python.version: 3.9
-          py39_32:
-            python.version: 3.9
-            architecture: x86
           py310:
             python.version: 3.10
           py311:
--- a/setup.py
+++ b/setup.py
@@ -160,11 +160,9 @@ if __name__ == "__main__":
         project_urls={
             "Source": "https://github.com/microsoft/debugpy",
         },
-        python_requires=">=3.8",
+        python_requires=">=3.10",
         classifiers=[
             "Development Status :: 5 - Production/Stable",
-            "Programming Language :: Python :: 3.8",
-            "Programming Language :: Python :: 3.9",
             "Programming Language :: Python :: 3.10",
             "Programming Language :: Python :: 3.11",
             "Programming Language :: Python :: 3.12",
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,6 +1,6 @@
 ## Used to run the tests:
 
-pytest
+pytest>=9.0.3  # CVE-2025-71176 (vulnerable <= 9.0.2)
 pytest-xdist
 pytest-cov
 pytest-timeout
@@ -18,7 +18,10 @@ django
 flask
 gevent
 numpy
-requests
+requests>=2.33.0  # CVE-2026-25645 (vulnerable < 2.33.0)
+# urllib3 is pulled in transitively by requests; pin a secure floor for
+# CVE-2026-44431 and CVE-2026-44432 (vulnerable 2.6.0 <= x < 2.7.0).
+urllib3>=2.7.0
 typing_extensions
 
 # Used to build pydevd attach to process binaries:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ exclude = '''
 '''
 
 [tool.pyright]
-pythonVersion = "3.8"
+pythonVersion = "3.10"
 include = ["src/**", "tests/**" ]
 extraPaths = ["src/debugpy/_vendored/pydevd", "src/debugpy/_vendored/pydevd/pydevd_attach_to_process"]
 ignore = ["src/debugpy/_vendored/pydevd", "src/debugpy/_version.py"]
@@ -71,8 +71,8 @@ line-length = 88
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
-# Assume Python 3.8
-target-version = "py38"
+# Assume Python 3.10
+target-version = "py310"
 
 [tool.ruff.per-file-ignores]
 "tests/debugpy/test_breakpoints.py" = ["F841"]
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{38,39,310,311,312,313,314}{,-cov}
+envlist = py{310,311,312,313,314}{,-cov}
 
 [testenv]
 deps = -rtests/requirements.txt
@@ -8,7 +8,5 @@ setenv =
     DEBUGPY_TEST=1
 commands_pre = python build_attach_binaries.py
 commands =
-    py{38,39}-!cov: python -m pytest {posargs}
-    py{38,39}-cov: python -m pytest --cov --cov-append --cov-config=.coveragerc {posargs}
     py{310,311,312,313,314}-!cov: python -Xfrozen_modules=off -m pytest {posargs}
     py{310,311,312,313,314}-cov: python -Xfrozen_modules=off -m pytest --cov --cov-append --cov-config=.coveragerc {posargs}


heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [azure-pipelines/pipelines.yaml:L219](https://github.com/microsoft/debugpy/blob/66a7d965e3b581917a85dac1e0d4b92b0c88f060/azure-pipelines/pipelines.yaml#L219).

Removing `py39_32` eliminates the only x86 CI leg, but `setup.py` still packages `win32` 32-bit wheels and native pydevd attach binaries. That leaves a shipped 32-bit target with zero 32-bit signal, so architecture-specific regressions (pointer width, ctypes struct layout, calling convention) can ship undetected. Dropping the EOL interpreter and dropping the sole x86 leg are independent decisions: either re-add one x86 leg on a supported interpreter (e.g. `py310_32: { python.version: 3.10, architecture: x86 }`) or explicitly confirm in the PR that 32-bit Windows wheels are being discontinued in lockstep.

heejaechang (approved):
Approved via Review Center.

rchiodo:
Fixed in 804b4639 — restored a 32-bit Windows CI leg on a supported interpreter (\`py310_32: { python.version: 3.10, architecture: x86 }\`) in the \`Test_Windows\` matrix.

azure-pipelines/pipelines.yaml
```diff
@@ -215,13 +211,11 @@ stages:
...
-            python.version: 3.9
-            architecture: x86
           py310:
             python.version: 3.10
+          py310_32:
+            python.version: 3.10
+            architecture: x86
           py311:
             python.version: 3.11
           py312:
```
heejaechang, line 218: The re-added `py310_32` (x86) leg runs `pip install -r tests/requirements.txt` on win32 CPython 3.10. Unpinned C-extension deps (`numpy`, `gevent`, `Cython`) have spotty 32-bit Windows wheel coverage — numpy 2.x dropped win32 wheels — so a missing cp310-win32 wheel could trigger a source build and fail the entire leg at install, the same failure class that motivated this PR. Smoke-install the file on win32 CPython 3.10 before merge.

tests/requirements.txt
```diff
@@ -1,6 +1,6 @@
 ## Used to run the tests:
 
-pytest
+pytest>=9.0.3  # CVE-2025-71176 (vulnerable <= 9.0.2)
 pytest-xdist
 pytest-cov
 pytest-timeout
```
heejaechang, line 6: `pytest>=9.0.3` forces a major 8→9 jump while the plugins (`pytest-xdist`, `pytest-cov`, `pytest-timeout`, `pytest-retry`) remain unpinned. A plugin that hasn't yet declared pytest 9 support could break the whole matrix at collection time. Please confirm a clean `pip install -r tests/requirements.txt` + `pytest --collect-only` on 3.10 actually resolves and collects against 9.x.

heejaechang:
Approving — the security floors and EOL-interpreter removal are correct. Two follow-ups worth checking before merge: (1) verify the dep set actually installs/collects on the re-added win32 CPython 3.10 leg, and (2) the PR description says "32-bit Windows testing is dropped," but the diff actually re-adds it as `py310_32` (architecture: x86) — reconcile the description with the diff.

heejaechang:
Change looks sound. Main residual risk is install-time fragility on the newly-added 32-bit 3.10 leg and the major pytest 8→9 bump with unpinned plugins — worth a quick smoke-install/collect before merge, but not blocking.

heejaechang (approved):
Approved via Review Center.

azure-pipelines/pipelines.yaml
rchiodo: @C:\Users\rchiodo\AppData\Local\Temp\pr-watch-clones\rchiodo-debugpy-pr-2045\.generated\pr-feedback\reply1.md

tests/requirements.txt
rchiodo: @C:\Users\rchiodo\AppData\Local\Temp\pr-watch-clones\rchiodo-debugpy-pr-2045\.generated\pr-feedback\reply2.md

azure-pipelines/pipelines.yaml
```diff
@@ -215,11 +211,6 @@ stages:
...
       strategy:
         matrix:
-          py39:
-            python.version: 3.9
-          py39_32:
-            python.version: 3.9
-            architecture: x86
           py310:
             python.version: 3.10
           py311:
```
heejaechang, line 216: 📍 azure-pipelines/pipelines.yaml:211
This revision removes `py39` and `py39_32` and adds nothing in their place, so the Windows matrix is now `py310`–`py314` with zero 32-bit legs — yet `setup.py:is_applicable` still packages `win32` artifacts (`return plat in ("win32", "win-amd64")`). debugpy continues to ship a `win32` wheel surface with no x86 CI signal. Either restore a single `py310_32` (`architecture: x86`) leg — and smoke-install `tests/requirements.txt` on win32 cp310 to confirm the unpinned C-extension deps resolve — or remove `win32` from `setup.py:is_applicable` so packaging and CI agree. Do not merge with a shipped-but-untested architecture.

setup.py
```diff
@@ -160,11 +160,9 @@ def tail_is(*suffixes):
         project_urls={
             "Source": "https://github.com/microsoft/debugpy",
         },
-        python_requires=">=3.8",
+        python_requires=">=3.10",
```
heejaechang, line 163: 📍 setup.py:163
This PR converges `python_requires`, the classifiers, `tox.ini` envlist, and the CI matrices on 3.10, but `pyproject.toml` was left untouched: `[tool.pyright] pythonVersion = "3.8"` and `[tool.ruff] target-version = "py38"`. The Lint stage this PR bumped to 3.10 will still run `ruff check` against a `py38` target the PR claims is no longer supported, and pyright validates 3.8 semantics. Bump both to `3.10`/`py310` (and update the `# Assume Python 3.8` comment) in this PR so all declaration sites agree.

heejaechang:
Solid security-floor + EOL-cleanup change overall. One blocking item: reconcile dropping the only 32-bit Windows test leg with still shipping win32 wheels (restore a 32-bit leg on a supported Python, or stop packaging win32). Also align the remaining py38 static-analysis declarations with the new >=3.10 floor.

setup.py
rchiodo: Fixed in e87f214e. Bumped `[tool.pyright] pythonVersion` `3.8` → `3.10` and `[tool.ruff] target-version` `py38` → `py310` (and updated the `# Assume Python 3.8` comment to `3.10`). All declaration sites — `python_requires`, classifiers, `tox.ini` envlist, the CI matrices, and now pyproject — agree on the 3.10 floor, so the Lint stage's `ruff check` and pyright both validate against the supported version. (Remaining `3.8` references in the tree are all under the upstream-owned vendored `pydevd` subtree, which is out of scope for this PR.)

azure-pipelines/pipelines.yaml
rchiodo: Thanks — this is a real tension, but I don't think either literal option is the right change to make inside this PR, so I'm flagging it for a maintainer call rather than forcing it here. Reasoning, with evidence:

**Option A — restore a `py310_32` (x86) leg:** This re-introduces the exact failure class this PR exists to remove. As established in the (now-resolved) thread above, `gevent` ships **no** cp310-win32 wheel (only `...-cp310-cp310-win_amd64.whl`), so `pip install -r tests/requirements.txt` on win32 cp310 falls back to a source build of gevent (greenlet/libev/c-ares) and risks failing the leg at install. There's no PEP 508 marker that reliably distinguishes a 32-bit interpreter on 64-bit Windows (`platform_machine` reflects the OS, not the interpreter), so gevent can't be cleanly gated off only the x86 leg without either still source-building or silently dropping gevent coverage. Adding the leg back is therefore a coverage/risk tradeoff that you flagged yourself — a maintainer decision, not something I should force in autonomously.

**Option B — remove `win32` from `setup.py:is_applicable`:** This doesn't actually stop debugpy from shipping a win32 wheel — `is_applicable` only filters which vendored pydevd native binaries (`.exe/.dll/.pdb/.pyd`) get copied into the wheel. Dropping `win32` there would still build a win32 wheel, just a **broken** one missing the attach-to-process binaries. Genuinely retiring the win32 wheel surface means changing the build/packaging matrix and would break existing 32-bit Windows users — clearly out of scope for "drop EOL 3.9/3.8 + pin secure dep floors."

Note this PR didn't newly create the gap: the only prior 32-bit leg was `py39_32`, removed here solely because Python 3.9 is EOL. The win32 *packaging* surface is unchanged.

My recommendation: keep win32 packaging as-is in this PR and track "restore an x86 test signal (wheel-only deps, gevent excluded on x86)" or "retire the win32 wheel" as a separate, deliberate follow-up. Happy to implement whichever direction you choose — leaving this thread open for your call.

heejaechang, line 216: This revision removes both `py39` and `py39_32` — the base branch's only `architecture: x86` leg — leaving the Windows matrix at `py310`–`py314` under the global `architecture: x64`. Meanwhile `setup.py` (`is_applicable`) still ships win32 wheels (`plat in ("win32", "win-amd64")`), so debugpy continues to package an x86 surface with zero x86 CI signal: a win32-only regression in the attach binaries could ship undetected. Consider either restoring one `py310` `architecture: x86` leg (at minimum smoke-installing `tests/requirements.txt` on win32 cp310), or narrowing `setup.py:is_applicable` to drop `win32` so packaging and CI coverage agree.

heejaechang:
Security floors and the 3.8/3.9 removal look consistent across setup.py, pyproject.toml, tox.ini, and the CI matrices. Two non-blocking things worth a look before merge: (1) dropping the only x86 leg while still shipping win32 wheels leaves that surface without CI signal, and (2) the pytest 8→9 major bump with floor-less plugins — a quick `pip install -r tests/requirements.txt` + `pytest --collect-only` on 3.10 would confirm the matrix still resolves.

heejaechang (approved):
Approved via Review Center.

azure-pipelines/pipelines.yaml
rchiodo: Thanks — this is the same x86-CI vs. win32-packaging tension raised (and discussed in depth) in the thread above, so I'll consolidate here and keep it open for a maintainer call rather than force a change in this PR.

Both options you list have been evaluated, with evidence:

- **Restore a `py310` x86 leg.** I actually tried this in this PR (`804b4639` "ci: restore x86 Windows leg on Python 3.10") and had to revert it (`d2c3ac40` "ci: drop x86 Windows leg (gevent has no cp310-win32 wheel)"). `gevent` publishes no cp310-win32 wheel, so `pip install -r tests/requirements.txt` on win32 cp310 falls back to a source build (greenlet/libev/c-ares) and risks failing the leg at install — the exact failure class this PR exists to remove. There's also no PEP 508 marker that reliably distinguishes a 32-bit interpreter on 64-bit Windows (`platform_machine` reflects the OS, not the interpreter), so gevent can't be cleanly excluded on just the x86 leg.

- **Drop `win32` from `setup.py:is_applicable`.** `is_applicable` only filters which vendored pydevd native binaries get copied into the wheel; it does not stop a win32 wheel from being built. Removing `win32` there would still ship a win32 wheel — just a broken one missing the attach-to-process binaries. Genuinely retiring the win32 wheel surface is a packaging-matrix change that would break existing 32-bit users, which is out of scope for "drop EOL 3.9/3.8 + pin secure dep floors."

For context, this PR didn't create the gap: the only prior x86 leg was `py39_32`, removed here solely because Python 3.9 is EOL; the win32 *packaging* surface is unchanged.

Recommendation: keep win32 packaging as-is in this PR (which you've approved) and track "restore an x86 test signal with wheel-only deps (gevent excluded on x86)" or "retire the win32 wheel surface" as a separate, deliberate follow-up. Happy to implement whichever direction you choose — leaving this open for your call.

rchiodo: Thanks — this is the same x86-CI vs. win32-packaging tension raised (and discussed in depth) in the thread above, so I'll consolidate here and keep it open for a maintainer call rather than force a change in this PR.

Both options you list have been evaluated, with evidence:

- **Restore a `py310` x86 leg.** I actually tried this in this PR (`804b4639` "ci: restore x86 Windows leg on Python 3.10") and had to revert it (`d2c3ac40` "ci: drop x86 Windows leg (gevent has no cp310-win32 wheel)"). `gevent` publishes no cp310-win32 wheel, so `pip install -r tests/requirements.txt` on win32 cp310 falls back to a source build (greenlet/libev/c-ares) and risks failing the leg at install — the exact failure class this PR exists to remove. There's also no PEP 508 marker that reliably distinguishes a 32-bit interpreter on 64-bit Windows (`platform_machine` reflects the OS, not the interpreter), so gevent can't be cleanly excluded on just the x86 leg.

- **Drop `win32` from `setup.py:is_applicable`.** `is_applicable` only filters which vendored pydevd native binaries get copied into the wheel; it does not stop a win32 wheel from being built. Removing `win32` there would still ship a win32 wheel — just a broken one missing the attach-to-process binaries. Genuinely retiring the win32 wheel surface is a packaging-matrix change that would break existing 32-bit users, which is out of scope for "drop EOL 3.9/3.8 + pin secure dep floors."

For context, this PR didn't create the gap: the only prior x86 leg was `py39_32`, removed here solely because Python 3.9 is EOL; the win32 *packaging* surface is unchanged.

Recommendation: keep win32 packaging as-is in this PR (which you've approved) and track "restore an x86 test signal with wheel-only deps (gevent excluded on x86)" or "retire the win32 wheel surface" as a separate, deliberate follow-up. Happy to implement whichever direction you choose — leaving this open for your call.

PR #2049, opened 2026-07-23
Fix race between adapter connection publication and session attachment

aperez:
`Connection.__init__` publishes a new connection to `_connections` under `_lock`, then releases the lock before classifying it. In that window another thread can attach the connection to a session (via an `attach` request or the launch flow), which sets `conn.server` and the session's `pid`. When the constructor resumes, `sessions.get(self.pid)` returns that just-attached session, and because it is not a process replacement the old code logged "is not expecting replacement" and closed the channel, dropping a session that was already live. This shows up with subprocess debugging and in-place process replacement.

Fix it by re-checking under `_lock` whether the connection was already attached (`self.server is not None`) before treating it as an unexpected replacement. Closing the channel under `_lock` is deadlock-free: `JsonMessageChannel.close()` only takes the channel's own lock and the disconnect handler runs on the parser thread.

Add a regression test that overlaps publication and attachment.

### DIFF
--- a/tests/debugpy/adapter/test_servers.py
+++ b/tests/debugpy/adapter/test_servers.py
@@ -1,0 +1,159 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root
+# for license information.
+
+import threading
+
+from debugpy import adapter
+from debugpy.adapter import servers, sessions
+
+
+class FakeMessage:
+    def __init__(self, values):
+        self.values = values
+
+    def __call__(self, name, _validator, optional=False):
+        if optional and name not in self.values:
+            return ()
+        return self.values[name]
+
+
+class FakeStream:
+    name = "fake stream"
+
+
+class FakeChannel:
+    def __init__(self, stream, handlers, pid):
+        self.stream = stream
+        self.handlers = handlers
+        self.pid = pid
+        self.name = "fake channel"
+        self.closed = False
+
+    def start(self):
+        pass
+
+    def request(self, command, arguments=None):
+        assert command == "pydevdSystemInfo"
+        return FakeMessage(
+            {
+                "process": FakeMessage(
+                    {
+                        "pid": self.pid,
+                    }
+                )
+            }
+        )
+
+    def close(self):
+        self.closed = True
+
+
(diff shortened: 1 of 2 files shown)


aperez:
@microsoft-github-policy-service agree company="Meta"

rchiodo:

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

tests/debugpy/adapter/test_servers.py
```diff
@@ -0,0 +1,159 @@
...
+    connection_thread = threading.Thread(
+        target=create_connection,
+        name="connection",
+        daemon=True,
+    )
+    attach_thread.start()
+    connection_thread.start()
+    try:
+        assert attachment_started.wait(5)
+        assert classification_waiting.wait(5)
```
rchiodo: This test feels a bit over-coupled to the implementation. The `ObservedLock` + `classification_waiting` scaffolding asserts that the fix re-acquires `_lock` *specifically during classification* — a correct-but-different fix (e.g. snapshotting `self.server` a different way) would fail here even though the behavior is fine. The real value is in the final behavioral checks (`connections[0].server is attached_session.server` and `not channel.closed`). Would it make sense to do that instead?

aperez: Thanks for the feedback! 
I rewrote the test to be behavioral: it now drives the real attach path (`wait_for_connection` → `attach_to_session`) and asserts only outcomes, `connections[0].server is session.server` and `not channel.closed`. I dropped the `ObservedLock`/`classification_waiting` scaffolding entirely. The test still fails against the pre-fix code (channel gets closed).

rchiodo (approved):

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

aperez:
@rchiodo @heejaechang the Windows py312 failure looks to be unrelated to the changes in this PR

rchiodo:
> @rchiodo @heejaechang the Windows py312 failure looks to be unrelated to the changes in this PR

Yeah sorry. The test runner is flakey when shutting down tests sometimes. PR looks good. Thanks for the submission.

PR #2050, opened 2026-07-30
Respect isolated mode when patching sys.path

### LINKED ISSUE #1916: debugpy ignores isolated mode (-I)

## Environment data

- debugpy version: 1.8.14
- OS and version: Debian 12, Mac OS 15.5
- Python version (& distribution if applicable, e.g. Anaconda): python3.9, python3.11

## Actual behavior

sys.path contains current script directory in [isolated mode](https://docs.python.org/3/using/cmdline.html#cmdoption-I) (-I) with debugpy. 

Relevant code is in [cli.py](https://github.com/microsoft/debugpy/blob/99b202ee21441f844b4c8fa3f79d518834fa5ff5/src/debugpy/server/cli.py#L338-L345), looks like debugpy should check for isolated mode before doing `sys.path.insert(0, dir)`

## Expected  behavior

debugpy respects isolated mode and sys.path under debugpy running in isolated mode doesn't contain current script directory

## Steps to reproduce:

sys.path contains directory of the script being run in isolated mode with debugpy:

```shell
python -I -m debugpy --listen 5678 -c 'import sys; print(f"{sys.path[0]=!r}")'
sys.path[0]=''
```

sys.path doesn't contain script directory in isolated mode without debugpy:
```shell
python -I -c 'import sys; print(f"{sys.path[0]=!r}")'
sys.path[0]='/opt/homebrew/Cellar/python@3.9/3.9.22/Frameworks/Python.framework/Versions/3.9/lib/python39.zip'
```

karandhaodiyal28-hash:
## Fixes #1916

Python's isolated mode (`-I`) must not prepend the script directory (for a file target) or the current directory (for `-m` / `-c`) to `sys.path` — that's the whole point of the flag, and it exists to prevent local modules from shadowing stdlib/third-party imports. debugpy re-implements Python's startup `sys.path` handling in `src/debugpy/server/cli.py`, but did so unconditionally, so a debuggee launched under `-I` still got the directory injected:

```console
$ python -I -m debugpy --listen 5678 -c "import sys; print(repr(sys.path[0]))"
''                                          # cwd injected - isolated mode violated

$ python -I -c "import sys; print(repr(sys.path[0]))"
'/usr/lib/python3.11'                       # plain Python, correct
```

The comments in those three functions already say "like Python itself does for -m/-c" — but Python itself skips the insert under `-I`. This guards all three sites (`run_file`, `run_module`, `run_code`) with `sys.flags.isolated`, matching CPython semantics exactly. `sys.flags.isolated` has existed since 3.4, and non-isolated runs are completely unaffected.

The attach-to-PID injection path also inserts a directory, but it deletes it again right after importing its helper, so it's left as-is.

### Testing

Added `test_isolated_mode_sys_path` (parametrized over file / `-m` / `-c`) to `tests/debugpy/test_cli_args.py`, asserting the target directory is not prepended to `sys.path[0]` when the debuggee runs under `-I`. The three new cases fail without this change and pass with it. Verified locally on Windows / Python 3.12; `test_run.py` shows no new failures introduced (identical pass/fail set with and without the change).

### DIFF
--- a/src/debugpy/server/cli.py
+++ b/src/debugpy/server/cli.py
@@ -348,10 +348,13 @@ def run_file():
     # if the target is a file (rather than a directory), it does not add its
     # parent directory to sys.path. Thus, importing other modules from the
     # same directory is broken unless sys.path is patched here.
+    # In isolated mode, Python itself doesn't add the script directory, so
+    # don't do it here either.
 
     if target is not None and os.path.isfile(target):
-        dir = os.path.dirname(target)
-        sys.path.insert(0, dir)
+        if not sys.flags.isolated:
+            dir = os.path.dirname(target)
+            sys.path.insert(0, dir)
     else:
         log.debug("Not a file: {0!r}", target)
 
@@ -362,9 +365,11 @@ def run_file():
 
 
 def run_module():
-    # Add current directory to path, like Python itself does for -m. This must
-    # be in place before trying to use find_spec below to resolve submodules.
-    sys.path.insert(0, str(""))
+    # Add current directory to path, like Python itself does for -m, unless
+    # it's suppressed by isolated mode. This must be in place before trying
+    # to use find_spec below to resolve submodules.
+    if not sys.flags.isolated:
+        sys.path.insert(0, str(""))
 
     # We want to do the same thing that run_module() would do here, without
     # actually invoking it.
@@ -396,8 +401,10 @@ def run_module():
 
 def run_code():
     if options.target is not None:
-        # Add current directory to path, like Python itself does for -c.
-        sys.path.insert(0, str(""))
+        # Add current directory to path, like Python itself does for -c,
+        # unless it's suppressed by isolated mode.
+        if not sys.flags.isolated:
+            sys.path.insert(0, str(""))
         code = compile(options.target, str("<string>"), str("exec"))
 
         start_debugging(str("-c"))
--- a/src/debugpy/server/cli.py
+++ b/src/debugpy/server/cli.py
@@ -340,6 +340,14 @@ def start_debugging(argv_0):
     os.environ["DEBUGPY_RUNNING"] = "true"
 
 
+def _skip_sys_path_prepend():
+    # Python does not prepend the script's parent directory (for a file target)
+    # or the current directory (for -m / -c) to sys.path in isolated mode (-I) or
+    # safe-path mode (-P / PYTHONSAFEPATH), so debugpy shouldn't either.
+    # `sys.flags.safe_path` was added in Python 3.11; fall back to False on older versions.
+    return sys.flags.isolated or getattr(sys.flags, "safe_path", False)
+
+
 def run_file():
     target = options.target
     start_debugging(target)
@@ -348,11 +356,11 @@ def run_file():
     # if the target is a file (rather than a directory), it does not add its
     # parent directory to sys.path. Thus, importing other modules from the
     # same directory is broken unless sys.path is patched here.
-    # In isolated mode, Python itself doesn't add the script directory, so
-    # don't do it here either.
+    # In isolated / safe-path mode, Python itself doesn't add the script directory,
+    # so don't do it here either.
 
     if target is not None and os.path.isfile(target):
-        if not sys.flags.isolated:
+        if not _skip_sys_path_prepend():
             dir = os.path.dirname(target)
             sys.path.insert(0, dir)
     else:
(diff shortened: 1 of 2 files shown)


karandhaodiyal28-hash:

rchiodo:

src/debugpy/server/cli.py
```diff
@@ -362,9 +365,11 @@ def run_file():
 
 
 def run_module():
-    # Add current directory to path, like Python itself does for -m. This must
-    # be in place before trying to use find_spec below to resolve submodules.
-    sys.path.insert(0, str(""))
+    # Add current directory to path, like Python itself does for -m, unless
+    # it's suppressed by isolated mode. This must be in place before trying
```
rchiodo: > **Warning** · Non-blocking recommendation

CPython suppresses the `sys.path[0]` prepend whenever `safe_path` is set, not only under `isolated`. `-P` / `PYTHONSAFEPATH=1` set `safe_path` without setting `isolated`, so `python -P -m debugpy ...` still injects the current directory. Please broaden the gate to account for `sys.flags.safe_path` (with a compatibility fallback for Python <3.11), or explicitly document why this fix intentionally covers only `-I`.

[verified]

rchiodo (approved):
Approved via Review Center.

karandhaodiyal28-hash:
Thanks @rchiodo, that's a great catch — you're right that `safe_path` (via `-P` / `PYTHONSAFEPATH`) suppresses the prepend independently of `isolated`.

I've broadened the gate to cover both. All three sites (`run_file`, `run_module`, `run_code`) now go through a small helper:

```python
def _skip_sys_path_prepend():
    # `sys.flags.safe_path` was added in Python 3.11; fall back to False on older versions.
    return sys.flags.isolated or getattr(sys.flags, "safe_path", False)
```

The `getattr(..., False)` keeps it safe on the pre-3.11 versions debugpy still supports. I also parametrized the regression test over both `-I` and `-P` (the `-P` cases are skipped below 3.11). Verified locally: with only the previous `isolated` gate the three `-P` cases fail, and they pass with this change; the full `test_cli_args.py` is green.

tests/debugpy/test_cli_args.py
```diff
@@ -33,3 +39,59 @@ def code_to_debug():
...
+    )
+
+    output = subprocess.check_output(
+        [sys.executable, python_flag, "-c", wrapper],
+        cwd=tmpdir.strpath,
+        stderr=subprocess.DEVNULL,
+    )
+
+    sys_path_0 = output.decode("utf-8").strip().splitlines()[-1]
+    assert sys_path_0 != repr(not_expected)
```
rchiodo: > **Warning** · Non-blocking recommendation

Harden this subprocess-based regression test: add a timeout and preserve child stderr so a hung or failing debuggee produces a timely, diagnosable failure. The positional last-line parsing and negative-only assertion could also mask unrelated output or an incorrect replacement path; use an explicit marker and assert the expected path behavior. [verified]

rchiodo (approved):
Approved via Review Center.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [tests/debugpy/test_cli_args.py:L169](https://github.com/microsoft/debugpy/blob/f733580d8eee1d17698b22bff8327a87d09fbf07/tests/debugpy/test_cli_args.py#L169).

> **Warning** · Non-blocking recommendation

`assert completed.stdout` checks raw bytes, but parsing strips before indexing. Whitespace-only stdout passes the assertion and then raises `IndexError`, losing the captured stderr diagnostics. Guard the parsed lines instead before reading `lines[-1]`.

[verified]

rchiodo:
The new integration test documents occasional Windows hangs while starting a real debug server. Please consider making it deterministic by testing the sys.path branch directly or avoiding `--listen`.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95:

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

karandhaodiyal28-hash:
Hey! Just a gentle reminder — this PR has all approvals and CI is green (mergeable_state: CLEAN). Would appreciate a review when you get a chance. Thanks! 🙏

rchiodo:
Yep sorry, forgot to merge. Looks good. Thanks for the submission.

PR #2051, opened 2026-07-31
Return the endpoint from listen() with the in-process adapter

### LINKED ISSUE #1656: debugpy.listen() doesn't return endpoint if "in_process_debug_adapter=True" is passed

## Environment data

- debugpy version: `1.8.5`
- OS and version: `Linux 6.8.0-40-generic #40~22.04.3-Ubuntu`
- Python version (& distribution if applicable, e.g. Anaconda): `3.8.19`
- Using VS Code or Visual Studio: `VS Code`

## Actual behavior

Calling `debugpy.listen(0, in_process_debug_adapter=True)` returns `None`.

## Expected  behavior

The [API reference](https://github.com/microsoft/debugpy/wiki/API-Reference) says `debugpy.listen` should return a `(host, port)` tuple, and indeed it does do that when calling `debugpy.listen(0)` or `debugpy.listen(0, in_process_debug_adapter=False)`.

Makes it a little inconvenient when passing 0 as the port since I have no way to know what port is actually being listened on.

## Steps to reproduce:

```
import debugpy
try:
  host, port = debugpy.listen(5678, in_process_debug_adapter=True)
  print(host, port)
except Exception as e:
  print(f"Exception: {type(e)}: {e}")
```
Outputs: "Exception: <class 'TypeError'>: cannot unpack non-iterable NoneType object"

karandhaodiyal28-hash:
`debugpy.listen()` is documented to return the `(host, port)` it ends up listening on, and it does — except when you pass `in_process_debug_adapter=True`, where it returns `None`:

```python
host, port = debugpy.listen(5678, in_process_debug_adapter=True)
# TypeError: cannot unpack non-iterable NoneType object
```

The out-of-process path returns `client_host, client_port` at the end, but the in-process branch just did a bare `return`. This is especially annoying when you pass port `0`, because there's then no way to find out which port was actually chosen.

This returns the endpoint from the in-process branch as well, and sets `listen.called` so both branches behave the same way with respect to the "listen already called" guard.

I added a small regression test that stubs out `_settrace` and checks the return value.

Fixes #1656.

### DIFF
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -156,8 +156,18 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
             block_until_connected=False,
             **settrace_kwargs
         )
+        # pydevd binds the listening socket on a background reader thread, so
+        # the OS-assigned port (when port=0 is passed) isn't visible from
+        # here directly. Wait for the socket to be ready and read the actual
+        # bound endpoint back from the global debugger.
+        pydb = get_global_debugger()
+        if pydb is not None:
+            pydb.wait_for_server_socket_ready()
+            actual_host, actual_port = pydb._server_socket_name
+        else:
+            actual_host, actual_port = host, port
         listen.called = True
-        return host, port
+        return actual_host, actual_port
 
     import subprocess
 
--- a/src/debugpy/server/api.py
+++ b/src/debugpy/server/api.py
@@ -156,7 +156,8 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
             block_until_connected=False,
             **settrace_kwargs
         )
-        return
+        listen.called = True
+        return host, port
 
     import subprocess
 
--- a/tests/debugpy/test_api.py
+++ b/tests/debugpy/test_api.py
@@ -1,0 +1,18 @@
+import pytest
+
+import debugpy.server.api as _api
+
+
+@pytest.fixture
+def no_settrace(monkeypatch):
+    # Avoid actually starting pydevd; we only care about listen()'s return value.
+    monkeypatch.setattr(_api, "_settrace", lambda **kwargs: None)
+    monkeypatch.setattr(_api.listen, "called", False)
+
+
+def test_listen_in_process_returns_endpoint(no_settrace):
+    # Regression test for #1656: listen(..., in_process_debug_adapter=True) used
+    # to return None, unlike the out-of-process path which returns (host, port).
+    endpoint = _api.listen(("127.0.0.1", 5678), in_process_debug_adapter=True)
+    assert endpoint == ("127.0.0.1", 5678)
+    assert _api.listen.called is True
--- a/tests/debugpy/test_api.py
+++ b/tests/debugpy/test_api.py
@@ -16,3 +16,42 @@ def test_listen_in_process_returns_endpoint(no_settrace):
     endpoint = _api.listen(("127.0.0.1", 5678), in_process_debug_adapter=True)
     assert endpoint == ("127.0.0.1", 5678)
     assert _api.listen.called is True
+
+
+class _FakePyDB:
+    """Minimal stand-in for the global pydevd debugger.
+
+    Mirrors the bits of the real PyDB that listen()'s in-process path reads
+    after _settrace(): ``wait_for_server_socket_ready`` and
+    ``_server_socket_name`` (set by the reader thread once the listening
+    socket is bound).
+    """
+
+    def __init__(self, host, port):
+        self._server_socket_name = (host, port)
+
+    def wait_for_server_socket_ready(self):
+        return None
+
+
+@pytest.fixture
+def in_process_debugger(monkeypatch):
+    # Stub out _settrace (don't actually start pydevd) and inject a fake
+    # global debugger so that listen()'s in-process path can read the
+    # actual bound endpoint back from it.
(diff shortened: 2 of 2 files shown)


rchiodo:

src/debugpy/server/api.py
```diff
@@ -156,7 +156,8 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
             block_until_connected=False,
             **settrace_kwargs
         )
-        return
+        listen.called = True
+        return host, port
```
rchiodo: > **Issue** · Please address or respond

`port` is still the requested input value here, so `listen(0, in_process_debug_adapter=True)` now returns `(host, 0)` rather than the OS-assigned port. This does not meet the documented endpoint contract or the PR's stated port-0 motivation. Please obtain and return pydevd's bound port, or narrow the change and its description to explicitly document this limitation.

[verified]

tests/debugpy/test_api.py
```diff
@@ -0,0 +1,18 @@
...
+    # Avoid actually starting pydevd; we only care about listen()'s return value.
+    monkeypatch.setattr(_api, "_settrace", lambda **kwargs: None)
+    monkeypatch.setattr(_api.listen, "called", False)
+
+
+def test_listen_in_process_returns_endpoint(no_settrace):
+    # Regression test for #1656: listen(..., in_process_debug_adapter=True) used
+    # to return None, unlike the out-of-process path which returns (host, port).
+    endpoint = _api.listen(("127.0.0.1", 5678), in_process_debug_adapter=True)
+    assert endpoint == ("127.0.0.1", 5678)
```
rchiodo, line 17: > **Warning** · Non-blocking recommendation

The fixed-port, no-op `_settrace` test only verifies that the input endpoint is echoed back; it cannot catch the `port=0` contract gap. Add coverage that establishes whether the in-process path returns the resolved bound port for port 0.

[verified]

karandhaodiyal28-hash:
Updated! Retrieved the actual bound port from pydevd for the in-process adapter path, and added regression test coverage for port 0.

src/debugpy/server/api.py
```diff
@@ -156,7 +156,18 @@ def listen(address, settrace_kwargs, in_process_debug_adapter=False):
...
+        # pydevd binds the listening socket on a background reader thread, so
+        # the OS-assigned port (when port=0 is passed) isn't visible from
+        # here directly. Wait for the socket to be ready and read the actual
+        # bound endpoint back from the global debugger.
+        pydb = get_global_debugger()
+        if pydb is not None:
+            pydb.wait_for_server_socket_ready()
+            actual_host, actual_port = pydb._server_socket_name
+        else:
+            actual_host, actual_port = host, port
```
rchiodo, line 168: > **Warning** · Non-blocking recommendation

If `_settrace()` succeeds without installing a global debugger, this fallback returns the requested endpoint and can silently return port `0`, violating the documented actual-endpoint contract. Treat the missing debugger as an invariant failure rather than returning a success-shaped value.

[verified]

tests/debugpy/test_api.py
```diff
@@ -0,0 +1,57 @@
+import pytest
+
+import debugpy.server.api as _api
+
+
+@pytest.fixture
+def no_settrace(monkeypatch):
+    # Avoid actually starting pydevd; we only care about listen()'s return value.
+    monkeypatch.setattr(_api, "_settrace", lambda **kwargs: None)
+    monkeypatch.setattr(_api.listen, "called", False)
```
rchiodo, line 10: > **Info** · Optional note

The fixture stubs `_settrace` but not `ensure_logging()`, which the `@_starts_debugging` wrapper still calls and can leave process-global logging state behind for other tests. Stub or restore that state so this unit test remains isolated.

[verified]

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

karandhaodiyal28-hash:
Thanks for the CI run, @rchiodo!

The Windows py310 and py312 failures are both the same single test: \	ests/debugpy/test_threads.py::test_thread_count[program-launch-1]\. The failure is \ssert 2 == 1\ (got MainThread + Dummy-6, expected count=1) — a timing/visibility race in the test itself, not a regression from this PR.

Key points:
- This PR only changes the **in-process adapter** path in \pi.py\ (after the in-process \_settrace\ setup, the listen() in-process branch now reads the actual bound port via \py_db._server_socket_name\). \	est_thread_count[program-launch-1]\ uses the **out-of-process** adapter (\Program\ target → spawns an external Python script), so this code path is not exercised at all.
- Windows py311, py313, py314 all pass the same test on the same Windows runner image. The version-specific failure pattern is consistent with a thread-visibility race, not a regression.
- The test asserts \len(threads['threads']) == count\, but the debuggee always has \MainThread + N user threads\, so the test is also structurally racy (it should be \== count + 1\ to include the main thread, or filter it out).

Could you re-trigger with \/azp run\? If the same test fails again on the same Python versions, that'd be a stronger signal and worth investigating, but I'd expect it to pass on the next attempt.

StellaHuang95:

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:
LGTM, thanks for the PR.

PR #2052, opened 2026-07-31
Add lldb attach-to-PID option for debugpy on Linux

pdepetro:
Add an option to prefer lldb (falling back to gdb) when attaching to a process by PID on Linux.

debugpy's attach-by-PID flow injects the debug server into a target process. On Linux it has always shelled out to gdb; on macOS it uses lldb. In some environments lldb is the available/preferred injector on Linux. This adds an opt-in preference, surfaced as a VS Code launch configuration option "linuxAttachPreferLldb".

The selector flows end to end: adapter/clients.py reads linuxAttachPreferLldb from the attach request and forwards --linux-attach-prefer-lldb to the injector subprocess; server/cli.py parses that switch and sets the PYDEVD_ATTACH_PREFER_LLDB environment variable; add_code_to_python_process.py reads it at call time in the new run_python_code_linux dispatcher, which uses lldb when it is preferred and present on PATH, and otherwise transparently falls back to gdb.

Implementation notes:
- run_python_code_linux_lldb mirrors the existing run_python_code_mac lldb invocation but resolves the Linux .so via get_target_filename() and reuses the shared linux_and_mac/lldb_prepare.py driver. No native code changes are needed: both paths call the same exported DoAttach entry point.
- The PEP 768 sys.remote_exec fast path (Python 3.14+) runs first and is unaffected; the lldb preference only applies on the gdb/lldb fallback path.

### DIFF
--- a/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py
+++ b/src/debugpy/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py
@@ -71,6 +71,7 @@ See: attach_pydevd.py to attach the pydev debugger to a running python process.
 # nasm.asm&x:\nasm\nasm-2.07-win32\nasm-2.07\ndisasm.exe -b arch nasm
 import ctypes
 import os
+import shutil
 import struct
 import subprocess
 import sys
@@ -395,7 +396,7 @@ def _win_write_to_shared_named_memory(python_code, pid):
         CloseHandle(filemap)
 
 
-def run_python_code_linux(pid, python_code, connect_debugger_tracing=False, show_debug_info=0):
+def run_python_code_linux_gdb(pid, python_code, connect_debugger_tracing=False, show_debug_info=0):
     assert "'" not in python_code, "Having a single quote messes with our command."
 
     target_dll = get_target_filename()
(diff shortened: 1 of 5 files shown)


rchiodo:

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [tests/debugpy/server/test_cli.py:L390](https://github.com/microsoft/debugpy/blob/d00cc12f45690217e81958496acbd34613910018/tests/debugpy/server/test_cli.py#L390).

> **Warning** · Non-blocking recommendation

This assertion hard-codes `show_debug_info=0`, but production reads `DEBUGPY_ATTACH_BY_PID_DEBUG_INFO`. The test isolates `PYDEVD_ATTACH_PREFER_LLDB` but not that variable, so environments that set it will fail this test spuriously. Clear it in the patched environment or assert `show_debug_info=some.int`.

src/debugpy/_vendored/pydevd/pydevd_attach_to_process/add_code_to_python_process.py
```diff
@@ -544,6 +545,64 @@ def run_python_code_mac(pid, python_code, connect_debugger_tracing=False, show_d
...
+        ]
+    )
+
+    env = os.environ.copy()
+    # Remove the PYTHONPATH (if lldb has a builtin Python it could fail if we
+    # have the PYTHONPATH for a different python version or some forced encoding).
+    env.pop("PYTHONIOENCODING", None)
+    env.pop("PYTHONPATH", None)
+    print("Running: %s" % (" ".join(cmd)))
+    subprocess.check_call(" ".join(cmd), shell=True, env=env)
```
rchiodo: > **Warning** · Non-blocking recommendation

`run_python_code_linux_lldb` duplicates the macOS lldb injector, including command construction, quoting, environment scrubbing, and subprocess execution. This creates two sources of truth for lldb fixes. Please extract the shared lldb flow into a helper, or explicitly document why the implementations must remain mirrored.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:
LGTM, thanks for the PR. I'll submit it once the tests pass.

PR #2055, opened 2026-08-05
Fix debugger hang when expanding objects with blocking property getters (#2053)

### LINKED ISSUE #2053: Debugger hangs when inspecting LanceDBConnection object on Python 3.13
Description

The VS Code Python debugger hangs when a breakpoint is hit after creating a LanceDB connection. The issue only occurs with Python 3.13.5. The same code works correctly with Python 3.11.

The program runs normally outside the debugger.

Environment
OS: Debian 13
Python: 3.13.5
VS Code Python Debugger extension: 2026.6.0
LanceDB: 0.26.1
Debugger: debugpy (latest version installed)
Steps to reproduce

Minimal example:

import lancedb

print("before connect")
db = lancedb.connect("testdb")
print("after connect")

Steps:

Open the file in VS Code.
Start debugging with F5.
Add a breakpoint on:
print("after connect")
Run the debugger.
Expected behavior

The debugger should stop at the breakpoint and allow normal debugging.

Actual behavior

The debugger stops at the breakpoint but hangs. VS Code becomes unresponsive while the breakpoint is active.

The program continues to run normally if no breakpoint is set.

Additional observations

The issue appears related to debugger inspection of the LanceDBConnection object.

The following works:

print(repr(db))

Output:

LanceDBConnection(uri='/home/msadmin/Projects/AI/prod/lancedb/testdb')

Also:

print(vars(db))

Output:

{
 'storage_options': None,
 '_conn': ListingDatabase(uri=/home/msadmin/Projects/AI/prod/lancedb/testdb, read_consistency_interval=None)
}

The debugger does not hang if the object is removed before the breakpoint:

db = lancedb.connect("testdb")
db = None

print("after connect")
Comparison

The same code and LanceDB version work correctly on another machine with:

Python 3.11

The issue only reproduces on:

Python 3.13.5
Suspected cause

It appears that debugpy hangs while inspecting the native-backed LanceDB object (ListingDatabase) when Python 3.13 is used.

Workaround

Removing the LanceDB connection object before hitting breakpoints avoids the problem:

db = None

or using Python 3.11 avoids the issue completely.

rchiodo:
## Summary

Fixes #2053.

Inspecting a `LanceDBConnection` in the debugger hung on Python 3.13. The root cause is **not** a slow `repr` — it's a **thread deadlock during variable expansion**.

At a breakpoint, debugpy suspends *all* threads (in `multi_threads_single_notification` mode), including background daemon threads. `LanceDBConnection` exposes `@property` getters such as `read_consistency_interval` that dispatch work to a background asyncio event loop running in a daemon thread and block on the result (`asyncio.run_coroutine_threadsafe(...).result()`). When the variables view expands the object, debugpy evaluates those properties on the suspended thread, which waits forever on the *also-suspended* background loop thread → deadlock.

This is why:
- `repr(db)` is fast — it never touches the properties.
- `vars(db)` works — it only reads the instance `__dict__`, no property evaluation.
- Setting `db = None` before the breakpoint avoids the hang.

## Changes

**1. Recover instead of hang (the real cure)**
- New constant `PYDEVD_UNBLOCK_THREADS_ON_VARIABLES_TIMEOUT` (default `3.0`s). When resolving a variable's children takes longer than the timeout, the other threads are resumed until the resolution finishes, then re-suspended — so a property that depends on another thread can complete instead of deadlocking.
- Reuses the existing evaluate-path anti-deadlock machinery, extracted into a shared `unblock_threads_on_timeout` context manager in `pydevd_vars.py`. `_run_with_unblock_threads` now uses it (evaluate-path behavior unchanged).
- Wired into `internal_get_variable_json` (the DAP variables request).

**2. Dedupe double `getattr`**
- `DefaultResolver._get_py_dictionary` previously did `hasattr()` + `getattr()`, evaluating every property/descriptor **twice**. It now calls `getattr()` a single time inside `try/except AttributeError`. This halves the cost (and side effects) of expanding objects with expensive properties.

## Tests

- New resource `_debugger_case_deadlock_thread_variables.py` and `test_debugger_case_deadlock_thread_variables` reproduce the exact scenario (a property getter blocking on a suspended background thread). The test **passes** with the fix and **deadlocks/fails** when the timeout is disabled, proving it catches the regression.
- All existing evaluate-path deadlock/unblock tests pass (`test_debugger_case_deadlock_thread_eval`, `test_debugger_case_breakpoint_on_unblock_thread_eval`, `test_debugger_case_unblock_manually`, `test_debugger_case_deadlock_notify_evaluate_timeout`, `test_debugger_case_deadlock_interrupt_thread`).
- `test_hasattr_failure`, `test_getattr_warning`, and the broader variable/resolver test set pass, confirming the `getattr` dedupe is behavior-preserving.

All changes are in pure-Python pydevd modules (no Cython regeneration needed).

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
@@ -8,6 +8,7 @@ from _pydevd_bundle.pydevd_constants import get_frame, get_current_thread_id, it
 from _pydevd_bundle.pydevd_xml import ExceptionOnEvaluate, get_type, var_to_xml
 from _pydev_bundle import pydev_log
 import functools
+from contextlib import contextmanager
 from _pydevd_bundle.pydevd_thread_lifecycle import resume_threads, mark_thread_suspended, suspend_all_threads, suspend_threads_lock
 from _pydevd_bundle.pydevd_comm_constants import CMD_SET_BREAK
 
@@ -309,22 +310,36 @@ def _run_with_interrupt_thread(original_func, py_db, curr_thread, frame, express
             return original_func(py_db, frame, expression, is_exec)
 
 
-def _run_with_unblock_threads(original_func, py_db, curr_thread, frame, expression, is_exec):
+@contextmanager
+def unblock_threads_on_timeout(py_db, curr_thread, timeout, on_timeout_message=None):
+    """
+    Context manager that runs the wrapped block and, if it doesn't finish until ``timeout`` seconds
+    elapse, resumes all other threads -- so that a computation depending on another (currently
+    suspended) thread can make progress -- and then suspends them again when the block finishes.
+
+    This only has an effect when threads are suspended/resumed as a group
+    (i.e.: ``py_db.multi_threads_single_notification``) and ``timeout >= 0``; otherwise the block
+    simply runs unchanged.
+
+    :param on_timeout_message:
+        Optional warning message sent to the client if the timeout elapses and the other threads are
+        resumed (so the user understands why other threads ran while execution was suspended).
+    """
     on_timeout_unblock_threads = None
     timeout_tracker = py_db.timeout_tracker  # : :type timeout_tracker: TimeoutTracker
 
-    if py_db.multi_threads_single_notification:
-        unblock_threads_timeout = pydevd_constants.PYDEVD_UNBLOCK_THREADS_TIMEOUT
-    else:
-        unblock_threads_timeout = -1  # Don't use this if threads are managed individually.
-
-    if unblock_threads_timeout >= 0:
-        pydev_log.info("Doing evaluate with unblock threads timeout: %s.", unblock_threads_timeout)
+    if py_db.multi_threads_single_notification and timeout >= 0:
+        pydev_log.info("Running with unblock threads timeout: %s.", timeout)
         tid = get_current_thread_id(curr_thread)
 
         def on_timeout_unblock_threads():
             on_timeout_unblock_threads.called = True
-            pydev_log.info("Resuming threads after evaluate timeout.")
+            pydev_log.info("Resuming threads after timeout.")
+            if on_timeout_message is not None:
+                try:
+                    py_db.writer.add_command(py_db.cmd_factory.make_warning_message(on_timeout_message))
+                except Exception:
+                    pydev_log.exception()
             resume_threads("*", except_thread=curr_thread)
             py_db.threads_suspended_single_notification.on_thread_resume(tid, curr_thread)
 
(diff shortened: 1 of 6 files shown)


StellaHuang95:

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
```diff
@@ -346,6 +360,16 @@ def on_timeout_unblock_threads():
                 py_db.threads_suspended_single_notification.on_thread_suspend(tid, curr_thread, CMD_SET_BREAK)
```
StellaHuang95: > **Issue** · Please address or respond

`called` is set before `resume_threads(...)` without sharing `suspend_threads_lock` with the `finally` block. The `finally` can observe `called`, re-suspend, and then the callback can resume threads afterward, leaving debugger state inconsistent. Perform the callback's state transition and resume under `suspend_threads_lock` so it is serialized with re-suspension.

src/debugpy/_vendored/pydevd/tests_python/resources/_debugger_case_deadlock_thread_variables.py
```diff
@@ -0,0 +1,81 @@
...
+property whose getter dispatches work to that secondary thread and blocks waiting for the result.
+
+So, when all threads are stopped at a breakpoint, *expanding* that object in the variables view (which
+evaluates its properties) would be locked until the secondary thread is allowed to run.
+
+This mirrors real-world objects such as lancedb's ``LanceDBConnection``, whose property getters block
+on a background asyncio event loop running in a daemon thread.
+See: https://github.com/microsoft/debugpy/issues/2053
+"""
+
```
StellaHuang95, line 11: > **Issue** · Please address or respond

Please remove the inline GitHub issue URL. The repository rule requires issue references to remain in PR or commit metadata rather than committed source or test comments.

src/debugpy/_vendored/pydevd/tests_python/test_debugger_json.py
```diff
@@ -6178,6 +6178,36 @@ def get_environ(self):
         writer.finished_ok = True
 
 
+def test_debugger_case_deadlock_thread_variables(case_setup_dap):
+    # Expanding a variable whose property getter blocks on another (suspended) thread must not
+    # deadlock the debugger: the other threads should be resumed until the resolution finishes.
```
StellaHuang95, line 6183: > **Issue** · Please address or respond

Please remove the inline GitHub issue URL. The repository rule requires issue references to remain in PR or commit metadata rather than committed source or test comments.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py
rchiodo: Fixed — the resume (in the timeout callback) and the re-suspend (in the finally) now both run under suspend_threads_lock, sharing a small state dict (resumed/finished). The callback bails out if the block already finished, and the finally only re-suspends if the callback actually resumed, so the two transitions are serialized in both directions (no re-suspend-then-resume window, and no resume-after-finished).

StellaHuang95 (approved):
Approved via Review Center.

PR #2056, opened 2026-08-05
Fix duplicate stopped event when two threads hit a breakpoint at once

aperez:
Suspending other threads marks them `CMD_THREAD_SUSPEND`, which is what lets the pydevd single notification strategy drop their redundant stopped events. But a line event checks breakpoints before it checks for a pending suspend, so a thread suspended while blocked still reports its own breakpoint on waking, overwriting that reason and emitting a second stopped event.

Leave an already suspended thread's reason alone. It then keeps the reason it first stopped for, which is invisible over DAP and shows only on the legacy protocol with a suspend-all breakpoint policy.

### DIFF
--- a/tests/debugpy/test_threads.py
+++ b/tests/debugpy/test_threads.py
@@ -6,7 +6,8 @@ import pytest
 import sys
 import time
 
-from tests import debug
+from tests import debug, timeline
+from tests.patterns import some
 
 
 @pytest.mark.parametrize("count", [1, 3])
@@ -187,3 +188,59 @@ def test_debug_this_thread(pyfile, target, run):
 
         session.wait_for_stop()
         session.request_continue()
+
+
+def test_concurrent_breakpoint_hits(pyfile, target, run):
+    """
+    A thread that reaches its breakpoint while already suspended by another thread's
+    breakpoint must not produce a second stop.
+
+    A breakpoint hit suspends every other thread, and a line event evaluates breakpoints
+    before it checks for a pending suspend, so the second thread still reports its own
+    breakpoint. Re-marking it as suspended overwrites the CMD_THREAD_SUSPEND reason that
+    would have made the single notification behavior drop the duplicate.
+    """
+
+    @pyfile
+    def code_to_debug():
+        import debuggee
+        import threading
+        import time
+
+        debuggee.setup()
+
+        # Never released: `second` blocks in the C level acquire, running no traced
+        # line, so it can be marked suspended while its next traced line is still its
+        # own breakpoint. Main must reach @bp1 before that acquire times out.
+        gate = threading.Lock()
+        gate.acquire()
+
+        def second():
+            gate.acquire(timeout=2)
+            print("second")  # @bp2
+
+        thread = threading.Thread(target=second, name="second")
+        thread.start()
+        time.sleep(0.5)
+
+        print("main")  # @bp1
+        thread.join()
+
+    with debug.Session() as session:
+        session.expected_exit_code = some.int
+        with run(session, target(code_to_debug)):
+            session.set_breakpoints(code_to_debug, ["bp1", "bp2"])
+
+        session.wait_for_stop()
+        # Outlast the acquire timeout, so a duplicate notification has time to arrive.
+        time.sleep(3)
+
+        # Otherwise the test passes vacuously whenever `second` never reaches @bp2.
+        threads = {t["name"]: t["id"] for t in session.request("threads")["threads"]}
+        stack_trace = session.request("stackTrace", {"threadId": threads["second"]})
+        assert stack_trace["stackFrames"][0]["line"] == code_to_debug.lines["bp2"]
+
+        session.disconnect()
+
+    stops = session.timeline.all_occurrences_of(timeline.Event("stopped"))
+    assert len(stops) == 1
--- a/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
@@ -1537,7 +1537,10 @@ def _internal_line_event(func_code_info, frame, line):
     # print('line event', info.pydev_state, line, threading.current_thread(), code)
     # If we reached here, it was not filtered out.
 
-    if func_code_info.breakpoint_found:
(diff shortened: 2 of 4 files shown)


rchiodo:

tests/debugpy/test_threads.py
```diff
@@ -187,3 +187,59 @@ def foo(x):
...
+            print("second")  # @bp2
+
+        thread = threading.Thread(target=second, name="second")
+        thread.start()
+        time.sleep(0.5)
+
+        print("main")  # @bp1
+        thread.join()
+
+    with debug.Session() as session:
```
rchiodo, line 229: > **Issue** · Please address or respond

`None` means the debuggee is not expected to exit, but this launch-mode debuggee exits with code 0 after `disconnect()`, causing teardown to fail with `assert 0 == None`. Use the harness's any-integer expectation (`some.int`, importing it from `tests.patterns`) instead.

aperez: I am using `expected_exit_code = some.int` now.

rchiodo (approved):
Approved via Review Center.

rchiodo:
It looks like this is breaking django/flask breakpoints:

These two tests are continually failing for all os/versions:

test_django_template_exception_no_multiproc[launch]
test_flask_template_exception_no_multiproc[launch]

aperez:
> It looks like this is breaking django/flask breakpoints:
> 
> These two tests are continually failing for all os/versions:
> 
> test_django_template_exception_no_multiproc[launch] test_flask_template_exception_no_multiproc[launch]

Ah , I wasn't running the django and flask tests while working on this fix. Let me edit the fix so it does not regress those tests.

aperez:
@rchiodo latest version filters by stop reason to make sure we're only including suspends.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_thread_lifecycle.py
```diff
@@ -28,6 +28,13 @@ def pydevd_find_thread_by_id(thread_id):
 def mark_thread_suspended(thread, stop_reason: int, original_step_cmd: int = -1, main_suspend: bool = True):
     pydev_log.info("Marking thread suspended. Name: %s, stop_reason: %s, main_suspend: %s", thread.name, stop_reason, main_suspend)
     info = set_additional_thread_info(thread)
+    # A thread suspended only as a side effect of another thread's stop has not reported
+    # a reason of its own yet, and the single notification behavior keys off that reason
+    # to drop its redundant stopped event. Any other reason is a real stop, so it must
+    # still be allowed to overwrite an earlier one.
+    if info.pydev_state != STATE_RUN and getattr(thread, "stop_reason", None) == CMD_THREAD_SUSPEND:
```
rchiodo: > **Warning** · Non-blocking recommendation

This intentionally fixes the duplicate at the shared write rather than reordering the line-event handler that evaluates breakpoints before pending suspends. Please add an explicit issue/reference for the deferred ordering fix so that root cause remains tracked.

src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_thread_lifecycle.py
```diff
@@ -28,6 +28,13 @@ def pydevd_find_thread_by_id(thread_id):
 def mark_thread_suspended(thread, stop_reason: int, original_step_cmd: int = -1, main_suspend: bool = True):
     pydev_log.info("Marking thread suspended. Name: %s, stop_reason: %s, main_suspend: %s", thread.name, stop_reason, main_suspend)
     info = set_additional_thread_info(thread)
+    # A thread suspended only as a side effect of another thread's stop has not reported
+    # a reason of its own yet, and the single notification behavior keys off that reason
+    # to drop its redundant stopped event. Any other reason is a real stop, so it must
+    # still be allowed to overwrite an earlier one.
```
rchiodo: > **Warning** · Non-blocking recommendation

This early return skips the remaining suspend-type, stepping, and stop-state updates for every caller that re-marks an already suspended thread with `CMD_THREAD_SUSPEND`, not only this concurrent-breakpoint path. Please audit those callers; if any legitimately update those fields, narrow the guard to preserve only the stop reason.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

aperez:
@rchiodo Force-pushed a reworked fix; the test is unchanged. It turns out this only ever reproduced on the sys.monitoring backend: [pydevd_frame.py:684](https://github.com/microsoft/debugpy/blob/main/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_frame.py#L684) already folds `pydev_state != STATE_SUSPEND` into its line-breakpoint condition, so the `sys.settrace` tracer never reports a breakpoint on a thread that another thread has already suspended, whereas `_internal_line_event` had no such check and evaluated breakpoints before checking for a pending suspend. 
My change now does the same thing on the `sys.monitoring` side, which leaves `mark_thread_suspended` untouched  and fixes the ordering outright.
The same test still passes and still fails without the change, Django and Flask are green when I tested them locally. I put the codegen for the .pyx and .c is a separate commit.

tests/debugpy/test_threads.py
```diff
@@ -187,3 +188,59 @@ def foo(x):
...
+
+        # Otherwise the test passes vacuously whenever `second` never reaches @bp2.
+        threads = {t["name"]: t["id"] for t in session.request("threads")["threads"]}
+        stack_trace = session.request("stackTrace", {"threadId": threads["second"]})
+        assert stack_trace["stackFrames"][0]["line"] == code_to_debug.lines["bp2"]
+
+        session.disconnect()
+
+    stops = session.timeline.all_occurrences_of(timeline.Event("stopped"))
+    assert len(stops) == 1
```
rchiodo, line 246: > **Info** · Optional note

The test depends on `main` reaching `@bp1` and suspending all threads before this 2-second acquire times out. On a cold or heavily loaded runner, `second` could wake and stop independently, making the single-stop assertion flaky. Consider synchronizing this transition rather than relying on fixed timing.

rchiodo (approved):
Approved via Review Center.

rchiodo:
Thanks @aperez for bearing with us and fixing stuff. I'll submit this once the tests finish.

StellaHuang95 (approved):
Approved via Review Center.

PR #2061, opened 2026-08-10
Fix thread identity when the first traced call is another thread's is_alive()

aperez:
To learn which `threading.Thread` a running thread is, the debugger walks up the stack to the frame that started it and reads the `Thread` out of that frame's locals. `is_alive` counts as such a frame, and its `self` is taken to be the running thread. But `is_alive` is public and normally asks about a different thread, so `other.is_alive()` yields `other`, not the caller. The result is cached for the thread's lifetime, so its breakpoints, suspends and steps are then all reported under another thread's id, and once two threads share an id pydevd logs:
>  "Something is wrong. Tracker being added twice to the same thread id."

This change rejects the candidate when its `ident` is not the running thread's, which falls back to `threading`'s own id-to-thread table. The check is on identity rather than on the frame's name because `is_alive` is not the only way a wrong `self` can be read: frames are matched by file basename, so a `_bootstrap_inner` in any user module named `threading.py` qualifies too.

The test mirrors what an IDE user hits: a poller thread watching a worker's `is_alive()`, with the breakpoint set while the program is already running. The poller waits on a plain lock, which runs no Python, so `worker.is_alive()` is reliably its first traced call once the breakpoint turns tracing on. Without the fix the stop is reported on `worker` rather than `poller`.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
+++ b/src/debugpy/_vendored/pydevd/_pydevd_sys_monitoring/_pydevd_sys_monitoring.py
@@ -352,7 +352,7 @@ def _create_thread_info(depth):
         if f_bootstrap_frame.f_code.co_name in ("__bootstrap_inner", "_bootstrap_inner", "is_alive"):
             # Note: be careful not to use threading.current_thread to avoid creating a dummy thread.
             t = f_bootstrap_frame.f_locals.get("self")
-            if not isinstance(t, threading.Thread):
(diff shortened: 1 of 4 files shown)


StellaHuang95:

rchiodo (approved):

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

PR #2062, opened 2026-08-10
Add debugpy release agent

rchiodo:
## Summary

- add a user-invocable Debugpy Release agent
- automate version selection, tagging, signed internal builds, PyPI verification, and GitHub release creation
- discover internal pipeline configuration at runtime so confidential infrastructure details are not committed
- enforce ordering and safety gates for partial or failed releases

### DIFF
--- a/.github/agents/debugpy-release.agent.md
+++ b/.github/agents/debugpy-release.agent.md
@@ -25,6 +25,10 @@ Interpret requests as follows:
 - "Resume" or "finalize" inventories the external state for an explicit version
   and continues from the first incomplete phase, but only after verifying every
   completed phase against the same tag, commit, build, and package version.
+- If a new-release request discovers a different version already published to
+  PyPI but missing from GitHub, report that incomplete release and ask before
+  changing the requested target. A resume/finalize request for that exact
+  version may continue without additional confirmation.
 - "Next patch", "next minor", and "next major" are calculated from the highest
   stable version published by `microsoft/debugpy`, not from local tags.
 - An explicit version may be written with or without the leading `v`. Normalize
@@ -100,8 +104,10 @@ described below.
    PyPI, ignoring prereleases. Reconcile them deterministically:
    - If the highest stable versions match, use that version as the baseline.
    - If PyPI contains a newer version than GitHub, treat it as an incomplete
-     release and resume at GitHub release creation. Do not calculate or publish
-     a newer version.
+     release. For a new-release request, stop and ask whether to finalize that
+     existing version instead. For a resume/finalize request naming that exact
+     version, continue at GitHub release creation. Do not calculate or publish a
+     newer version.
    - If GitHub contains a newer version than PyPI, stop because customer
      workflows may already be inconsistent. Report that PyPI publication for
      the existing GitHub release must be recovered before another release.
@@ -157,8 +163,10 @@ Stop here for prepare/dry-run requests.
    developer clone may use `origin` for a fork.
 3. Verify the remote tag resolves to the intended full commit SHA.
 
-If a local tag was created but the push failed, report that exact state. Never
-move or force-push an existing release tag.
+If a local tag was created but the push failed, a later resume may re-push it
+only after verifying that the remote tag is still absent and the local annotated
+tag resolves to the previously verified commit. Otherwise stop and report the
+exact recovery action. Never move or force-push an existing release tag.
 
 ## Phase 3: Queue the Internal Real-Signed Build
 
@@ -225,22 +233,25 @@ only on the Azure DevOps run result.
 
 ## Phase 6: Create the GitHub Release
 
-1. Generate release notes from the previous stable tag through the new tag.
-   Keep relevant issue and pull request links. Review the generated text for
-   unrelated changes or internal information.
+1. Select the release-notes base deterministically. For a backport or older
+   release line, use the greatest stable tag lower than the new version with the
+   same major and minor components. Otherwise use the greatest stable tag lower
+   than the new version. Stop if no valid lower tag exists. Generate release
+   notes from that tag through the new tag, keep relevant issue and pull request
+   links, and review the text for unrelated changes or internal information.
 2. Compare the new version to the highest stable GitHub release that existed
-   before this workflow. Add `--latest` only when the new version is greater.
-   Omit it for backports and older release lines.
+   before this workflow. Pass `--latest` when the new version is greater and
+   `--latest=false` for backports and older release lines.
 3. Create the release in `microsoft/debugpy`:
 
    ```text
-   gh release create <tag> --repo microsoft/debugpy --title "debugpy <tag>" --notes-file <notes-file> [--latest]
+   gh release create <tag> --repo microsoft/debugpy --title "debugpy <tag>" --notes-file <notes-file> <--latest|--latest=false>
    ```
 
 4. Do not pass artifact files to `gh release create`.
 5. Verify the release is public, points to the intended tag, and contains no
-   attached binaries. Verify it is marked latest only when `--latest` was
-   required.
+   attached binaries. Verify it is marked latest when `--latest` was used and
+   is not marked latest when `--latest=false` was used.
 6. Delete any temporary release-notes file after successful publication.
 
 ## Completion Report
--- a/.github/agents/debugpy-release.agent.md
+++ b/.github/agents/debugpy-release.agent.md
@@ -99,7 +99,19 @@ described below.
    `https://github.com/microsoft/debugpy.git`. Fetch it if necessary and require
    `HEAD` to equal that commit. Do not release a fork-only commit. For resume,
    resolve the commit from the existing remote tag instead and do not retag the
-   current `main`.
+   current `main`. If the remote tag is absent but an exact local tag exists,
+   treat it as a local-tag recovery candidate only when:
+   - It is an annotated tag and the requested tag name matches exactly.
+   - The remote tag, PyPI version, GitHub release, and internal release state are
+     all absent.
+   - Its target equals the current authoritative `microsoft/debugpy` `main`
+     commit and that commit passes the required-check gate below.
+   - Rebuilding metadata from that tag produces the requested normalized
+     package version.
+
+   When every condition holds, route to Phase 2 to push that exact tag. If the
+   target no longer equals current `main`, do not move the tag; stop and report
+   that the user must choose whether to delete the local-only tag and restart.
 6. Read the complete sets of stable public versions from GitHub releases and
    PyPI, ignoring prereleases. Reconcile them deterministically:
    - If the highest stable versions match, use that version as the baseline.
@@ -113,11 +125,19 @@ described below.
      the existing GitHub release must be recovered before another release.
    - If the sets disagree in any other way that affects the proposed version,
      stop and report both sets. Never choose one source arbitrarily.
+   Record the highest stable GitHub version observed here as the immutable
+   latest-selection baseline for Phase 6. Do not try to reconstruct this
+   historical snapshot later during resume.
 7. Inventory the proposed version's tag, internal build, internal release run,
    PyPI files, and GitHub release. For a new release, require all to be absent.
-   For resume/finalize, require each existing item to match the same commit,
-   exact signed build, and package version, then continue from the first absent
-   item. Never recreate, move, or overwrite an existing item.
+   Except for the PyPI override below, resume/finalize requires every existing
+   item to match the same commit, exact signed build, and package version, then
+   continues from the first absent item. Never recreate, move, or overwrite an
+   existing item. PyPI publication overrides the generic first-absent rule:
+   when the exact version and files already exist on PyPI but the GitHub release
+   is absent, require the matching remote tag and proceed only to Phase 6. Do
+   not re-queue build or release pipelines even if their historical records are
+   unavailable or expired.
 8. Discover the authoritative required checks for `main` using the GitHub
    rules-for-branch API:
 
@@ -239,9 +259,10 @@ only on the Azure DevOps run result.
    than the new version. Stop if no valid lower tag exists. Generate release
    notes from that tag through the new tag, keep relevant issue and pull request
    links, and review the text for unrelated changes or internal information.
-2. Compare the new version to the highest stable GitHub release that existed
-   before this workflow. Pass `--latest` when the new version is greater and
-   `--latest=false` for backports and older release lines.
+2. Compare the new version to the immutable latest-selection baseline recorded
+   during Phase 1. Pass `--latest` when the new version is greater and
+   `--latest=false` for backports and older release lines. Do not re-query or
+   re-derive the baseline during a later resume.
 3. Create the release in `microsoft/debugpy`:
 
    ```text
--- a/.github/agents/debugpy-release.agent.md
+++ b/.github/agents/debugpy-release.agent.md
@@ -109,9 +109,12 @@ described below.
    - Rebuilding metadata from that tag produces the requested normalized
      package version.
 
-   When every condition holds, route to Phase 2 to push that exact tag. If the
-   target no longer equals current `main`, do not move the tag; stop and report
-   that the user must choose whether to delete the local-only tag and restart.
+   When every condition holds, continue through the remainder of Phase 1,
+   including baseline recording, inventory, required-check verification,
+   metadata validation, and the preflight summary. Then route to Phase 2 to push
+   that exact tag. If the target no longer equals current `main`, do not move
+   the tag; stop and report that the user must choose whether to delete the
+   local-only tag and restart.
 6. Read the complete sets of stable public versions from GitHub releases and
    PyPI, ignoring prereleases. Reconcile them deterministically:
    - If the highest stable versions match, use that version as the baseline.
@@ -125,9 +128,10 @@ described below.
      the existing GitHub release must be recovered before another release.
    - If the sets disagree in any other way that affects the proposed version,
      stop and report both sets. Never choose one source arbitrarily.
-   Record the highest stable GitHub version observed here as the immutable
-   latest-selection baseline for Phase 6. Do not try to reconstruct this
-   historical snapshot later during resume.
+   Record the highest stable GitHub version observed here as the
+   latest-selection baseline for this invocation. Recompute it deterministically
+   whenever a later invocation resumes the release; do not claim that an
+   unpersisted value survives across sessions.
 7. Inventory the proposed version's tag, internal build, internal release run,
    PyPI files, and GitHub release. For a new release, require all to be absent.
    Except for the PyPI override below, resume/finalize requires every existing
@@ -259,10 +263,13 @@ only on the Azure DevOps run result.
    than the new version. Stop if no valid lower tag exists. Generate release
    notes from that tag through the new tag, keep relevant issue and pull request
    links, and review the text for unrelated changes or internal information.
-2. Compare the new version to the immutable latest-selection baseline recorded
-   during Phase 1. Pass `--latest` when the new version is greater and
-   `--latest=false` for backports and older release lines. Do not re-query or
-   re-derive the baseline during a later resume.
+2. Recheck that the stable GitHub release set has not changed since the Phase 1
+   snapshot. If it changed, repeat Phase 1 reconciliation and update the
+   invocation's latest-selection baseline before continuing. For backports and
+   older release lines, always pass `--latest=false` regardless of the
+   baseline. Otherwise pass `--latest` only when the new version is greater
+   than the current invocation's baseline; pass `--latest=false` when it is
+   not.
 3. Create the release in `microsoft/debugpy`:
 
    ```text
--- a/.github/agents/debugpy-release.agent.md
+++ b/.github/agents/debugpy-release.agent.md
@@ -1,0 +1,195 @@
+---
+description: "Release debugpy to PyPI and GitHub. Use when asked to prepare or publish the next patch, minor, major, or explicit debugpy version."
+name: "Debugpy Release"
+argument-hint: "Specify patch, minor, major, or an explicit version. Say 'prepare' for a dry run."
+tools: [vscode/askQuestions, execute/runInTerminal, execute/getTerminalOutput, execute/awaitTerminal, read/readFile, search/fileSearch, search/textSearch, web/fetch, todo]
+user-invocable: true
+---
+
+# Debugpy Release Agent
+
+Automate the documented internal debugpy release process. Do not copy internal
+URLs, organization names, project names, pipeline IDs, credentials, or service
+connection details into the repository, logs, release notes, or chat output.
+
+## Intent
+
+Interpret requests as follows:
+
+- "Prepare" or "dry run" performs every read-only preflight and reports the
+  exact commands and release version, but does not create or push a tag, queue
+  publishing pipelines, or create a GitHub release.
+- "Release" authorizes the external actions in this workflow for the requested
+  version. Do not ask for redundant confirmation unless the resolved version or
+  target commit is ambiguous.
+- "Next patch", "next minor", and "next major" are calculated from the highest
+  stable version published by `microsoft/debugpy`, not from local tags.
+- An explicit version may be written with or without the leading `v`. Normalize
+  Git tags to `v<version>`.
+
+For example, if the latest stable release is `v1.8.21`, the next minor release
+is `v1.9.0`.
+
+## Non-Negotiable Safety Rules
+
+1. Release only from the current `microsoft/debugpy` `main` commit.
+2. The worktree must be clean. Never stash, discard, or include local changes.
+3. Versioneer derives the package version from the Git tag. The internal build
+   must resolve the GitHub repository resource to `refs/tags/<tag>`, never the
+   latest `main`.
+4. The build must use real signing. Never publish artifacts from a test-signed
+   or unsigned build.
+5. Queue the release pipeline only after the exact tagged build succeeds.
+6. Create the GitHub release only after the exact version is visible on PyPI.
+   Publishing the GitHub release first can break customer workflows.
+7. Do not attach binaries to the GitHub release.
+8. Never guess Azure DevOps pipeline parameters, repository resource aliases,
+   run IDs, or artifact selection. Inspect them and verify the queued run.
+9. Never upload with `twine` or a personal PyPI token. PyPI publication must go
+   through the approved authenticated internal release pipeline.
+10. Stop on any failed or partially successful stage. Do not continue with a
+    newer build or a different artifact.
+
+## Phase 1: Preflight
+
+1. Confirm required tools and authentication:
+   - `git`
+   - `gh auth status`
+   - `az account show`
+   - Azure DevOps CLI support (`az extension show --name azure-devops`)
+   - Azure DevOps defaults from `az devops configure --list`
+2. If the Azure DevOps extension is missing, install it with:
+
+   ```text
+   az extension add --name azure-devops
+   ```
+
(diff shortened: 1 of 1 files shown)


StellaHuang95:

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+   ```text
+   gh release create <tag> --repo microsoft/debugpy --title "debugpy <tag>" --notes-file <notes-file> --latest
+   ```
+
+3. Do not pass artifact files to `gh release create`.
+4. Verify the release is public, marked latest, points to the intended tag, and
+   contains no attached binaries.
+5. Delete any temporary release-notes file after successful publication.
+
+## Completion Report
```
StellaHuang95: > **Issue** · Please address or respond

`--latest` is unconditional even though explicit versions may be backports. Releasing `v1.8.22` after `v1.9.0` would incorrectly make the older line GitHub's latest release. Only pass `--latest` when the resolved version exceeds the current highest stable version, and make the verification conditional too.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+    - GitHub repository
+    - Discovered internal build and release pipeline names
+
+Stop here for prepare/dry-run requests.
+
+## Phase 2: Create and Push the Tag
+
+1. Create an annotated tag on the verified `microsoft/debugpy` `main` commit:
+
+   ```text
```
StellaHuang95: > **Warning** · Non-blocking recommendation

A failure after PyPI publication but before `gh release create` cannot be recovered with this agent: preflight rejects the existing tag/PyPI version, so it cannot finalize the incomplete release. Add a resume/finalize mode that detects prior external state and continues from the first incomplete phase.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+7. Fail if the proposed tag or PyPI version already exists.
+8. Verify required checks for the target commit have completed successfully.
+9. Run the repository's existing targeted packaging/version checks if
+   available. At minimum, use a temporary clone or worktree with a local-only
+   proposed tag to build package metadata without publishing, verify the
+   resulting version is exactly the proposed version, and remove the temporary
+   location afterward.
+10. Present a concise preflight summary containing:
+    - Previous stable version
+    - Proposed version and tag
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Define a deterministic reconciliation rule when GitHub releases and PyPI disagree. This can occur after a partial release, and without a rule the next-version calculation and existing-version gate can use inconsistent baselines across runs.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+Require a successful response and confirm that release files are present.
+Compare the published filenames against the successful release artifacts when
+that information is available. Do not proceed based only on the Azure DevOps
+run result.
+
+## Phase 6: Create the GitHub Release
+
+1. Generate release notes from the previous stable tag through the new tag.
+   Keep relevant issue and pull request links. Review the generated text for
+   unrelated changes or internal information.
```
StellaHuang95: > **Info** · Optional note

Use the exact package version produced by the Phase 1 metadata build for the PyPI lookup, and bound polling with an interval, maximum duration or attempts, and an explicit timeout failure action. Otherwise normalization differences can miss a valid release and a stalled PyPI replication can leave the agent polling indefinitely.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+    - Proposed version and tag
+    - Full target commit SHA and subject
+    - GitHub repository
+    - Discovered internal build and release pipeline names
+
+Stop here for prepare/dry-run requests.
+
+## Phase 2: Create and Push the Tag
+
+1. Create an annotated tag on the verified `microsoft/debugpy` `main` commit:
```
StellaHuang95: > **Info** · Optional note

Specify how to discover the authoritative required checks and verify their success for the target SHA. As written, the gate has no mechanism or definition of “required”, leaving the agent to invent a command or silently skip it.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,195 @@
...
+   Cancel the run if any value is wrong.
+6. Wait for completion. Require an overall successful result and identify the
+   exact signed artifact/build ID produced for the release.
+
+## Phase 4: Queue the Internal Release Pipeline
+
+1. Discover the internal release pipeline from the authenticated Azure DevOps
+   project. It must consume the signed debugpy build, run final tests, and
+   publish to PyPI through the approved internal publishing path.
+2. Inspect how it selects the input build. Pin it to the exact successful build
```
StellaHuang95: > **Info** · Optional note

Selecting a release input by tag or branch leaves a race if another build for the tag appears between verification and consumption. Require an explicit Phase 3 build/run ID whenever possible; when that is unavailable, define an abort condition equivalent to Phase 3.5 before publication can proceed.

rchiodo:
Addressed all six review findings in c75f78b4:

- deterministic reconciliation when GitHub and PyPI disagree
- required-check discovery through repository rules/branch protection
- resume/finalize support for partial releases
- exact Phase 3 build ID pinning, with abort conditions when pinning is unavailable
- exact normalized package version plus bounded PyPI polling
- conditional `--latest` handling for backports

I also added support for direct internal pipeline locations via `DEBUGPY_INTERNAL_BUILD_PIPELINE_URL` and `DEBUGPY_INTERNAL_RELEASE_PIPELINE_URL`. Their values remain outside the public repository and are validated against the authenticated Azure DevOps scope at runtime.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,257 @@
...
+   Keep relevant issue and pull request links. Review the generated text for
+   unrelated changes or internal information.
+2. Compare the new version to the highest stable GitHub release that existed
+   before this workflow. Add `--latest` only when the new version is greater.
+   Omit it for backports and older release lines.
+3. Create the release in `microsoft/debugpy`:
+
+   ```text
+   gh release create <tag> --repo microsoft/debugpy --title "debugpy <tag>" --notes-file <notes-file> [--latest]
+   ```
```
StellaHuang95, line 284: > **Issue** · Please address or respond

Omitting `--latest` does not force a non-latest release: `gh release create` uses its automatic latest-selection heuristic. A backport released after a newer version can therefore still replace the Latest pointer. Pass `--latest=false` for backports and older release lines. [verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,257 @@
...
+   git tag -a <tag> <full-commit-sha> -m "debugpy <tag>"
+   ```
+
+2. Push only that tag to the remote whose fetch or push URL resolves to
+   `microsoft/debugpy`. Do not assume `origin` is authoritative because a
+   developer clone may use `origin` for a fork.
+3. Verify the remote tag resolves to the intended full commit SHA.
+
+If a local tag was created but the push failed, report that exact state. Never
+move or force-push an existing release tag.
```
StellaHuang95: > **Warning** · Non-blocking recommendation

If creating the local tag succeeds but pushing it fails, a later invocation cannot proceed: new-release preflight expects no tag, while resume resolves the tag from the remote. Define recovery for a local-only tag that points to the verified commit, such as re-pushing that exact tag; otherwise stop with an explicit recovery action. [verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,257 @@
...
+   `https://github.com/microsoft/debugpy.git`. Fetch it if necessary and require
+   `HEAD` to equal that commit. Do not release a fork-only commit. For resume,
+   resolve the commit from the existing remote tag instead and do not retag the
+   current `main`.
+6. Read the complete sets of stable public versions from GitHub releases and
+   PyPI, ignoring prereleases. Reconcile them deterministically:
+   - If the highest stable versions match, use that version as the baseline.
+   - If PyPI contains a newer version than GitHub, treat it as an incomplete
+     release and resume at GitHub release creation. Do not calculate or publish
+     a newer version.
```
StellaHuang95: > **Warning** · Non-blocking recommendation

When PyPI is newer than GitHub, this directs the agent to finalize that existing version even if the user requested a different next-version release. Require confirmation before changing the requested release target, or make this automatic recovery behavior explicit in the intent rules. [verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,257 @@
...
+If the deadline expires, stop in a resumable state and report that publication
+may have succeeded but public index verification did not. Do not proceed based
+only on the Azure DevOps run result.
+
+## Phase 6: Create the GitHub Release
+
+1. Generate release notes from the previous stable tag through the new tag.
+   Keep relevant issue and pull request links. Review the generated text for
+   unrelated changes or internal information.
+2. Compare the new version to the highest stable GitHub release that existed
```
StellaHuang95: > **Warning** · Non-blocking recommendation

For an explicit backport, the “previous stable tag” must be the nearest lower tag on the same release line, not the globally highest stable tag. Otherwise the generated release-notes range can be inverted or include unrelated newer-line changes.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed — backports and older release lines now explicitly pass `--latest=false`, preventing GitHub's automatic heuristic from changing the Latest pointer.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed — resume may re-push a local-only annotated tag only after verifying the remote tag is absent and the local tag still resolves to the previously verified commit; otherwise it stops with the recovery action.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed — a new-release request now stops and asks before switching to a different incomplete PyPI version; only resume/finalize for that exact version proceeds automatically.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed — release notes now use the greatest lower stable tag on the same major/minor line for backports, avoiding inverted or cross-line ranges.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,268 @@
...
+   ```text
+   git status --short
+   git remote -v
+   ```
+
+5. For a new release, resolve the authoritative `main` commit directly from
+   `https://github.com/microsoft/debugpy.git`. Fetch it if necessary and require
+   `HEAD` to equal that commit. Do not release a fork-only commit. For resume,
+   resolve the commit from the existing remote tag instead and do not retag the
+   current `main`.
```
StellaHuang95: > **Warning** · Non-blocking recommendation

A fresh resume cannot reach the Phase 2 local-tag recovery when the original tag push failed: Phase 1 resolves resume commits from a *remote* tag, which is absent in this state, and “previously verified commit” is not available across sessions. Add a Phase 1 recovery branch that verifies an eligible local annotated tag against re-derivable release state before routing to the re-push path.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,268 @@
...
+   - If GitHub contains a newer version than PyPI, stop because customer
+     workflows may already be inconsistent. Report that PyPI publication for
+     the existing GitHub release must be recovered before another release.
+   - If the sets disagree in any other way that affects the proposed version,
+     stop and report both sets. Never choose one source arbitrarily.
+7. Inventory the proposed version's tag, internal build, internal release run,
+   PyPI files, and GitHub release. For a new release, require all to be absent.
+   For resume/finalize, require each existing item to match the same commit,
+   exact signed build, and package version, then continue from the first absent
+   item. Never recreate, move, or overwrite an existing item.
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Make this PyPI-already-published path explicitly override the generic “first absent item” rule in step 7. Otherwise, an expired internal build record could lead a resume to re-queue build/release pipelines and attempt to republish a version that is already on PyPI, rather than proceeding only to GitHub release creation.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,268 @@
...
+
+1. Select the release-notes base deterministically. For a backport or older
+   release line, use the greatest stable tag lower than the new version with the
+   same major and minor components. Otherwise use the greatest stable tag lower
+   than the new version. Stop if no valid lower tag exists. Generate release
+   notes from that tag through the new tag, keep relevant issue and pull request
+   links, and review the text for unrelated changes or internal information.
+2. Compare the new version to the highest stable GitHub release that existed
+   before this workflow. Pass `--latest` when the new version is greater and
+   `--latest=false` for backports and older release lines.
```
StellaHuang95: > **Info** · Optional note

Use the baseline resolved during Phase 1 rather than re-deriving the “highest stable GitHub release that existed before this workflow.” That historical snapshot is not reliably reconstructible during a later resume, while Phase 1 already establishes the authoritative reconciled baseline.

StellaHuang95 (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 5c2eaf80 — Phase 1 now recognizes a local-only annotated tag as a recovery candidate only after re-deriving its eligibility from current main, required checks, absent external state, and exact package metadata.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 5c2eaf80 — an existing exact PyPI publication now explicitly overrides first-absent phase selection and proceeds only to GitHub release creation, without re-queueing expired or missing internal runs.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 5c2eaf80 — Phase 1 records an immutable latest-selection baseline, and Phase 6 uses it directly instead of attempting to reconstruct the earlier GitHub state during resume.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,289 @@
...
+   through the approved authenticated internal release pipeline.
+10. Stop on any failed or partially successful stage. Do not continue with a
+    newer build or a different artifact. Resume only from verified state.
+
+## Pipeline Locations
+
+Prefer these environment variables when they are set:
+
+- `DEBUGPY_INTERNAL_BUILD_PIPELINE_URL`
+- `DEBUGPY_INTERNAL_RELEASE_PIPELINE_URL`
```
StellaHuang95, line 69: > **Warning** · Non-blocking recommendation

📍 .github/agents/debugpy-release.agent.md:65
The Skeptic found that step 5's local-tag recovery says "route to Phase 2 to push that exact tag," which can be read as short-circuiting the rest of Phase 1. If so, step 6 (recording the immutable latest-selection baseline) never runs — yet Phase 6 step 2 requires that baseline to choose `--latest` vs `--latest=false`. This lets the resume fix quietly undermine the twice-verified `--latest` fix on the recovery path. Reword to "proceed through the remainder of Phase 1 (baseline recording, inventory, required-check gate, metadata build, preflight summary), then Phase 2" so the recovery path still records the baseline it depends on.

[verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,289 @@
...
+     version, continue at GitHub release creation. Do not calculate or publish a
+     newer version.
+   - If GitHub contains a newer version than PyPI, stop because customer
+     workflows may already be inconsistent. Report that PyPI publication for
+     the existing GitHub release must be recovered before another release.
+   - If the sets disagree in any other way that affects the proposed version,
+     stop and report both sets. Never choose one source arbitrarily.
+   Record the highest stable GitHub version observed here as the immutable
+   latest-selection baseline for Phase 6. Do not try to reconstruct this
+   historical snapshot later during resume.
```
StellaHuang95: > **Info** · Optional note

📍 .github/agents/debugpy-release.agent.md:128
The Skeptic and Advocate note that "immutable latest-selection baseline recorded during Phase 1" overstates the guarantee: there is no persisted store, so a cross-session resume recomputes the baseline from current GitHub state rather than reading the original snapshot. The safety-critical `--latest=false` backport branch is unaffected, but for a non-backport resume a newly-appeared higher release could flip the decision. Soften to "recompute the baseline deterministically each run; for backports/older lines always pass `--latest=false` regardless of baseline," or persist the baseline with the resume state.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 1c08e1b9 — local-tag recovery now completes all remaining Phase 1 work, including baseline recording and every safety gate, before proceeding to Phase 2.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 1c08e1b9 — the baseline is now explicitly per-invocation and deterministically recomputed on resume. Phase 6 revalidates the release set, while backports always force `--latest=false` regardless of the baseline.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,296 @@
...
+   PyPI files, and GitHub release. For a new release, require all to be absent.
+   Except for the PyPI override below, resume/finalize requires every existing
+   item to match the same commit, exact signed build, and package version, then
+   continues from the first absent item. Never recreate, move, or overwrite an
+   existing item. PyPI publication overrides the generic first-absent rule:
+   when the exact version and files already exist on PyPI but the GitHub release
+   is absent, require the matching remote tag and proceed only to Phase 6. Do
+   not re-queue build or release pipelines even if their historical records are
+   unavailable or expired.
+8. Discover the authoritative required checks for `main` using the GitHub
```
StellaHuang95, line 144: > **Warning** · Non-blocking recommendation

The local-tag recovery path reaches this inventory with an expected local tag, but a new release is required to have its "tag" absent. Clarify that this refers to the remote tag, or explicitly exempt the local recovery tag, so the valid recovery path does not reject itself. Also align the recovery-state checks with the separate internal build and release-run inventory.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,296 @@
...
+   notes from that tag through the new tag, keep relevant issue and pull request
+   links, and review the text for unrelated changes or internal information.
+2. Recheck that the stable GitHub release set has not changed since the Phase 1
+   snapshot. If it changed, repeat Phase 1 reconciliation and update the
+   invocation's latest-selection baseline before continuing. For backports and
+   older release lines, always pass `--latest=false` regardless of the
+   baseline. Otherwise pass `--latest` only when the new version is greater
+   than the current invocation's baseline; pass `--latest=false` when it is
+   not.
+3. Create the release in `microsoft/debugpy`:
```
StellaHuang95, line 280: > **Warning** · Non-blocking recommendation

Repeating Phase 1 reconciliation here can classify this invocation's own in-flight version as a PyPI-newer-than-GitHub incomplete release and stop before `gh release create`. Scope the recheck to releases newer than the version being published, or exclude the already-verified in-flight version from that comparison.

StellaHuang95 (approved):
Approved via Review Center.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 13563bf8 — inventory now distinguishes local and remote tags, explicitly permits the validated local recovery candidate, and checks internal build and release-run absence separately.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 13563bf8 — Phase 6 excludes the exact Phase 5-verified in-flight PyPI version from concurrency reconciliation, while still stopping or recalculating for any other release-state change.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,302 @@
+---
+description: "Release debugpy to PyPI and GitHub. Use when asked to prepare or publish the next patch, minor, major, or explicit debugpy version."
+name: "Debugpy Release"
+argument-hint: "Specify patch, minor, major, or an explicit version. Say 'prepare' for a dry run."
+tools: [vscode/askQuestions, execute/runInTerminal, execute/getTerminalOutput, execute/awaitTerminal, read/readFile, search/fileSearch, search/textSearch, web/fetch, todo]
```
StellaHuang95: > **Issue** · Please address or respond

This non-empty allow-list uses tool aliases as `<server>/<tool>` identifiers. Unrecognized entries are ignored, so the agent can lose the terminal, read, and search capabilities required for every release phase. Omit `tools:` like sibling agents, or replace these with verified IDs supported by the target VS Code host.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,302 @@
...
+   snapshot. When comparing it with PyPI, exclude only the exact in-flight
+   version already verified in Phase 5; its expected PyPI-only state must not be
+   treated as a new incomplete release. If any other release state changed,
+   repeat Phase 1 reconciliation with the in-flight version explicitly marked
+   as expected and update the invocation's latest-selection baseline before
+   continuing. For backports and older release lines, always pass
+   `--latest=false` regardless of the baseline. Otherwise pass `--latest` only
+   when the new version is greater than the current invocation's baseline; pass
+   `--latest=false` when it is not.
+3. Create the release in `microsoft/debugpy`:
```
StellaHuang95, line 280: > **Warning** · Non-blocking recommendation

The PyPI-only resume path goes directly from the Phase 1 inventory to Phase 6, so its version was not verified in Phase 5. Define the exclusion as the in-flight version verified present on PyPI, whether that happened in Phase 5 or during the Phase 1 override inventory; otherwise this recheck can reclassify the release being finalized as an incomplete release.

[verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,302 @@
...
+
+1. Select the release-notes base deterministically. For a backport or older
+   release line, use the greatest stable tag lower than the new version with the
+   same major and minor components. Otherwise use the greatest stable tag lower
+   than the new version. Stop if no valid lower tag exists. Generate release
+   notes from that tag through the new tag, keep relevant issue and pull request
+   links, and review the text for unrelated changes or internal information.
+2. Recheck that the stable GitHub release set has not changed since the Phase 1
+   snapshot. When comparing it with PyPI, exclude only the exact in-flight
+   version already verified in Phase 5; its expected PyPI-only state must not be
```
StellaHuang95: > **Info** · Optional note

Define “backport or older release line” explicitly. This classification controls the release-notes range, and without an objective criterion the agent can select an incorrect range even though the `--latest` decision has a numeric backstop.

[verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,302 @@
...
+   current `main`. If the remote tag is absent but an exact local tag exists,
+   treat it as a local-tag recovery candidate only when:
+   - It is an annotated tag and the requested tag name matches exactly.
+   - The remote tag, internal build, internal release run, PyPI version, and
+     GitHub release are all absent.
+   - Its target equals the current authoritative `microsoft/debugpy` `main`
+     commit and that commit passes the required-check gate below.
+   - Rebuilding metadata from that tag produces the requested normalized
+     package version.
+
```
StellaHuang95: > **Info** · Optional note

The local-tag recovery checks duplicate the later authoritative inventory's absence predicate with slightly different terms (“PyPI version” versus “PyPI files”). Make the inventory the single source of truth, or otherwise unify these checks, so future safety-rule changes cannot silently diverge.

[verified]

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,302 @@
...
+   ```
+
+   Query check runs and commit statuses for the target SHA, match the required
+   contexts and integration IDs, and require every required result to be
+   successful. Do not substitute "all visible checks" for required checks. Stop
+   if the rules cannot be retrieved or mapped unambiguously.
+9. Run the repository's existing targeted packaging/version checks if
+   available. At minimum, use a temporary clone or worktree with a local-only
+   proposed tag to build package metadata without publishing, verify the
+   resulting normalized package version is exactly the proposed version, save
```
StellaHuang95, line 164: > **Info** · Optional note

Specify cleanup for failure paths as well. A failed metadata build or GitHub release creation can leave the temporary worktree or release-notes file behind; either require cleanup or explicitly retain it for diagnosis.

[verified]

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 48046cfb — removed the unverified non-empty tool allow-list so the target host supplies the capabilities needed by the release workflow.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 48046cfb — the verified in-flight version is now established either by the Phase 1 PyPI override inventory or Phase 5 and is excluded consistently during Phase 6 reconciliation.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 48046cfb — backports are now defined objectively by comparing the target and highest stable `(major, minor)` tuples, with an additional monotonicity requirement on the target release line.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 48046cfb — local-tag detection now defers all external absence checks to the authoritative step 7 inventory instead of maintaining a duplicate predicate.

.github/agents/debugpy-release.agent.md
rchiodo: Fixed in 48046cfb — temporary clones, worktrees, metadata output, and release-notes files must now be cleaned on success and failure, with any cleanup failure reported by exact path.

.github/agents/debugpy-release.agent.md
```diff
@@ -0,0 +1,304 @@
...
+   target. Stop if no valid lower tag exists. Generate release notes from that
+   tag through the new tag, keep relevant issue and pull request links, and
+   review the text for unrelated changes or internal information.
+2. Recheck that the stable GitHub release set has not changed since the Phase 1
+   snapshot. When comparing it with PyPI, exclude only the verified in-flight
+   version established either by the Phase 1 PyPI override inventory or by
+   Phase 5; its expected PyPI-only state must not be treated as a new incomplete
+   release. If any other release state changed, repeat Phase 1 reconciliation
+   with the in-flight version explicitly marked as expected and update the
+   invocation's latest-selection baseline before continuing. For backports and
```
StellaHuang95, line 275: > **Warning** · Non-blocking recommendation

When resuming a backport after PyPI has already published the verified in-flight version, this check can compare the target against itself and reject a valid Phase 6 completion. Exclude the verified in-flight version here as step 2 does, or explicitly evaluate this ordering check against the pre-release Phase 1 baseline. [verified]

StellaHuang95 (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

PR #2065, opened 2026-08-14
Check against sys.executable when determining if we should instrument subprocess

pdepetro:
Native python binaries might not include 'python' in the file basename, causing the multiprocess monkey-patching code in pydevd to not recognize its subprocesses as python processes.

Add a fallback check against sys.executable in is_python() so that subprocesses launched via the same executable are correctly instrumented for debugging.

This is useful for custom Python builds or embedded distributions where the executable name doesn't contain 'python', 'jython', or 'pypy'.

### DIFF
--- a/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
+++ b/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
@@ -300,6 +300,9 @@ def is_python(path) -> bool:
         if filename.find(name) != -1:
             return True
 
+    if path == sys.executable:
+        return True
+
     return False
 
 
--- a/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
+++ b/src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
@@ -300,8 +300,41 @@ def is_python(path) -> bool:
         if filename.find(name) != -1:
             return True
 
-    if path == sys.executable:
-        return True
+    # Fallback: check if the path refers to the same executable as sys.executable.
+    # This handles cases where the binary name doesn't contain "python" (e.g.
+    # custom builds, embedded distributions). We use a platform-aware
+    # comparison that covers symlinks, relative paths, and Windows case
+    # differences.
+    try:
+        path_str = path.decode(sys.getfilesystemencoding()) if isinstance(path, bytes) else path
+        sys_exec_str = sys.executable
+
+        # Fast path: exact string match
+        if path_str == sys_exec_str:
+            return True
+
+        # If both exist, samefile is the most reliable (handles symlinks, hardlinks)
+        try:
+            if os.path.exists(path_str) and os.path.exists(sys_exec_str):
+                if os.path.samefile(path_str, sys_exec_str):
+                    return True
+        except (OSError, AttributeError, NotImplementedError, ValueError):
+            pass
+
+        # Fallback: compare normalized realpaths
+        # - realpath resolves symlinks
+        # - abspath is implied by realpath, but we ensure it
+        # - normpath collapses redundant separators
+        # - normcase normalizes case and separators on Windows
+        try:
+            norm_path = os.path.normcase(os.path.normpath(os.path.realpath(path_str)))
+            norm_sys = os.path.normcase(os.path.normpath(os.path.realpath(sys_exec_str)))
+            if norm_path == norm_sys:
+                return True
+        except Exception:
+            pass
+    except Exception:
+        pass
 
     return False
 
--- a/src/debugpy/_vendored/pydevd/tests_python/test_pydev_monkey.py
+++ b/src/debugpy/_vendored/pydevd/tests_python/test_pydev_monkey.py
@@ -512,3 +512,76 @@ def test_monkey_patch_args_c_with_bytes() -> None:
     # The result should contain our patched debugpy setup somewhere
     result_str: str = b"".join(item.encode() if isinstance(item, str) else item for item in result).decode()
     assert "pydevd.settrace" in result_str
+
+
+def test_is_python_sys_executable_exact():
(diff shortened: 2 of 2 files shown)


rchiodo (approved):

rchiodo:

src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -300,6 +300,9 @@ def is_python(path) -> bool:
         if filename.find(name) != -1:
             return True
 
+    if path == sys.executable:
+        return True
```
rchiodo: > **Warning** · Non-blocking recommendation

Raw equality accepts only the literal `sys.executable` spelling, so the same executable invoked through a symlink, relative path, or Windows case variation is still not recognized. Define whether this is intentionally an exact-path contract; if executable identity is intended, use a platform-aware path-equivalence check and cover alternate spellings.

rchiodo: > **Warning** · Non-blocking recommendation

Add a focused regression test that mocks an executable basename without `python`, `jython`, or `pypy` and verifies that `is_python()` recognizes it through `sys.executable`. The test should also establish the intended path-equivalence behavior.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:
Triage output unparseable; posting all candidates.

src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
rchiodo: > **Warning** · Non-blocking recommendation

Raw equality can reject the same executable when represented with different Windows casing, a relative path, or a symlink. Use platform-aware path canonicalization if equivalent spellings should match, and add focused coverage for those forms.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):

heejaechang (approved):

src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -300,6 +300,42 @@ def is_python(path) -> bool:
...
+                    return True
+        except (OSError, AttributeError, NotImplementedError, ValueError):
+            pass
+
+        # Fallback: compare normalized realpaths
+        # - realpath resolves symlinks
+        # - abspath is implied by realpath, but we ensure it
+        # - normpath collapses redundant separators
+        # - normcase normalizes case and separators on Windows
+        try:
```
rchiodo, line 329: > **Warning** · Non-blocking recommendation

A bare command such as `myapp` is resolved through `PATH` by subprocess creation, but these checks resolve it relative to the current working directory. When `sys.executable` has a non-Python basename and is launched by its bare name, the fallback will still classify it as non-Python. Resolve bare executable names with `shutil.which()` before comparing identity, and add a `PATH`-based regression test.

src/debugpy/_vendored/pydevd/_pydev_bundle/pydev_monkey.py
```diff
@@ -300,6 +300,42 @@ def is_python(path) -> bool:
         if filename.find(name) != -1:
             return True
 
+    # Fallback: check if the path refers to the same executable as sys.executable.
+    # This handles cases where the binary name doesn't contain "python" (e.g.
+    # custom builds, embedded distributions). We use a platform-aware
+    # comparison that covers symlinks, relative paths, and Windows case
+    # differences.
+    try:
```
rchiodo, line 308: > **Warning** · Non-blocking recommendation

Strict decoding with `sys.getfilesystemencoding()` rejects valid POSIX byte paths containing undecodable bytes, causing the outer handler to classify the executable as non-Python. Use `os.fsdecode(path)` so surrogate-escaped filesystem paths can still be compared with `sys.executable`.

src/debugpy/_vendored/pydevd/tests_python/test_pydev_monkey.py
```diff
@@ -512,3 +512,76 @@ def test_monkey_patch_args_c_with_bytes() -> None:
     # The result should contain our patched debugpy setup somewhere
     result_str: str = b"".join(item.encode() if isinstance(item, str) else item for item in result).decode()
     assert "pydevd.settrace" in result_str
+
+
+def test_is_python_sys_executable_exact():
+    # Exact sys.executable should be considered python even if basename doesn't contain "python"
```
rchiodo, line 518: > **Info** · Optional note

This test normally passes through the existing basename heuristic and therefore does not prove the new exact-identity fallback. Monkeypatch `sys.executable` to a custom-named alias and assert that the same value is recognized.

src/debugpy/_vendored/pydevd/tests_python/test_pydev_monkey.py
```diff
@@ -512,3 +512,76 @@ def test_monkey_patch_args_c_with_bytes() -> None:
...
+        # On case-sensitive systems, upper path likely doesn't exist, so should be False
+        # unless it accidentally matches via other heuristics - just ensure bool
+        assert isinstance(pydev_monkey.is_python(upper), bool)
+
+
+def test_is_python_non_python():
+    assert pydev_monkey.is_python("/bin/ls") is False
+    assert pydev_monkey.is_python("/usr/bin/myapp") is False
+
+
```
rchiodo, line 581: > **Warning** · Non-blocking recommendation

`/usr/bin/myapp` can legitimately be `sys.executable` on the custom distributions targeted by this PR, making this negative assertion fail correctly. Use a guaranteed-distinct temporary file instead.

rchiodo:

PR #2066, opened 2026-08-16
Set TCP_NODELAY on pydevd client connections

### LINKED ISSUE #2054: pydevd_comm.py start_client() does not set TCP_NODELAY on the connections it opens
## Environment data

- debugpy version: 1.8.20
- OS and version: Linux and MacOS
- Python version (& distribution if applicable, e.g. Anaconda): 3.11 and later
- Using VS Code or Visual Studio: No

## Actual behavior

TCP_NODELAY is not set on the connections debugpy opens

## Expected  behavior

TCP_NODELAY is set on the connections debugpy opens

## Steps to reproduce:

A connection established with `debugpy --connect` does not set TCP_NODELAY on the socket connection, which can cause latency when sending responses and events. This is can be verified simply by inspecting the pydevd_comm.py source code, or by using external tools to inspect the connection once it has been established.

We patched our copy of debugpy 1.8.20 as follows:

```python
% diff --git a/3rdparty/debugpy/DEBUGPY/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py b/3rdparty/debugpy/DEBUGPY/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
index 64ac454f29..812bf59eb0 100644
--- a/3rdparty/debugpy/DEBUGPY/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
+++ b/3rdparty/debugpy/DEBUGPY/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py
@@ -524,6 +524,15 @@ def start_client(host, port):
     except (AttributeError, OSError):
         pass  # May not be available everywhere.
 
+    try:
+        # TVT-46072: "Missing TCP_NODELAY on debugpy's DAP tracer connection
+        # causes Nagle-induced latency in Python debugging". Disable Nagle's
+        # algorithm so small DAP messages are sent immediately instead of
+        # being buffered, reducing round-trip latency.
+        s.setsockopt(socket_module.IPPROTO_TCP, socket_module.TCP_NODELAY, 1)
+    except (AttributeError, OSError):
+        pass  # May not be available everywhere.
+
     try:
         # 10 seconds default timeout
         timeout = int(os.environ.get("PYDEVD_CONNECT_TIMEOUT", 10))
%
```

nightcityblade:
Fixes #2054

## Summary

- disable Nagle's algorithm on outbound pydevd client sockets
- tolerate platforms that do not expose or support `TCP_NODELAY`
- add a focused socket mock test for the new option

## Testing

- `PYTHONPATH=. python3 -m pytest -q tests_python/test_pydevd_comm.py tests_python/test_fixtures.py` (4 passed)
- `python3 -m ruff format --check _pydevd_bundle/pydevd_comm.py tests_python/test_pydevd_comm.py`
- Ruff lint checks on both changed files

### DIFF
--- a/src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
+++ b/src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
@@ -1,0 +1,43 @@
+from unittest import mock
+
+from _pydevd_bundle import pydevd_comm
+
+
+def start_client(monkeypatch, sock):
+    monkeypatch.setattr(pydevd_comm, "socket", lambda *_: sock)
+    monkeypatch.setattr(pydevd_comm.socket_module, "getaddrinfo", lambda *_: [])
+    return pydevd_comm.start_client("localhost", 5678)
+
+
+def test_start_client_sets_tcp_nodelay(monkeypatch):
+    sock = mock.Mock()
+    tcp_nodelay = mock.sentinel.tcp_nodelay
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", tcp_nodelay)
+
+    assert start_client(monkeypatch, sock) is sock
+    sock.setsockopt.assert_any_call(
+        pydevd_comm.socket_module.IPPROTO_TCP,
+        tcp_nodelay,
+        1,
+    )
+
+
+def test_start_client_without_tcp_nodelay(monkeypatch):
+    sock = mock.Mock()
+    monkeypatch.delattr(pydevd_comm.socket_module, "TCP_NODELAY", raising=False)
+
+    assert start_client(monkeypatch, sock) is sock
+
+
+def test_start_client_ignores_tcp_nodelay_error(monkeypatch):
+    sock = mock.Mock()
+    tcp_nodelay = mock.sentinel.tcp_nodelay
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", tcp_nodelay)
+
+    def set_option(_level, option, _value):
+        if option is tcp_nodelay:
+            raise OSError
+
+    sock.setsockopt.side_effect = set_option
+
+    assert start_client(monkeypatch, sock) is sock
--- a/src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
+++ b/src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
@@ -18,7 +18,7 @@ def start_client(monkeypatch, sock):
 
 def test_start_client_sets_tcp_nodelay(monkeypatch):
     sock = mock.Mock()
-    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay)
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay, raising=False)
 
     start_client(monkeypatch, sock)
     sock.setsockopt.assert_any_call(pydevd_comm.socket_module.IPPROTO_TCP, mock.sentinel.tcp_nodelay, 1)
@@ -28,7 +28,7 @@ def test_start_client_sets_tcp_nodelay(monkeypatch):
 def test_start_client_ignores_tcp_nodelay_error(monkeypatch, error):
     sock = mock.Mock()
     sock.setsockopt.side_effect = error
-    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay)
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay, raising=False)
 
     start_client(monkeypatch, sock)
     sock.setsockopt.assert_any_call(pydevd_comm.socket_module.IPPROTO_TCP, mock.sentinel.tcp_nodelay, 1)
@@ -39,7 +39,7 @@ def test_start_server_sets_tcp_nodelay(monkeypatch):
     address = ("127.0.0.1", 5678)
     server.configure_mock(**{"accept.return_value": (accepted, address), "getsockname.return_value": address})
     monkeypatch.setattr(pydevd_comm, "create_server_socket", mock.Mock(return_value=server))
-    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay)
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay, raising=False)
     assert pydevd_comm.start_server(0) is accepted
     accepted.setsockopt.assert_called_once_with(pydevd_comm.socket_module.IPPROTO_TCP, mock.sentinel.tcp_nodelay, 1)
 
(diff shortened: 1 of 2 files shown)


rchiodo:

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
```diff
@@ -0,0 +1,16 @@
...
+def test_start_client_sets_tcp_nodelay(monkeypatch):
+    sock = mock.Mock()
+    monkeypatch.setattr(pydevd_comm, "socket", lambda *_: sock)
+    monkeypatch.setattr(pydevd_comm.socket_module, "getaddrinfo", lambda *_: [])
+
+    assert pydevd_comm.start_client("localhost", 5678) is sock
+    sock.setsockopt.assert_any_call(
+        pydevd_comm.socket_module.IPPROTO_TCP,
+        pydevd_comm.socket_module.TCP_NODELAY,
+        1,
```
rchiodo: > **Warning** · Non-blocking recommendation

The assertion directly accesses `TCP_NODELAY`, so this test itself will fail on platforms where that constant is absent—the case the production code now handles. Please use a sentinel for the success case and add coverage for both a missing constant and `setsockopt` raising `OSError`.

rchiodo (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
nightcityblade: Implemented in 6100e76: the success test now uses a sentinel, and separate tests cover a missing TCP_NODELAY constant and OSError from setsockopt. The focused pydevd test file passes (3 tests), and Ruff passes as well. Thank you for the recommendation!

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
```diff
@@ -0,0 +1,43 @@
+from unittest import mock
+
+from _pydevd_bundle import pydevd_comm
+
+
+def start_client(monkeypatch, sock):
+    monkeypatch.setattr(pydevd_comm, "socket", lambda *_: sock)
+    monkeypatch.setattr(pydevd_comm.socket_module, "getaddrinfo", lambda *_: [])
+    return pydevd_comm.start_client("localhost", 5678)
```
rchiodo: > **Warning** · Non-blocking recommendation

Returning no address candidates exercises the early no-connection path. Return a realistic mocked address, mock `connect`, and assert both the connection and `TCP_NODELAY` calls so the test covers the intended outbound path.

[verified]

rchiodo (approved):

heejaechang (approved):
Approved via Review Center.

rchiodo:
Should `start_server()` also set `TCP_NODELAY` on `new_socket` after `accept()`? Both connection modes carry the same latency-sensitive protocol, and limiting this to `start_client()` leaves server-mode traffic asymmetric. Please use the same guarded `setsockopt()` handling.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py:L64](https://github.com/microsoft/debugpy/blob/6100e767424c3f331513cd49cda987db77782a19/src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py#L64).

> **Info** · Optional note

📍 src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py:53
The exception tests cover a missing `TCP_NODELAY` attribute and `setsockopt()` raising `OSError`, but not `setsockopt()` raising `AttributeError`. Consider parameterizing this test over both caught exception types so the full best-effort contract is explicit.

[verified]

rchiodo (approved):
Approved via Review Center.

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
nightcityblade: Addressed in 26ff987. The test now uses realistic getaddrinfo results, exercises the selected-address connect path, and asserts both connect and TCP_NODELAY. I also parameterized the best-effort behavior for AttributeError and OSError.

nightcityblade:
Addressed the remaining TCP_NODELAY feedback in 26ff987:

- set TCP_NODELAY best-effort on the accepted start_server socket as well
- exercise realistic getaddrinfo and connect behavior in the client test
- cover both AttributeError and OSError from setsockopt
- add server-side coverage

Validation: 4 focused tests passed; Ruff check on the test file and Ruff format checks on both changed files passed; git diff --check passed.

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
```diff
@@ -0,0 +1,42 @@
...
+    sock.setsockopt.assert_any_call(pydevd_comm.socket_module.IPPROTO_TCP, pydevd_comm.socket_module.TCP_NODELAY, 1)
+
+
+def test_start_server_sets_tcp_nodelay(monkeypatch):
+    server, accepted = mock.Mock(), mock.Mock()
+    address = ("127.0.0.1", 5678)
+    server.configure_mock(**{"accept.return_value": (accepted, address), "getsockname.return_value": address})
+    monkeypatch.setattr(pydevd_comm, "create_server_socket", mock.Mock(return_value=server))
+    assert pydevd_comm.start_server(0) is accepted
+    accepted.setsockopt.assert_called_once_with(pydevd_comm.socket_module.IPPROTO_TCP, pydevd_comm.socket_module.TCP_NODELAY, 1)
```
rchiodo: > **Warning** · Non-blocking recommendation

On platforms without `TCP_NODELAY`, these assertions dereference `pydevd_comm.socket_module.TCP_NODELAY` directly and will raise `AttributeError`, even though the implementation intentionally supports that case. Use a sentinel or explicitly test the missing-attribute path; add equivalent server-side error-tolerance coverage too.

rchiodo (approved):

heejaechang (approved):

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
nightcityblade: Addressed in 6c3f75f. The client and server tests now inject a TCP_NODELAY sentinel instead of dereferencing the platform constant, and the server path covers both AttributeError and OSError from setsockopt. Validation: all 6 focused tests passed; Ruff check/format and git diff --check passed.

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
```diff
@@ -0,0 +1,57 @@
...
+        "getaddrinfo",
+        lambda *_: [(pydevd_comm.AF_INET, pydevd_comm.SOCK_STREAM, 0, "", ("127.0.0.1", 5678))],
+    )
+    assert pydevd_comm.start_client("localhost", 5678) is sock
+    sock.connect.assert_called_once_with(("localhost", 5678))
+
+
+def test_start_client_sets_tcp_nodelay(monkeypatch):
+    sock = mock.Mock()
+    monkeypatch.setattr(pydevd_comm.socket_module, "TCP_NODELAY", mock.sentinel.tcp_nodelay)
```
rchiodo: > **Warning** · Non-blocking recommendation

`monkeypatch.setattr` defaults to `raising=True`, so this setup fails on platforms where `TCP_NODELAY` is genuinely absent. Use `raising=False` for the sentinel setup and add a test that removes `socket_module.TCP_NODELAY` to exercise the intended compatibility path for both client and server startup.

rchiodo (approved):

heejaechang (approved):

src/debugpy/_vendored/pydevd/tests_python/test_pydevd_comm.py
nightcityblade: Addressed in 2682bbe. All `TCP_NODELAY` sentinel injections now use `raising=False`, and a new regression test removes `socket_module.TCP_NODELAY` and exercises both client and server startup.

Validation: `python3 -m pytest tests_python/test_pydevd_comm.py -q` (7 passed); Ruff check and format check passed; `git diff --check` passed.

rchiodo (approved):

heejaechang (approved):

rchiodo:

rchiodo:

bschnurr (approved):
Approved via Review Center.

PR #2069, opened 2026-08-27
Add python 3.15 to tests

finnagin:


rchiodo (approved):

finnagin:
The tests are all passing now. There were 2 issues uncovered by updating to 3.15.

The smaller issue was that there was a deprecation warning for calling `os.fork` from a process with multiple threads. That warning was added in 3.12 but was not originally raised to be an error by cpython when `PYTHONWARNINGS=error` was set. This was a bug and was fixed in 3.15  (https://github.com/python/cpython/issues/135427) and now, because it's raising an error, caused a few multiprocessing tests to fail.

This was fixed by adding a step in those tests to ignore that specific deprecation warning since those tests were specifically testing `os.fork`.

The larger issue involved duplicate exception stops when using the `sys.monitoring` backend in the vendored pydevd code. Pydevd keeps a temporary thread-local state identifying the exception currently being tracked as unhandled. That state was used both to locate the exception’s unhandled boundary and to determine whether the exception had already caused a debugger stop.

While pydevd was processing the first unhandled `SystemExit` stop, it formatted the exception’s traceback. Python 3.15 traceback code now has a lazy import for `_colorize`. This lazy import triggers during the handling of the `SystemExit` stop. During that import, the import machinery probes `collections.abc.__path__`, which raises a benign, caught `AttributeError`. Pydevd’s monitoring callbacks observed that nested exception and replaced the temporary state that had identified the original `SystemExit` with a new temporary state for the `AttributeError`.

When execution continued, the original `SystemExit` resumed unwinding. Because its identifying state had been replaced by the nested `AttributeError`, pydevd no longer recognized it as already reported and stopped on the same `SystemExit` a second time. Since the import in 3.14 was eager, the caught `AttributeError` happened earlier in the code, so the bug wasn't triggered.

This second issue was fixed by tracking an already-reported exception separately from the temporary state used to find its unhandled boundary. Nested exceptions then can replace the temporary state, but won't erase the record that the original exception has already caused a debugger stop. Also, a test was added to check this specific bug instead of relying on it incidentally getting triggered by an unrelated change in the timing of when a module gets imported.

bschnurr:
🔒 Automated review in progress — @bschnurr is auto-reviewing this PR.

bschnurr:

**Result:** `could-not-verify`

**Summary:** Verification could not run because no trusted sandbox image is configured for \`microsoft/debugpy\`. The PR adds one regression test for repeated unhandled-exception reporting in the pure-Python monitoring implementation. No executable pass/fail signal was obtained. Coverage is still missing for the parallel Cython implementation and Python 3.15 attach-version detection.

**Test runs:** 1 not run

- :warning: **Not run** | <code>Verification environment preflight</code> | <code>git --no-pager status --short &amp;&amp; git --no-pager log -3 --oneline &amp;&amp; git --no-pager diff --name-status HEAD^ HEAD &amp;&amp; python --version &amp;&amp; python -m pytest --version</code>

<details><summary>:warning: <code>Verification environment preflight</code> diagnostic output</summary>

<pre>Container verification could not start and local execution was not authorized: No trusted sandbox image is configured for microsoft/debugpy.</pre>
</details>

bschnurr (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #2073, opened 2026-09-21
Fix busy loop when a peer disconnects in the middle of a message header

### LINKED ISSUE #2072: Adapter spins at 100% CPU when a peer disconnects in the middle of a message header
## Environment data

- debugpy version: 1.8.22, running from source at e2208059
- OS and version: macOS 26.6.2 (arm64)
- Python version (& distribution if applicable, e.g. Anaconda): 3.10.6, CPython from python.org
- Using VS Code or Visual Studio: neither. The repro below is plain Python.

## Actual behavior

`JsonIOStream.read_json()` never returns and never raises when the stream hits EOF part way through a header line. It burns a full core while it does that. `_read_line()` keeps calling `reader.readline()`, which returns `b""` right away on a socket that is already at EOF, so the loop has nothing to block on and no bound.

Anything listening for a DAP client is exposed to this. `python -m debugpy --listen 0.0.0.0:5678 --wait-for-client app.py` opens a port, and any peer that writes a few bytes with no CRLF and then closes pins a core: a port scanner, a health check, an HTTPS client aimed at the wrong port. It also happens with no stranger involved. If the client or the debuggee is killed while a header is half flushed, the side that survives spins instead of ending the session. That thread is a daemon, so nothing reaps it, and every connection that does this leaves another one spinning.

This is near #1308 but it is not the same thing. A peer that closes without sending anything is handled correctly today. Only a partial line spins, which the first line of output below shows.

## Expected  behavior

`NoMoreMessages`, same as when the peer closes before sending anything. That is what `JsonMessageChannel._parse_incoming_messages()` is waiting for so it can fail the outstanding requests, queue the disconnect and close the channel.

## Steps to reproduce:

1. Check out e2208059 and run this with `src` on `PYTHONPATH`. Stdlib only. No debug session and no client.

```python
import socket
import threading
import time

from debugpy.common.messaging import JsonIOStream


def read_after(payload):
    ours, theirs = socket.socketpair()
    stream = JsonIOStream.from_socket(ours)
    theirs.sendall(payload)
    theirs.close()

    done = threading.Event()

    def run():
        try:
            stream.read_json()
        except Exception:
            pass
        finally:
            done.set()

    cpu = time.process_time()
    threading.Thread(target=run, daemon=True).start()
    finished = done.wait(3)
    print(
        "payload %-22r finished=%-5s CPU=%.2fs"
        % (payload, finished, time.process_time() - cpu)
    )


read_after(b"")  # peer closes without sending anything
read_after(b"Content-Length: 24")  # peer closes part way through a header line
```

2. Look at the second line of output.

What I get:

```
payload b''                    finished=True  CPU=0.00s
payload b'Content-Length: 24'  finished=False CPU=2.93s
```

The second read is still going when the script stops waiting after 3 seconds, and almost the whole wait was CPU time. That is one core pinned, and it does not stop. Across four runs the number sits between 2.93 and 3.00. I expected both lines to read `finished=True CPU=0.00s`.

Om-singhaI:
Fixes #2072

`JsonIOStream._read_line()` decides it has hit EOF by testing `line`, the accumulator it is building up, rather than the bytes it just read. Once any part of a header line has arrived, `line` is never empty again, so `if not line` can only ever fire on the very first read. When a stream ends part way through a header, `readline()` returns `b""` from then on, `line` never grows and never ends with CRLF, and the loop spins. No sleep, nothing to block on, no bound.

`readline()` returning `b""` here means EOF and only EOF. `from_socket()` does `sock.settimeout(None)`, and the only other `settimeout` calls on a DAP stream outside `_vendored` also pass `None`. The one real timeout, in `sockets.py`, is on the listening socket and not on a stream. So the fix is to test the chunk, and to treat a truncated header the way the body loop a few lines further down already treats a truncated body: there are no more messages. The clean disconnect path is untouched, since the first `readline()` then returns `b""` while the accumulator is still empty.

What changed:

- `_read_line()` reads into `chunk`, raises `NoMoreMessages` when `chunk` is empty, then appends it to `line`.
- A comment on that check saying why it is on the chunk, so it does not get folded back.
- `TestJsonIOStream.test_read_truncated_header` feeds a reader that yields `b"Content-Length: 24"` and then EOF forever, and asserts `NoMoreMessages` plus exactly one read past EOF.
- That reader raises `ReaderSpinning` after ten reads past EOF instead of hanging, so on the old code the test fails in under a second with a message that names the problem, rather than sitting there until the suite timeout fires.
- The reader derives from `io.RawIOBase`, so it satisfies the `reader` annotation on `JsonIOStream.__init__`, and its `readline` takes `size: int | None = -1` to match the `IOBase` signature.

Testing, on macOS 26.6.2 with CPython 3.10.6, each run with `-o addopts=` to keep it off the 8 worker default:

- `pytest tests/debugpy/common/test_messaging.py -k "not fuzz"`: 13 passed, 1 deselected in 2.32s.
- `pytest tests/debugpy/common/test_messaging.py -k "fuzz"`: 1 passed, 13 deselected in 0.42s.
- `pytest tests/debugpy/common/test_json.py tests/debugpy/common/test_socket.py`: 10 passed in 1.76s.
- Negative control. With `src/debugpy/common/messaging.py` put back to e2208059 and the new test kept: 1 failed, 12 passed, 1 deselected in 2.23s. The failure is `ReaderSpinning: readline() was called 11 times at EOF`, raised out of `messaging.py:182`.
- The repro from the issue, over a real `socket.socketpair()`. Before, the truncated case reported `finished=False CPU=2.93s`. After, `finished=True CPU=0.00s`. The case where the peer closes without sending anything reads `finished=True CPU=0.00s` both before and after.
- `python -m ruff check .`, which is what the Lint stage runs: all checks passed, on ruff 0.16.8.
- Pyright 1.1.411 in `standard` mode, on `tests/debugpy/common/test_messaging.py`: 5 errors, the same 5 that are on it at e2208059, all `reportOptionalMemberAccess` in tests this PR does not touch. The new helper adds none.
- `python -m flake8 src/debugpy/common/messaging.py tests/debugpy/common/test_messaging.py`: the same 5 findings as on e2208059, all of them in code this PR does not touch. Nothing new. The test file is clean.
- Black 26.5.1 leaves every added line as written. Both files already differ from Black on e2208059 in untouched code, so I left that alone instead of mixing a reformat into this.

I left the same unbounded shape in `write_json()` as it is. I could not build a writer that reaches it, so there would be nothing behind the change.


bschnurr:
🔒 Automated review in progress — @bschnurr is auto-reviewing this PR.

bschnurr:
**Result:** ⚠️ `partially-verified`

<details><summary>Verification details</summary>

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Offline editable dependency bootstrap. The relevant tests could not be fully run in the isolated environment; this review is not fully verified.

**Summary:** The new truncated-header regression test passed, and the complete messaging, JSON, and socket test files passed with 25 total test executions. The initial test invocation could not import \`debugpy\`, and editable installation failed because the sandbox lacked \`setuptools.build\_meta\`; running directly from \`src\` provided full targeted coverage. Verification is partial because the standard installation path could not be established, though no test assertion failed.

**Test runs:** 4 passed, 1 failed, 1 not run

- :warning: **Not run** | <code>Messaging tests using default environment</code> | <code>python -m pytest tests/debugpy/common/test_messaging.py -k &quot;not fuzz&quot; -o addopts= -q</code>
- :x: **Failed** | unrelated to this PR | <code>Offline editable dependency bootstrap</code> | <code>python -m venv --system-site-packages .venv &amp;&amp; .venv/bin/python -m pip install --no-index --no-build-isolation -e .</code>
- :white_check_mark: **Passed** | <code>Truncated header regression test</code> | <code>PYTHONPATH=src python -m pytest tests/debugpy/common/test_messaging.py::TestJsonIOStream::test_read_truncated_header -o addopts= -q</code>
- :white_check_mark: **Passed** | <code>Complete messaging tests</code> | <code>PYTHONPATH=src python -m pytest tests/debugpy/common/test_messaging.py -o addopts= -q</code>
- :white_check_mark: **Passed** | <code>JSON and socket tests</code> | <code>PYTHONPATH=src python -m pytest tests/debugpy/common/test_json.py tests/debugpy/common/test_socket.py -o addopts= -q</code>
- :white_check_mark: **Passed** | <code>Sandbox and dependency discovery</code> | <code>printf &#x27;AUTOMATION_SANDBOX_PROFILE=%s\n&#x27; &quot;${AUTOMATION_SANDBOX_PROFILE:-}&quot;; git diff --name-status HEAD^ HEAD; python --version; python - &lt;&lt;&#x27;PY&#x27;
mods = [&#x27;pytest&#x27;, &#x27;pytest_timeout&#x27;]
for mod in mods:
    try:
        m = __import__(mod)
        print(f&#x27;{mod}: available {getattr(m, &quot;__version__&quot;, &quot;&quot;)}&#x27;.rstrip())
    except Exception as exc:
        print(f&#x27;{mod}: unavailable: {exc}&#x27;)
PY
printf &#x27;\nDependency manifests:\n&#x27;; ls -1 pyproject.toml pytest.ini setup.cfg requirements*.txt 2&gt;/dev/null || true</code>

<details><summary>:warning: <code>Messaging tests using default environment</code> diagnostic output</summary>

<pre>ImportError while loading conftest &#x27;/workspace/tests/conftest.py&#x27;.
tests/__init__.py:54: in &lt;module&gt;
    from debugpy.common import json, log
E   ModuleNotFoundError: No module named &#x27;debugpy&#x27;</pre>
</details>

<details><summary>:x: <code>Offline editable dependency bootstrap</code> diagnostic output</summary>

<pre>Obtaining file:///workspace
Checking if build backend supports build_editable: finished with status &#x27;done&#x27;
pip._vendor.pyproject_hooks._impl.BackendUnavailable: Cannot import &#x27;setuptools.build_meta&#x27;</pre>
</details>
</details>

bschnurr (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):

rchiodo:
Thanks for the PR.

